PR #6, opened 2022-04-11
feat: javascript modules auto deployment to npm based on labels

mathewmeconry:
## Description

Adds automation for javascript packages do be published and released based on pull request labels.
It also automatically adds the pending changelog entries (everything under the [TBD] title) to the github releases and updates the changelog file with a new [TBD] section.

## Type of change

## Checklist:

- [x] I have updated the ``CHANGELOG.md`` file in the root folder.


.github/workflows/javascript_publish.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          prerelease: false
+          body_path: ${{ steps.notes.outputs.notes }}
+      - name: Update Changelog
+        run: |
+          git fetch
+          git pull
+          git config --global user.name "Arabot-1"
+          git config --global user.email "arabot-1@users.noreply.github.com"
+          sed -i "s/\[TBD\]/[TBD]  \n  \n\## [1.0.5\] - $(date)/" "javascript/modules/${{ needs.checkTag.outputs.package }}/CHANGELOG.md"
```
brickpop: Is the `1.0.5` literal expected?

mathewmeconry: uhh nop. good spot. I'll change that shortly

marcvelmer (commented):
Great job!

I think this should immediately be merged once we have the unit testing working on the CI, which will be the first issue to be tackled on this new sprint.

brickpop:
Thank you for the changes @mathewmeconry!

I wonder if there is a way to start/stop the ganache service which is self contained in the tests. In dvote-js we were doing something like this programmatically:
- https://github.com/vocdoni/dvote-js/blob/main/packages/client/test/helpers/web3-service.ts
- https://github.com/vocdoni/dvote-js/blob/main/packages/client/test/integration/processes.ts#L54-L70

I thought that there would be a way to do this with HardHat, but can't seem to find anything in the docs :-/

Similar thing with the contract deployment, is it doable to deploy them from the test itself within the `beforeEach()`?

mathewmeconry:
> Thank you for the changes @mathewmeconry!
> 
> I wonder if there is a way to start/stop the ganache service which is self contained in the tests. In dvote-js we were doing something like this programmatically:
> 
> * https://github.com/vocdoni/dvote-js/blob/main/packages/client/test/helpers/web3-service.ts
> * https://github.com/vocdoni/dvote-js/blob/main/packages/client/test/integration/processes.ts#L54-L70
> 
> I thought that there would be a way to do this with HardHat, but can't seem to find anything in the docs :-/
> 
> Similar thing with the contract deployment, is it doable to deploy them from the test itself within the `beforeEach()`?

done. I added it into the beforeAll because then we can speed up the tests a bit and the tests don't need a clean factory every time.

javascript/modules/client/src/context.ts
```diff
@@ -6,7 +6,7 @@ import { UnsupportedProtocolError } from "@aragon/sdk-common";
 export { ContextParams } from "./internal/interfaces/context";
 
 const supportedProtocols = ["https:"];
-if(process.env.CI) {
+if(process.env.TESTING) {
```
brickpop: I see this was already introduced, but I'm not sure if the NodeJS-only `process.env` will bring us trouble when running on the browser

brickpop:
The PR looks perfect, my only doubt is about introducing the `process` dependent piece of code

javascript/modules/client/src/context.ts
```diff
@@ -6,7 +6,7 @@ import { UnsupportedProtocolError } from "@aragon/sdk-common";
 export { ContextParams } from "./internal/interfaces/context";
 
 const supportedProtocols = ["https:"];
-if(process.env.TESTING) {
+if(process && process.env && process.env.TESTING) {
```
```suggestion
if(typeof process !== "undefined" && process.env?.TESTING) {
```

mathewmeconry: The change you suggest only checks for undefined. if you don't use typeof it also checks for empty strings, 0 and NULL

brickpop (approved):

marcvelmer (approved):

PR #9, opened 2022-05-06
Changing in the gas fee estimations functions

marcvelmer:
## Description

Changes in the gas estimation functions in order to

- Adapt the EIP-1559 changes from the `ethers` library
- Added a new `avg` estimation instead of only having the `max` estimation

Task: [ID]()

## Type of change

## Checklist:

- [x] I have updated the ``CHANGELOG.md`` file in the root folder of the package after the [TBD] title and before the latest version.

### DIFF
--- a/javascript/modules/client/src/client-dao.ts
+++ b/javascript/modules/client/src/client-dao.ts
@@ -2,6 +2,7 @@ import { ClientCore } from "./internal/client-core";
 import {
   DaoConfig,
   DaoRole,
+  IGasFeeEstimation,
   IClientDaoBase,
   IClientDaoERC20Voting,
   IClientDaoWhitelistVoting,
@@ -28,7 +29,9 @@ export class ClientDaoERC20Voting extends ClientCore
   dao = {
     create: async (params: ICreateDaoERC20Voting): Promise<string> => {
       if (!this.signer)
-        return Promise.reject(new Error("A signer is needed for creating a DAO"));
+        return Promise.reject(
+          new Error("A signer is needed for creating a DAO")
+        );
       const daoFactoryInstance = DAOFactory__factory.connect(
         this.daoFactoryAddress,
         this.connectedSigner
@@ -47,7 +50,8 @@ export class ClientDaoERC20Voting extends ClientCore
           const newDaoAddress = cr.events?.find(
             e => e.address === registryInstance.address
           )?.topics[1];
-          if (!newDaoAddress) return Promise.reject(new Error("Could not create DAO"));
+          if (!newDaoAddress)
+            return Promise.reject(new Error("Could not create DAO"));
 
           return "0x" + newDaoAddress.slice(newDaoAddress.length - 40);
         });
@@ -69,11 +73,15 @@ export class ClientDaoERC20Voting extends ClientCore
         params: ICreateProposal
       ): Promise<BigNumber> => {
         if (!this.signer)
-          return Promise.reject(new Error("A signer is needed for creating a DAO"));
+          return Promise.reject(
+            new Error("A signer is needed for creating a DAO")
+          );
         else if (!votingAddress)
-          return Promise.reject(new Error(
-            "A voting contract address is needed for creating a proposal"
-          ));
+          return Promise.reject(
+            new Error(
+              "A voting contract address is needed for creating a proposal"
+            )
+          );
         const erc20VotingInstance = ERC20Voting__factory.connect(
           votingAddress,
           this.connectedSigner
@@ -87,7 +95,9 @@ export class ClientDaoERC20Voting extends ClientCore
               e => e.event === "StartVote"
             );
             if (!startVoteEvent)
-              return Promise.reject(new Error("Could not find StartVote event"));
+              return Promise.reject(
+                new Error("Could not find StartVote event")
+              );
 
             return startVoteEvent.args?.voteId;
           });
@@ -116,17 +126,31 @@ export class ClientDaoERC20Voting extends ClientCore
 
   /** Estimation related methods */
   estimate = {
-    create: async (params: ICreateDaoERC20Voting): Promise<BigNumber> => {
+    create: async (
+      params: ICreateDaoERC20Voting
+    ): Promise<IGasFeeEstimation> => {
       if (!this.signer)
-        return Promise.reject(new Error("A signer is needed for creating a DAO"));
+        return Promise.reject(
+          new Error("A signer is needed for creating a DAO")
+        );
       const daoFactoryInstance = DAOFactory__factory.connect(
         this.daoFactoryAddress,
         this.connectedSigner
       );
 
-      return daoFactoryInstance.estimateGas.newERC20VotingDAO(
+      const gasLimit = daoFactoryInstance.estimateGas.newERC20VotingDAO(
         ...ClientDaoERC20Voting.createDaoParameters(params)
       );
+
+      return Promise.all([this.maxFeePerGas, gasLimit]).then(data => {
+        return {
+          avg: data[0]
+            .mul(data[1])
+            .div(this.feeEstimationReducer)
+            .mul(BigNumber.from(100)),
+          max: data[0].mul(data[1]),
+        };
+      });
     },
   };
 
@@ -168,7 +192,9 @@ export class ClientDaoWhitelistVoting extends ClientCore
   dao = {
     create: async (params: ICreateDaoWhitelistVoting): Promise<string> => {
       if (!this.signer)
-        return Promise.reject(new Error("A signer is needed for creating a DAO"));
+        return Promise.reject(
+          new Error("A signer is needed for creating a DAO")
+        );
       const daoFactoryInstance = DAOFactory__factory.connect(
         this.daoFactoryAddress,
         this.connectedSigner
@@ -189,7 +215,8 @@ export class ClientDaoWhitelistVoting extends ClientCore
           const newDaoAddress = cr.events?.find(
             e => e.address === registryInstance.address
           )?.topics[1];
-          if (!newDaoAddress) return Promise.reject(new Error("Could not create DAO"));
+          if (!newDaoAddress)
+            return Promise.reject(new Error("Could not create DAO"));
 
           return "0x" + newDaoAddress.slice(newDaoAddress.length - 40);
         });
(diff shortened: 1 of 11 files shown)


javascript/modules/client/src/context.ts
```diff
@@ -16,6 +16,7 @@ let defaultState: ContextState = {
   dao: "",
   daoFactoryAddress: "",
   web3Providers: [],
+  gasFeeReducer: 0.625,
```
brickpop: Will this value remain in the future? If s, I would suggest a clearer name, like `gasFeeEstimationFactor` or `gasFeeAverageFactor

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -4,15 +4,20 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
 import { IDAO } from "@aragon/core-contracts-ethers";
-import { BigNumberish } from "@ethersproject/bignumber";
+import { BigNumber, BigNumberish } from "@ethersproject/bignumber";
 
 export abstract class ClientCore implements IClientCore {
   private _web3Providers: JsonRpcProvider[] = [];
   private _web3Idx = -1;
   private _signer: Signer | undefined;
   private _daoFactoryAddress = "";
+  private _gasFeeReducer = 1;
```
brickpop: Same as above, if this will remain for the foreseeable future I'd use a clearer name.

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -123,4 +154,20 @@ export abstract class ClientCore implements IClientCore {
...
+
+  protected calculateGasFeeEstimation(
+    gasLimitEstimationFromCall: Promise<BigNumber>
+  ): Promise<IGasFeeEstimation> {
+    return Promise.all([this.maxFeePerGas, gasLimitEstimationFromCall]).then(
+      data => {
+        return {
+          avg: data[0]
+            .mul(data[1])
+            .div(this.feeEstimationReducer)
```
brickpop: Careful with this. If the reduction factor is `0.625` and we **divide** by that value, we are doing the inverse of a reduction

javascript/modules/client/src/internal/interfaces/dao.ts
```diff
@@ -128,3 +128,8 @@ export enum VoteOption {
   YEA,
   NAY,
 }
+
+export interface IGasFeeEstimation {
+  avg: BigNumber;
```
brickpop: Being a general purpose SDK, I would tend to use clearer words
```suggestion
  average: BigNumber;
```

javascript/modules/client/test/unit/client.test.ts
```diff
@@ -182,7 +182,10 @@ describe("Client instances", () => {
 
     const gasFeesEstimation = await client.estimate.create(daoCreationParams);
 
-    expect(BigNumber.isBigNumber(gasFeesEstimation)).toBeTruthy();
+    expect(typeof gasFeesEstimation).toEqual("object");
+    expect(BigNumber.isBigNumber(gasFeesEstimation.avg)).toBeTruthy();
+    expect(BigNumber.isBigNumber(gasFeesEstimation.max)).toBeTruthy();
+    expect(gasFeesEstimation.max.gt(gasFeesEstimation.avg)).toBeTruthy();
```
brickpop: I would definitely add a test to check that the average value is lower than the max one. This should surface that issue I commented above

javascript/modules/client/test/unit/client.test.ts
```diff
@@ -262,7 +265,10 @@ describe("Client instances", () => {
 
     const gasFeesEstimation = await client.estimate.create(daoCreationParams);
 
-    expect(BigNumber.isBigNumber(gasFeesEstimation)).toBeTruthy();
+    expect(typeof gasFeesEstimation).toEqual("object");
```
brickpop, line 268: Same as above

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -68,6 +77,28 @@ export abstract class ClientCore implements IClientCore {
...
+  }
+
+  get feeEstimationReducer(): BigNumber {
+    let estimationReducer = 100;
+    if (
+      this._gasFeeReducer &&
+      this._gasFeeReducer > 0 &&
+      this._gasFeeReducer < 1
+    ) {
+      estimationReducer = Math.trunc((1 / this._gasFeeReducer) * 100);
```
brickpop: Related to the message about dividing, this feels confusing to me.
I would expect things to be as simple as `fee = maxFee * avgFactor` and that's it. The double division feels counter intuitive.

brickpop: In addition, we should only allow to execute valid cases only.
Meaning:
- If `_gasFeeReducer` is not set, we return a factor of `1`.
- If `_gasFeeReducer` is out of `[0, 1]` then we throw an error
- Otherwise, we return a multiplicable value

javascript/modules/client/test/unit/client.test.ts
marcvelmer: It's already done in line 188

javascript/modules/client/src/internal/client-core.ts
marcvelmer: Not exactly. Because we are dealing with `BigNumber` (and therefore we cannot multiply neither divide by floats), the return of `this.feeEstimationReducer` is the factor multiplied by 100 and this is the value we use to divide. And that's why on the next line we multiply by 100, to keep the factor correctly calculated.

javascript/modules/client/src/internal/client-core.ts
marcvelmer: > I would expect things to be as simple as fee = maxFee * avgFactor and that's it. The double division feels counter intuitive.

It was also for me initially, the problem comes when dealing with `BigNumber` and to avoid losing accuracy by converting this to another primitive type. Therefore the only chance I got was converting the factor to `BigNumber` also, and that's why it seems confusing.



> In addition, we should only allow to execute valid cases only.
> Meaning:
> 
>     If _gasFeeReducer is not set, we return a factor of 1.
>     If _gasFeeReducer is out of [0, 1] then we throw an error
>     Otherwise, we return a multiplicable value

It's already being done, the only difference I see is the error throwing, which LGTM, because it means that an invalid value has been set. Right now it just converts it to `1`.

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -13,11 +13,13 @@ import { IDAO } from "@aragon/core-contracts-ethers";
 import { BigNumber, BigNumberish } from "@ethersproject/bignumber";
 
 export abstract class ClientCore implements IClientCore {
+  private static readonly PRECISION_FACTOR = 1000;
```
brickpop:
```suggestion
  private static readonly PRECISION_FACTOR_BASE = 1000;
```

javascript/modules/client/src/context.ts
```diff
@@ -134,6 +138,14 @@ export class Context {
     }
   }
 
+  useGasFeeEstimationFactor(gasFeeEstimationFactor: number): number {
```
brickpop: This feels still confusing. I'd name it something like `resolveXXX`.... but why doing this here, when we could intercept the error upon initialization?

javascript/modules/client/src/client-dao.ts
```diff
@@ -142,7 +142,7 @@ export class ClientDaoERC20Voting extends ClientCore
         ...ClientDaoERC20Voting.createDaoParameters(params)
       );
 
-      return this.calculateGasFeeEstimation(gasLimit);
+      return this.estimateGasFee(gasLimit);
```
brickpop, line 145:
```suggestion
      return this.estimateGasFromLimit(gasLimit);
```

javascript/modules/client/src/context.ts
```diff
@@ -56,7 +56,7 @@ export class Context {
       throw new Error("No DAO address defined");
     } else if (!contextParams.web3Providers) {
       throw new Error("No web3 endpoints defined");
-    } else if (!contextParams.gasFeeReducer) {
+    } else if (!contextParams.gasFeeEstimationFactor) {
       throw new Error("No gas fee reducer defined");
```
brickpop, line 62: **I would probably make this optional, indeed....**

```suggestion
      throw new Error("No gas fee estimation factor");
```

javascript/modules/client/src/context.ts
```diff
@@ -72,7 +72,9 @@ export class Context {
         contextParams.web3Providers,
         contextParams.network
       ),
-      gasFeeReducer: contextParams.gasFeeReducer,
+      gasFeeEstimationFactor: this.useGasFeeEstimationFactor(
```
brickpop: I would use something like `resolveXXX` or similar.
Why don't we have the constant here instead?

javascript/modules/client/src/internal/client-core.ts
brickpop: Shouldn't this be part of the Context? We are defining the values there... the client is simply a consumer of the factor and the base

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -155,18 +145,24 @@ export abstract class ClientCore implements IClientCore {
...
+              BigNumber.from(
+                Math.trunc(
+                  this._gasFeeEstimationFactor * ClientCore.PRECISION_FACTOR
+                )
+              )
+            )
+            .div(BigNumber.from(ClientCore.PRECISION_FACTOR)),
           max: data[0].mul(data[1]),
-        };
+        } as IGasFeeEstimation;
```
brickpop: Do we need this, given the function signature?

javascript/modules/client/src/context.ts
marcvelmer: This function is used in the initialization and where the check of the values is done, exactly like for the `web3providers`. It has to be changed?

javascript/modules/client/src/context.ts
marcvelmer: It's already optional, here you are in the `setFull` function which makes sense to check if the value exists.

brickpop: I know but even on the setFull method, I feel like the estimation factor is something unrelated to the UI itself, then could go 100% optional

javascript/modules/client/src/context.ts
```diff
@@ -275,4 +269,14 @@ export class Context {
   static getDefault() {
     return defaultState;
   }
+
+  // INTERNAL HELPERS
+
+  private static resolveGasFeeEstimationFactor(gasFeeEstimationFactor: number): number {
+    if (typeof gasFeeEstimationFactor === "undefined") return 1;
+    else if (typeof gasFeeEstimationFactor !== "number" || gasFeeEstimationFactor < 0 || gasFeeEstimationFactor > 1) {
```
marcvelmer: `typeof gasFeeEstimationFactor !== "number"` is not needed here, it's for sure a `number`, right?

brickpop: True. It is a more precise version of the `if(!gasFeeEstimationFactor)`, indeed I was about to add the `?` to the parameter
But then, I feel like this function is doing nothing. Still, I would add the number range check on the setters and get rid of this

brickpop (approved):

PR #17, opened 2022-05-17
Feature: Add IPFS integration

josemarinas:
## Description

Create an IPFS connection and add methods: 

- Pin
- FetchString
- FetchBytes

Task: [APP-317](https://aragonassociation.atlassian.net/browse/APP-317)

## Type of change

## Checklist:

- [ ] I have updated the ``CHANGELOG.md`` file in the root folder of the package after the [TBD] title and before the latest version.


marcvelmer (commented):
Before going deep in to the code:

1) The testing is not working in the CI, this should be reviewed.
2) Did you lint the code? It doesn't seem so.
3) Did you rebase the code with `develop`?

josemarinas:
> Before going deep in to the code:
> 
>     1. The testing is not working in the CI, this should be reviewed.
> 
>     2. Did you lint the code? It doesn't seem so.
> 
>     3. Did you rebase the code with `develop`?

The code is not rebased yet, I fixed the lint but the build is failing due to the node version, the environment uses node 14 but the `ipfs-http-client` needs node 15 or higher, I updated the `package.json` to require at leat this version

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
       }
     );
   }
+  public async pin(input: string | Uint8Array): Promise<string> {
```
brickpop, line 319:
```suggestion
  public pin(input: string | Uint8Array): Promise<string> {
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+    if (!this.ipfs)
+      return Promise.reject(new Error("IPFS client is not initialized"));
+    return this.ipfs
+      .add(input)
+      .then(res => {
+        return Promise.resolve(res.path);
+      })
```
brickpop:
```suggestion
      .then(res => res.path)
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+      .catch(e => {
+        return Promise.reject(e);
+      });
```
brickpop:
```suggestion
      .catch(e => new Error("Could not pin data: " + (e?.message || "")));
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+  }
+  public async fetchBytes(cid: string) {
+    if (!this.ipfs)
+      return Promise.reject(new Error("IPFS client is not initialized"));
```
brickpop:
```suggestion
      throw new Error("IPFS client is not initialized");
```

brickpop: An async function will reject the base promise when an error is thrown

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+    try {
+      for await (const chunk of this.ipfs.cat(cid)) {
+        return Promise.resolve(chunk);
```
brickpop: It looks like this is only going to return the first chunk, but we should definitely check on big files... and see if all chunks need to be serialized into one

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+      }
+    } catch (e) {
+      Promise.reject(e);
```
brickpop: We're not returning anything here.
I would personally throw an Error of ourselves

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+    }
+    return;
```
brickpop:
```suggestion
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+  }
+  public fetchString(cid: string): Promise<string> {
+    return this.fetchBytes(cid)
+      .then(bytes => {
+        return new TextDecoder().decode(bytes);
+      })
```
brickpop:
```suggestion
      .then(bytes => new TextDecoder().decode(bytes))
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -166,4 +175,37 @@ export abstract class ClientCore implements IClientCore {
...
+      .catch(e => {
+        return Promise.reject(e);
+      });
```
brickpop: This code has no effect, I would either remove the catch block or throw a specific error.... but since nothing specific can fail here, I'd simply transparently relay the eventual error error

javascript/modules/client/src/internal/interfaces/client-core.ts
```diff
@@ -15,6 +15,7 @@ export interface IClientVocdoniCore {
 }
 export interface IClientIpfsCore {
   // Add here
+  pin(entry: string | Uint8Array): Promise<string>;
```
brickpop: Aren't we missing the interaces for fetching as well?

javascript/modules/client/src/internal/interfaces/context.ts
```diff
@@ -1,5 +1,6 @@
 import { Signer } from "@ethersproject/abstract-signer";
 import { JsonRpcProvider, Networkish } from "@ethersproject/providers";
+import { Options, IPFSHTTPClient } from "ipfs-http-client";
```
brickpop, line 3:
```suggestion
import { Options as IPFSOptions, IPFSHTTPClient } from "ipfs-http-client";
```

javascript/modules/client/src/internal/client-core.ts
josemarinas: The function is not async, so I need to return a `Promise.reject`, this will force me to change the signature of the function

How about this?
```js
.catch(e => Promise.reject(new Error("Could not pin data: " + (e?.message || ""))));
```

javascript/modules/client/src/internal/client-core.ts
josemarinas: If i remove this, TS will give me this warning:
```
Not all code paths return a value.ts(7030)
```
I'm not sure why

marcvelmer (commented):
@josemarinas 

We should have some kind of unit testing and it's pretty important at this point, because we will have issues with the usage of `TextDecoder` and `jest`. I already faced that problem.

So because it's a draft PR I would simply put any dummy test in order to tackle this problem and then we can redo it later.

FYI this helped me a lot:
```ts
const { TextEncoder, TextDecoder } = require('util');
global.TextEncoder = TextEncoder;
global.TextDecoder = TextDecoder;
```

javascript/modules/client/src/internal/client-core.ts
josemarinas: I tried with small files and all the chunks get serialized, but I'm not sure with big files

marcvelmer: This part of the code is wrong for sure, we cannot return in the middle of a loop. Can you test with a big file to check what happens? Somehow we should store this into chunks and then create a `Buffer` from it before resolving.

javascript/modules/client/src/internal/client-core.ts
brickpop: The external function is not async, but whatever that happens inside the `then`/`catch` inner functions is handled accordingly. 
If you `throw new Error()` inside a `then` or `catch` clause is going to `reject` the external promise being returned

javascript/modules/client/src/internal/client-core.ts
josemarinas: Tested with a big file and only returned the first chunk, fixed it in the last commit

javascript/modules/client/src/internal/client-core.ts
josemarinas: solved when fixed the big files problem

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -244,4 +253,46 @@ export abstract class ClientCore implements IClientCore {
...
+      for await (const chunk of this.ipfs.cat(cid)) {
+        chunks.push(chunk);
+        totalArrayLength += chunk.length;
+      }
+      var mergedArray = new Uint8Array(totalArrayLength);
+      let lastIndex = 0;
+      chunks.forEach(chunk => {
+        mergedArray.set(chunk, lastIndex);
+        lastIndex += chunk.length;
+      });
```
brickpop: It's a bit more efficient to avoid creating new function contexts for every loop iteration

```suggestion
      for (const chunk of chunks) {
        mergedArray.set(chunk, lastIndex);
        lastIndex += chunk.length;
      }
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -244,4 +253,46 @@ export abstract class ClientCore implements IClientCore {
...
+  public async fetchBytes(cid: string) {
+    if (!this.ipfs) throw new Error("IPFS client is not initialized");
+    try {
+      let chunks: Uint8Array[] = [];
+      let totalArrayLength = 0;
+      for await (const chunk of this.ipfs.cat(cid)) {
+        chunks.push(chunk);
+        totalArrayLength += chunk.length;
+      }
+      var mergedArray = new Uint8Array(totalArrayLength);
```
brickpop:
```suggestion
      const mergedArray = new Uint8Array(totalArrayLength);
```

javascript/modules/client/src/internal/interfaces/context.ts
```diff
@@ -20,7 +16,8 @@ interface VocdoniContextParams {
   // Vocdoni
 }
 interface IpfsContextParams {
-  // ipfs: IpfsOptions;
+  ipfsOptions?: Options;
+  ipfsPeers?: string[];
```
brickpop: Do we need it?

javascript/modules/client/src/internal/interfaces/context.ts
brickpop: I would rename the usage of `Options` into something like `IpfsOptions`. Otherwise the typing becomes ambiguous to someone consuming the library.

javascript/modules/client/src/internal/interfaces/context.ts
```diff
@@ -20,7 +16,8 @@ interface VocdoniContextParams {
   // Vocdoni
 }
 interface IpfsContextParams {
-  // ipfs: IpfsOptions;
+  ipfsOptions?: Options;
```
brickpop: Shouldn't this parameter be mandatory within the IPFS scope? Also, shouldn't we allow for receiving N >= 1 nodes info?

brickpop:
@marcvelmer I'd be careful about the `TextEncoder` workaround. Having this on the testing is fine, but having this on the main source code is going to work on Node but fail on web browsers

javascript/modules/client/src/internal/interfaces/context.ts
josemarinas: If not specified, it will default to the lib default parameters that will try to connect to `localhost:5001`. We can overwrite these parameters ones, if not specified, set the default nodes to the aragon ones. About the having more than one node configuration makes sense, but I believe we should add a way to check the state of the IPFS node and In case that it is disconnected shift to the second one and the first selection between the available nodes should be random to avoid connecting always to the same, this does not make sense if Mathias plans to add a load balancer before the nodes.

javascript/modules/client/src/context.ts
```diff
@@ -1,12 +1,18 @@
...
 import { UnsupportedProtocolError } from "@aragon/sdk-common";
-// import { create as ipfsCreate, Options as IpfsOptions } from "ipfs-http-client";
+import {
+  create as ipfsCreate,
+  IPFSHTTPClient,
+  Options,
+} from "ipfs-http-client";
+// const { TextEncoder, TextDecoder } = require("util");
+// global.TextEncoder = TextEncoder;
+// global.TextDecoder = TextDecoder;
```
brickpop: Do we need to keep this now?

javascript/modules/client/src/context.ts
```diff
@@ -60,10 +65,9 @@ export class Context {
       throw new Error("No web3 endpoints defined");
     } else if (!contextParams.gasFeeEstimationFactor) {
       throw new Error("No gas fee reducer defined");
+    } else if (!contextParams.ipfsNodes) {
```
```suggestion
    } else if (!contextParams.ipfsNodes?.length) {
```

javascript/modules/client/src/context.ts
```diff
@@ -106,9 +110,9 @@ export class Context {
         contextParams.gasFeeEstimationFactor
       );
     }
-    // if (contextParams.ipfs) {
-    //   this.state.ipfs = ipfsCreate(contextParams.ipfs);
-    // }
+    if (contextParams.ipfsNodes) {
```
```suggestion
    if (contextParams.ipfsNodes?.length) {
```

javascript/modules/client/src/context.ts
```diff
@@ -140,6 +144,17 @@ export class Context {
     }
   }
 
+  useIpfsNodes(nodes: Options[]): IPFSHTTPClient[] {
+    if (nodes.length > 0) {
```
brickpop: Should we accept a 0 length array? I'd throw an error

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -26,8 +27,14 @@ export abstract class ClientCore implements IClientCore {
   private _signer: Signer | undefined;
   private _daoFactoryAddress = "";
   private _gasFeeEstimationFactor = 1;
+  private _ipfs: IPFSHTTPClient[] = [];
+  private _ipfsIdx: number = -1;
 
   constructor(context: Context) {
+    if (context.ipfs) {
+      this._ipfs = context.ipfs;
+      this._ipfsIdx = Math.floor(Math.random() * context.ipfs.length);
```
brickpop: If the array had 0 elements, we would have an index of `0`, which doesn't exist.
I'd throw an error if so

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -55,11 +62,18 @@ export abstract class ClientCore implements IClientCore {
     if (!signer) {
       throw new Error("Empty wallet or signer");
     }
-
     this._signer = signer;
     return this;
   }
 
+  public shiftIpfsNode() {
+    if (!this._ipfs.length) throw new Error("No endpoints");
```
brickpop: Might have not been initialized yet

```suggestion
    if (!this._ipfs?.length) throw new Error("No IPFS endpoints are available");
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -55,11 +62,18 @@ export abstract class ClientCore implements IClientCore {
...
+    else if (this._ipfs.length <= 1) {
```
brickpop:
```suggestion
    else if (this._ipfs?.length < 2) {
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -103,13 +117,29 @@ export abstract class ClientCore implements IClientCore {
     return this._daoFactoryAddress;
   }
 
+  get ipfs() {
+    return this._ipfs;
+  }
+
+  get ipfsIdx() {
```
brickpop: Getters should have friendly names. How about `ipfsIndex`?
Indeed... do we really need to export this? Consumers care about getting a client, not about an index. See above

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -103,13 +117,29 @@ export abstract class ClientCore implements IClientCore {
     return this._daoFactoryAddress;
   }
 
+  get ipfs() {
+    return this._ipfs;
```
brickpop: I'd return a usable one, not just all.

```suggestion
    if (!this._ipfs[this._ipfsIdx]) throw new Error("No IPFS endpoints available");
    return this._ipfs[this._ipfsIdx];
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -244,4 +274,62 @@ export abstract class ClientCore implements IClientCore {
...
+    if (!this.ipfs)
+      return Promise.reject(new Error("IPFS client is not initialized"));
+    // find online node
+    let isOnline = false;
+    for (var i = 0; i < this.ipfs.length; i++) {
+      isOnline = this.checkIpfsStatus();
+      if (isOnline) break;
+      this.shiftIpfsNode();
+    }
+    if (!isOnline) throw new Error("IPFS nodes are not available");
```
brickpop:
```suggestion
    if (!isOnline) throw new Error("No IPFS nodes are available");
```

javascript/modules/client/src/internal/client-core.ts
```diff
@@ -244,4 +274,62 @@ export abstract class ClientCore implements IClientCore {
...
+  public async fetchBytes(cid: string) {
+    if (!this.ipfs) throw new Error("IPFS client is not initialized");
+    // find online node
+    let isOnline = false;
+    for (var i = 0; i < this.ipfs.length; i++) {
+      isOnline = this.checkIpfsStatus();
+      if (isOnline) break;
+      this.shiftIpfsNode();
+    }
+    if (!isOnline) throw new Error("IPFS nodes are not available");
```
brickpop:
```suggestion
    if (!isOnline) throw new Error("No IPFS nodes are available");
```

javascript/modules/client/src/context.ts
josemarinas: I think so, the library have default values if not specified pointing to localhost, so we can keep this behavior, or force the user to always send a configuration, so this array will never be 0

brickpop: I would not expect `localhost` to be available. If there is an empty array, I would throw that error

brickpop (approved):

PR #18, opened 2022-05-17
`withdraw` action generation

marcvelmer:
## Description

Task: [ID]()

## Type of change

## Checklist:

- [x] I have updated the ``CHANGELOG.md`` file in the root folder of the package after the [TBD] title and before the latest version.

### DIFF
--- a/javascript/modules/client/src/internal/client-core.ts
+++ b/javascript/modules/client/src/internal/client-core.ts
@@ -260,7 +260,7 @@ export abstract class ClientCore implements IClientCore {
 
   protected static createWithdrawAction(
     to: string,
-    value: bigint,
+    value: number | bigint,
     params: IWithdraw
   ): IProposalAction {
     const data = ClientCore.createWithdrawActionData(params);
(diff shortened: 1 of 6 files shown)


javascript/modules/client/src/internal/client-core.ts
```diff
@@ -244,4 +257,21 @@ export abstract class ClientCore implements IClientCore {
...
+    to: string,
+    value: bigint,
+    params: IWithdraw
+  ): IProposalAction {
+    const data = ClientCore.createWithdrawActionData(params);
+    return { to, value, data };
+  }
+
+  protected static createWithdrawActionData(params: IWithdraw): string {
+    const daoInterface = DAO__factory.createInterface();
```
brickpop, line 271: Does this payload need to be generated by calling the smart contract?
For now it might be OK, but if the very same computation could be done on the client side, then I'd manually do the ABI encode packed thing that the contract does.... or create a task to do it in the future

javascript/modules/client/src/internal/interfaces/dao.ts
```diff
@@ -120,7 +123,7 @@ export interface ICreateProposal {
 
 export interface IProposalAction {
   to: string;
-  value: number;
+  value: bigint;
```
brickpop, line 121: Might we accept `number | bigint` and then make sure to always convert to bigint?
"No" is a great answer too

```suggestion
  value: number | bigint;
```

brickpop (approved):

PR #36, opened 2022-06-28
Feature: Dashboard read operations

Fabricevladimir:
## Description
- includes high level methods for dashboard read operations (uses dummy data)

**NOTE:** Merging into `future-proof-code-layout-mock` instead of `develop` because these operations do not block web-app

Task: [APP-589](https://aragonassociation.atlassian.net/browse/APP-589)

## Type of change

## Checklist:

### DIFF
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -275,7 +275,16 @@ for await (const step of client.methods.createProposal(
 
 ### Loading the list of members (ERC20)
 
-- **TODO**
+Retrieving all the members of an ERC20 DAO.
+
+```ts
+import { ClientErc20 } from "@aragon/sdk-client";
+
+const client = new ClientERC20(context);
+const members = await client.methods.getDaoMembers();
+
+console.log(members); // ["0x3585...", "0x1235...", "0x6785...",]
+```
 
 ### Loading the list of proposals (ERC20)
 
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -304,47 +304,55 @@ const proposals = await client.methods.getProposals(doaAddressOrEns);
 console.log(proposals);
 
 /*
- [{
+ [ {
     id: "0x56fb7bd9491ff76f2eda54724c84c8b87a5a5fd7_0x0",
-    daoAddress: "0x1234...",
+    daoAddress: "0x56fb7bd9491ff76f2eda54724c84c8b87a5a5fd7",
     daoName: "DAO 1",
-    creator: "0x1234...",
-
-    token: {
-      address: "0x1234...",
-      name: "DAO Token",
-      symbol: "DAO",
-      decimals: 18,
-    },
-
-    title: "New Founding for Lorex Lab SubDao",
-    summary: "As most community members know, Aragon has strived to deploy its products to more...",
-    proposal: "<h1>This is the super <strong>important</strong> proposal body<h1>",
-    resources: [{ url: "https://example.com", description: "Amazing Link" }],
+    creator: "0x8367dc645e31321CeF3EeD91a10a5b7077e21f70",
 
     startDate: <Date>,
     endDate: <Date>,
     createdAt: <Date>,
 
+    title: "New Founding for Lorex Lab SubDao",
+    summary: "As most community members know, Aragon has strived.",
+    proposal: "This is the super important proposal body",
+    resources: [{ url: "https://example.com", description: "Example" }],
+
     voteId: "0",
-    participationRequiredPct: 30,
-    supportRequiredPct: 52,
+    token: {
+      address: "0x9df6870250396e10d187b188b8bd9179ba1a9c18",
+      name: "DAO Token",
+      symbol: "DAO",
+      decimals: 18,
+    },
+
+    result: {
+      yea: 3,
+      nay: 1,
+      abstain: 2,
+    },
+
+    open: false,
+    executed: false,
+    status: "Pending",
+
+    config: {
+      participationRequiredPct: 30,
+      supportRequiredPct: 52,
+    },
+
     votingPower: 135,
 
-    yea: 3,
-    nay: 1,
-    abstain: 2,
-
-    open: true,
-    executed: false,
-    status: ProposalStatus.PENDING,
-    
     voters: [
-      { id: "0x1334...", voterState: VoteOptions.YEA, weight: 1 },
-      ...,
-      { id: "0x1834...", voterState: VoteOptions.ABSTAIN, weight: 1 },
+      {
+        address: "0x8367dc645e31321CeF3EeD91a10a5b7077e21f70",
+        voteValue: VoteOptions.YEA,
+        weight: 1,
+      },
+      {...}
     ],
-  }] */
+  },] */
 ```
 
 ## Multisig governance plugin client
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -188,10 +188,10 @@ import { Client } from "@aragon/sdk-client";
 const client = new Client(context);
 const doaAddressOrEns = "0x1234..."; // unique identifier; dao ENS domain or address
 
-const metadata = await client.methods.getMetadata(daoIdentifier);
+const metadata = await client.methods.getMetadata(doaAddressOrEns);
 console.log(metadata);
 
-/* 
+/*
 {
    address: "0x1234...",
    avatar: "http...",
@@ -250,14 +250,16 @@ const proposalCreationParams: ICreateProposalParams = {
 };
 
 const estimatedGas = await client.estimation.createProposal(
-  proposalCreationParams
+  proposalCreationParams,
 );
 console.log(estimatedGas.average); // bigint
 console.log(estimatedGas.max); // bigint
 
-for await (const step of client.methods.createProposal(
-  proposalCreationParams
-)) {
+for await (
+  const step of client.methods.createProposal(
+    proposalCreationParams,
+  )
+) {
   switch (step.idx) {
     case DaoDepositSteps.CREATING:
       console.log(step.txHash); // 0xb1c14a49...
@@ -290,7 +292,60 @@ console.log(members); // ["0x3585...", "0x1235...", "0x6785...",]
 
 ### Loading the list of proposals (ERC20)
 
-- **TODO**
+Retrieving the proposals of an ERC20 DAO.
+
+```ts
+import { Client } from "@aragon/sdk-client";
+
+const client = new Client(context);
+const doaAddressOrEns = "0x1234...";
+
+const proposals = await client.methods.getProposals(doaAddressOrEns);
+console.log(proposals);
+
+/*
+ [{
+    id: "0x56fb7bd9491ff76f2eda54724c84c8b87a5a5fd7_0x0",
+    daoAddress: "0x1234...",
+    daoName: "DAO 1",
+    creator: "0x1234...",
+
+    token: {
+      address: "0x1234...",
+      name: "DAO Token",
+      symbol: "DAO",
+      decimals: 18,
+    },
+
+    title: "New Founding for Lorex Lab SubDao",
+    summary: "As most community members know, Aragon has strived to deploy its products to more...",
+    proposal: "<h1>This is the super <strong>important</strong> proposal body<h1>",
+    resources: [{ url: "https://example.com", description: "Amazing Link" }],
+
+    startDate: <Date>,
+    endDate: <Date>,
+    createdAt: <Date>,
+
+    voteId: "0",
+    participationRequiredPct: 30,
+    supportRequiredPct: 52,
+    votingPower: 135,
+
+    yea: 3,
+    nay: 1,
+    abstain: 2,
+
+    open: true,
+    executed: false,
+    status: ProposalStatus.PENDING,
+    
+    voters: [
+      { id: "0x1334...", voterState: VoteOptions.YEA, weight: 1 },
+      ...,
+      { id: "0x1834...", voterState: VoteOptions.ABSTAIN, weight: 1 },
+    ],
+  }] */
+```
 
 ## Multisig governance plugin client
 
@@ -316,7 +371,8 @@ console.log(members); // ["0x3585...", "0x1235...", "0x6785...",]
 
 ## Action encoders
 
-Proposals will eventually need to execute some action on behalf of the DAO, which needs to be encoded in a low level format.
+Proposals will eventually need to execute some action on behalf of the DAO,
+which needs to be encoded in a low level format.
 
 The helpers above help encoding the most typical DAO operations.
 
@@ -356,19 +412,23 @@ See `ClientCore` ([source](./src/internal/core.ts)):
 
 When updating a `ClientXXX` (plugin) class:
 
-- **Update first** all affected enum's, types and interfaces in `src/internal/interfaces/plugins.ts`
+- **Update first** all affected enum's, types and interfaces in
+  `src/internal/interfaces/plugins.ts`
 
 When updating the `Client` class:
 
-- **Update first** all affected enum's, types and interfaces in `src/internal/interfaces/client.ts`
+- **Update first** all affected enum's, types and interfaces in
+  `src/internal/interfaces/client.ts`
 
 When updating the `ClientCore` class:
 
-- **Update first** all affected enum's, types and interfaces in `src/internal/interfaces/core.ts`
+- **Update first** all affected enum's, types and interfaces in
+  `src/internal/interfaces/core.ts`
 
 ## Developing a new Plugin client
 
-Create a new class that `extends` from `ClientCore`, receives a `Context` on the `constructor` and follows the structure of [ClientErc20](./src/client-erc20.ts).
+Create a new class that `extends` from `ClientCore`, receives a `Context` on the
+`constructor` and follows the structure of [ClientErc20](./src/client-erc20.ts).
 
 # Testing
 
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -195,12 +195,11 @@ console.log(metadata);
 {
    address: "0x1234...",
    avatar: "http...",
-   createdAt: 1656411653742,
+   createdAt: <Date>,
    description: "This dao...",
    links: [{label: "Website", url: "http..."}],
    name: "Abc Dao",
-   packages: ["0x1245..."],
-   token: {address: '0x123...', decimals: 18, name: "Test Token", symbol: "TKN"};
+   packages: ["0x1245...", "0x3456..."],
 } */
 ```
 
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -186,9 +186,9 @@ Handles retrieving DAO metadata using its address or ENS domain.
 import { Client } from "@aragon/sdk-client";
 
 const client = new Client(context);
-const daoIdentifier = "0x1234..."; // unique identifier; dao ENS domain or address
+const doaAddressOrEns = "0x1234..."; // unique identifier; dao ENS domain or address
 
-const metadata = await client.methods.getDaoMetadata(daoIdentifier);
+const metadata = await client.methods.getMetadata(daoIdentifier);
 console.log(metadata);
 
 /* 
@@ -281,7 +281,9 @@ Retrieving all the members of an ERC20 DAO.
 import { ClientErc20 } from "@aragon/sdk-client";
 
 const client = new ClientERC20(context);
-const members = await client.methods.getMembers();
+const doaAddressOrEns = "0x1234..."; // unique identifier; dao ENS domain or address
+
+const members = await client.methods.getMembers(daoAddressOrEns);
 
 console.log(members); // ["0x3585...", "0x1235...", "0x6785...",]
 ```
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -1,4 +1,5 @@
-## Aragon JS SDK Client
+Aragon JS SDK Client
+---
 
 @aragon/sdk-client provides easy access to the high level interactions to be
 made with an Aragon DAO. It consists of three different components:
@@ -196,9 +197,9 @@ console.log(metadata);
    avatar: "http...",
    createdAt: 1656411653742,
    description: "This dao...",
-   links: [{label: "Website", url: "http..."}];
+   links: [{label: "Website", url: "http..."}],
    name: "Abc Dao",
-   packages: ["ERC20VotingPackage"];
+   packages: ["0x1245..."],
    token: {address: '0x123...', decimals: 18, name: "Test Token", symbol: "TKN"};
 } */
 ```
@@ -281,7 +282,7 @@ Retrieving all the members of an ERC20 DAO.
 import { ClientErc20 } from "@aragon/sdk-client";
 
 const client = new ClientERC20(context);
-const members = await client.methods.getDaoMembers();
+const members = await client.methods.getMembers();
 
 console.log(members); // ["0x3585...", "0x1235...", "0x6785...",]
 ```
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -185,7 +185,7 @@ Handles retrieving DAO metadata using unique identifier.
 import { Client } from "@aragon/sdk-client";
 
 const client = new Client(context);
-const daoIdentifier = "0x1234..."; // unique identifier; dao name or address
+const daoIdentifier = "0x1234..."; // unique identifier; dao ENS domain or address
 
 const metadata = await client.methods.getDaoMetadata(daoIdentifier);
 console.log(metadata);
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -180,7 +180,7 @@ for await (const step of client.methods.deposit(depositParams)) {
 
 ### Loading DAO details
 
-Handles retrieving DAO metadata using unique identifier.
+Handles retrieving DAO metadata using its address or ENS domain.
 
 ```ts
 import { Client } from "@aragon/sdk-client";
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -1,5 +1,4 @@
-Aragon JS SDK Client
----
+## Aragon JS SDK Client
 
 @aragon/sdk-client provides easy access to the high level interactions to be
 made with an Aragon DAO. It consists of three different components:
@@ -139,7 +138,7 @@ Handles the flow of depositing ERC20 tokens to a DAO.
 - Similar to the example above
 - The `tokenAddress` field is required
 - Will attempt to increase the ERC20 allowance if not sufficient
-- More fintermediate steps are yielded
+- More intermediate steps are yielded
 
 ```ts
 import { Client, DaoDepositSteps, IDepositParams } from "@aragon/sdk-client";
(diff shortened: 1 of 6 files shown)


javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+      .then(gasLimit => {
+        return this.web3.getApproximateGasFee(gasLimit.toBigInt());
+      });
+  }
+
+  //// PRIVATE METHODS METADATA
+
+  private async _getDaoMetadata(daoIdentifier: string): Promise<DaoMetadata> {
+    // TODO: Implement actual fetch logic using subgraph.
+    // Note: it would be nice if the client could be instantiated with dao identifier
```
githubdoramon: Maybe this would be bad if for any reason you need to fetch info from 2 DAOs at the same time

Fabricevladimir: True story, true story

javascript/modules/client/README.md
```diff
@@ -1,5 +1,4 @@
-Aragon JS SDK Client
----
+## Aragon JS SDK Client
```
brickpop: Why this change?

This is the markdown document title, if we start using `#` or `##`, then everything else needs to be `###` or more

javascript/modules/client/README.md
```diff
@@ -180,16 +179,37 @@ for await (const step of client.methods.deposit(depositParams)) {
...
 ### Loading DAO details
 
-- __TODO__:  Name, metadata, installed plugin list
+Handles retrieving DAO metadata using unique identifier.
+
+```ts
+import { Client } from "@aragon/sdk-client";
+
+const client = new Client(context);
+const daoIdentifier = "0x1234..."; // unique identifier; dao name or address
```
```suggestion
const daoIdentifier = "0x1234..."; // unique identifier; dao ENS domain or address
```

javascript/modules/client/README.md
```diff
@@ -180,16 +179,37 @@ for await (const step of client.methods.deposit(depositParams)) {
...
+
+const metadata = await client.methods.getDaoMetadata(daoIdentifier);
+console.log(metadata);
+
+/* 
+{
+   address: "0x1234...",
+   avatar: "http..."
```
```suggestion
   avatar: "http...",
```

javascript/modules/client/README.md
```diff
@@ -180,16 +179,37 @@ for await (const step of client.methods.deposit(depositParams)) {
...
+   createdAt: 1656411653742,
```
brickpop: I would expect this to be a JS `Date()` as opposed to a simple number

javascript/modules/client/README.md
```diff
@@ -180,16 +179,37 @@ for await (const step of client.methods.deposit(depositParams)) {
...
+   description: "This dao...",
+   links: [{label: "Website", url: "http..."}];
+   name: "Abc Dao",
+   packages: ["ERC20VotingPackage"];
```
brickpop: That will probably change in the future, but for now, I would return the registry ID of the plugins installed
```suggestion
   packages: ["0x1234..."],
```

javascript/modules/client/README.md
```diff
@@ -180,16 +179,37 @@ for await (const step of client.methods.deposit(depositParams)) {
...
+   token: {address: '0x123...', decimals: 18, name: "Test Token", symbol: "TKN"};
```
brickpop: The token is no longer something generic to the DAO. A specific plugin like an ERC20 voting plugin may create its own. Probably we'd need feedback from @Rekard0 or @Michael-A-Heuer

javascript/modules/client/README.md
```diff
@@ -249,42 +271,50 @@ for await (
...
 ### Loading the list of members (ERC20)
 
-- __TODO__
+Retrieving all the members of an ERC20 DAO.
+
+```ts
+import { ClientErc20 } from "@aragon/sdk-client";
+
+const client = new ClientERC20(context);
+const members = await client.methods.getDaoMembers();
```
```suggestion
const members = await client.methods.getMembers();
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -53,6 +53,7 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
       this._setDaoConfig(address, config),
     setVotingConfig: (address: string, config: VotingConfig) =>
       this._setVotingConfig(address, config),
+    getDaoMembers: (address: string) => this._getDaoMembers(address),
```
brickpop:
```suggestion
    getMembers: (address: string) => this._getMembers(address),
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -163,46 +161,56 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+    return Promise.resolve({ average: BigInt(0), max: BigInt(0) });
   }
   private _estimateSetVotingConfig(daoAddress: string, config: VotingConfig) {
     // TODO: Unimplemented
-    return Promise.resolve(BigInt(0));
+    return Promise.resolve({ average: BigInt(0), max: BigInt(0) });
+  }
+
+  //// PRIVATE RETRIEVAL HANDLERS
+  private async _getDaoMembers(daoAddress: string) {
```
brickpop: If we're wrapping the result within a `Promise.resolve()`, there is not need for async yet.

```suggestion
  private _getDaoMembers(daoAddress: string) {
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -163,46 +161,56 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+    if (!daoAddress) {
+      throw new Error("A DAO address is needed");
```
brickpop:
```suggestion
      throw new Error("Invalid DAO address or ENS");
```

javascript/modules/client/src/client-erc20.ts
brickpop: Also, for consistency, I would expect `daoIdentifier` as stated on the readme (address / ENS)

javascript/modules/client/src/client.ts
```diff
@@ -17,7 +18,7 @@ import { BigNumber } from "@ethersproject/bignumber";
 import { AddressZero } from "@ethersproject/constants";
 import { ClientCore } from "./internal/core";
 import { DaoRole } from "./internal/interfaces/common";
-import { solidityPack } from "ethers/lib/utils";
+import { solidityPack, isAddress } from "ethers/lib/utils";
```
brickpop: I know you didn't introduce this, but could you fetch these functions from `@ethersproject/xxx`  instead?

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+
+  //// PRIVATE METHODS METADATA
+
+  private async _getDaoMetadata(daoIdentifier: string): Promise<DaoMetadata> {
+    // TODO: Implement actual fetch logic using subgraph.
+    // Note: it would be nice if the client could be instantiated with dao identifier
+    // thereby removing the need to pass in the identifier for all the data retrieval methods
+
+    if (!daoIdentifier) {
+      throw new Error("A DAO identifier is needed");
```
brickpop: Same as above: `Invalid DAO address or ENS` or similar

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+    return daoInstance.estimateGas
+      .deposit(tokenAddress, amount, reference, override)
+      .then(gasLimit => {
+        return this.web3.getApproximateGasFee(gasLimit.toBigInt());
+      });
+  }
+
+  //// PRIVATE METHODS METADATA
+
+  private async _getDaoMetadata(daoIdentifier: string): Promise<DaoMetadata> {
```
brickpop: Same as above, async not needed yet

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+        ? {
+            address: daoIdentifier,
+            name:
+              dummyDaoNames[
+                Math.floor(Math.random() * dummyDaoNames.length - 1)
+              ],
+          }
+        : {
+            address: "0x663ac3c648548eb8ccd292b41a8ff829631c846d",
+            name: daoIdentifier,
```
brickpop: We will not be indexing by DAO name. Indexing will be done by address only. If an ENS is provided, the SDK should resolve the ENS name and use that address. I don't understand the point on returning an input parameter itself

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+          }),
+
+      createdAt: new Date(
+        fromDate + Math.random() * (Date.now() - fromDate)
+      ).getTime(),
```
brickpop: I would return a native `Date`. consumers can always compute the timestamp themselves

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+        {
+          label: "Website",
+          url: "https://google.com",
+        },
+        {
+          label: "Discord",
+          url: "https://google.com",
+        },
+      ],
+      packages: ["WhitelistPackage", "ERC20VotingPackage"],
```
brickpop: I would return some 32 byte hex string

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+      token: {
+        address: "0x0000000000000000000000000000000000000000",
+        name: "Ethereum (Canonical)",
+        symbol: "ETH",
+        decimals: 18,
+      },
```
brickpop: Same as above, not sure about an ERC20 token being an opinionated global thing here. A DAO could also have an ERC721 (NFT) governance plugin and/or not have any token at all

brickpop: It should be up to each plugin's client to provide such information

javascript/modules/client/src/client.ts
```diff
@@ -282,32 +277,93 @@ export class Client extends ClientCore implements IClient {
...
+    "0x" +
+    params.plugins
+      .map(entry => {
+        const item = solidityPack(
+          ["uint256", "bytes[]"],
+          [entry.id, entry.data]
+        );
+        return strip0x(item);
+      })
+      .join("");
```
brickpop: Having conflicting code formatters is an issue. I suggest using `deno fmt` (plugin available). Which is yours?

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -8,15 +8,17 @@ export interface IClient extends IClientCore {
...
       where: string,
       who: string,
       role: DaoRole,
-      data: Uint8Array,
+      data: Uint8Array
     ) => Promise<void>;
     /** Deposits ether or an ERC20 token */
     deposit: (params: IDepositParams) => AsyncGenerator<DaoDepositStepValue>;
+    /** Retrieves metadata for DAO with given identifier*/
+    getDaoMetadata: (daoIdentifier: string) => Promise<DaoMetadata>;
```
brickpop:
```suggestion
    getMetadata: (daoIdentifier: string) => Promise<DaoMetadata>;
```

javascript/modules/client/src/client.ts
```suggestion
  private async _getMetadata(daoIdentifier: string): Promise<DaoMetadata> {
```

javascript/modules/client/src/client.ts
```diff
@@ -35,12 +36,17 @@ export class Client extends ClientCore implements IClient {
     create: (params: ICreateParams) => this._createDao(params),
     /** Deposits ether or an ERC20 token */
     deposit: (params: IDepositParams) => this._deposit(params),
-    /** Checks whether a role is granted by the curren DAO's ACL settings */
+
+    /** Retrieves metadata for DAO with given identifier*/
+    getDaoMetadata: (daoIdentifier: string) =>
+      this._getDaoMetadata(daoIdentifier),
```
```suggestion
    getMetadata: (daoIdentifier: string) =>
      this._getMetadata(daoIdentifier),
```

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -73,3 +75,24 @@ export type DaoDepositStepValue =
...
+
+// DAO DETAILS
+export type DaoResourceLink = { label: string; url: string };
+
+export type DaoToken = {
+  address: string;
+  decimals: number;
+  name: string;
+  symbol: string;
+};
```
brickpop: Given the above points, I'm not sure if this should belong to the ERC20 governance plugin

javascript/modules/client/README.md
Fabricevladimir: Automatically formatted

javascript/modules/client/src/client.ts
Fabricevladimir: I've prettier and eslint. Those are part of the dev guidelines we set way back when

javascript/modules/client/src/client.ts
Fabricevladimir: Because of the types from subgraph, I was under the impression that every DAO had a "DAO token" which seems to default to ETH if not explicitly defined.

javascript/modules/client/src/client.ts
Fabricevladimir: Jordi, it's _dummy_ data. It's simply a way when the function is called with an address to return the data with the same address. Obviously this will change when implemented, no?

javascript/modules/client/src/client.ts
Fabricevladimir: Heads up. I was unable to find the `solidityPack` method from `@ethersproject` for whatever reason. I've replaced this with the `pack` method from `@ehtersproject/solidity` which I believe is the same method.

javascript/modules/client/README.md
```diff
@@ -180,16 +180,36 @@ for await (const step of client.methods.deposit(depositParams)) {
 
 ### Loading DAO details
 
-- __TODO__:  Name, metadata, installed plugin list
+Handles retrieving DAO metadata using unique identifier.
```
```suggestion
Handles retrieving DAO metadata using its address or ENS domain.
```

javascript/modules/client/README.md
```diff
@@ -180,16 +180,36 @@ for await (const step of client.methods.deposit(depositParams)) {
...
+
+```ts
+import { Client } from "@aragon/sdk-client";
+
+const client = new Client(context);
+const daoIdentifier = "0x1234..."; // unique identifier; dao ENS domain or address
+
+const metadata = await client.methods.getDaoMetadata(daoIdentifier);
```
```suggestion
const daoAddressOrEns = "0x1234..."; // unique identifier; dao ENS domain or address

const metadata = await client.methods.getDaoMetadata(daoAddressOrEns);
```

javascript/modules/client/README.md
```diff
@@ -249,42 +271,50 @@ for await (
...
 ### Loading the list of members (ERC20)
 
-- __TODO__
+Retrieving all the members of an ERC20 DAO.
+
+```ts
+import { ClientErc20 } from "@aragon/sdk-client";
+
+const client = new ClientERC20(context);
+const members = await client.methods.getMembers();
```
brickpop: We are not passing any kind of DAO/plugin address. If the DAO/plugin address is passed via the context, then we should show it. If it is a `getMembers` parameter, we should show it as well

javascript/modules/client/src/client-erc20.ts
```diff
@@ -53,6 +53,7 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
       this._setDaoConfig(address, config),
     setVotingConfig: (address: string, config: VotingConfig) =>
       this._setVotingConfig(address, config),
+    getMembers: (daoAddressOrEns: string) => this._getMembers(daoAddressOrEns),
```
brickpop, line 55:
```suggestion
    /** Retrieves the list of Ethereum addresses that can cast weighted votes on the given DAO's proposals */
    getMembers: (daoAddressOrEns: string) => this._getMembers(daoAddressOrEns),
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -189,20 +187,30 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+
+    const mockAddresses = [
+      "0x8367dc645e31321CeF3EeD91a10a5b7077e21f70",
+      "0xDA9dfA130Df4dE4673b89022EE50ff26f6EA73Cf",
+      "0xBE0eB53F46cd790Cd13851d5EFf43D12404d33E8",
+      "0x2dB75d8404144CD5918815A44B8ac3f4DB2a7FAf",
+      "0xc1d60f584879f024299DA0F19Cdb47B931E35b53",
+    ];
+
+    return Promise.resolve(mockAddresses.filter(() => Math.random() > 0.4));
```
brickpop: Not strictly necessary, just FYI adding a small delay can help the UI check that a loading spinner appears while data is not available. Using a simple `setTimeout` could be of use. 

Example:

```ts
return new Promise(resolve => setTimeout(resolve, 1000)).then(() => mockAddresses.filter(() => Math.random() > 0.4));
```

brickpop (approved):

javascript/modules/client/src/client-erc20.ts
```diff
@@ -195,14 +201,27 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
     }
 
     const mockAddresses = [
-      "0x8367dc645e31321CeF3EeD91a10a5b7077e21f70",
+      "0x1234...",
```
brickpop: Why changing to an incomplete address?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -195,14 +201,27 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+      mockAddresses.filter(() => Math.random() > 0.4)
+    );
+  }
+
+  private _getProposals(daoAddressOrEns: string) {
+    if (!daoAddressOrEns) {
+      throw new Error("Invalid DAO address or ENS");
+    }
+
+    setProposalStatus(MOCK_PROPOSALS);
```
brickpop: Modifying an object passed by reference is a bit of an anti pattern. Better to have a new variable containing the result, which is a copy of the original. 

At this stage it doesn't matter, but in general, modifying a global object can have unintended side effects elsewhere

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
     params.reference ?? "",
   ];
 }
+
+/// HELPER FUNCTIONS TODO: move to utils?
+function setProposalStatus(proposals: IErc20VotingProposal[]) {
```
brickpop: I would:
- call it `getProposalsWithStatus`
- Do a copy of the input data (`const result = []`), then at each iteration `push` an `Object.assign({}, proposal, {status: ...})`  to the results array
- Return the new array

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+    creator: "0x1234...",
+
+    startDate,
+    endDate: new Date(startDate.getTime() + 7200000),
+    createdAt: new Date(dateWithinThisYear),
+
+    title: "New Founding for Lorex Lab SubDao",
+    summary:
+      "As most community members know, Aragon has strived to deploy its products to more cost-efficient blockchain networks to facilitate interaction.",
+    proposal: "<h1>This is the super important proposal body<h1>",
```
brickpop: Careful about HTML. We need to decide what we support and what we don't (HTML/markdown/plain). By now, this is an attack vector that allows anyong to inject `<script>` tags

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+    resources: [{ url: "https://example.com", description: "Example" }],
+
+    executed: false,
+    status: ProposalStatus.PENDING,
+
+    voteId: "0",
```
brickpop, line 338: Isn't this redundant with the `id` field at the beginning?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+    token: {
+      address: "0x1234...",
+      name: "DAO Token",
+      symbol: "DAO",
+      decimals: 18,
+    },
+
+    yea: 3,
+    nay: 1,
+    abstain: 2,
```
brickpop: Could we have this within a `results: {}` block?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+
+    open: true,
```
brickpop: What does `open` mean? Isn't this redundant with the `status`?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+    participationRequiredPct: 30,
+    supportRequiredPct: 52,
```
brickpop: Shouldn't this be part of a `config` block maybe?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+    votingPower: 135,
```
brickpop, line 361: Voting power of who?
Is that the census size?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -241,3 +260,79 @@ function unwrapWithdrawParams(
...
+
+    voters: [
+      { id: "0x1334...", voterState: VoteOptions.YEA, weight: 1 },
```
brickpop: - Does `id` refer to the address of the voter?
- Does `voterState` refer to the vote `value`?

javascript/modules/client/README.md
```diff
@@ -249,42 +273,106 @@ for await (
...
+
+    voteId: "0",
+    participationRequiredPct: 30,
+    supportRequiredPct: 52,
+    votingPower: 135,
+
+    yea: 3,
+    nay: 1,
+    abstain: 2,
+
```
brickpop: Does this reflect the Type defined below?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -233,3 +260,79 @@ function unwrapWithdrawParams(
...
+/// MOCK DATA FOR PROPOSALS
+const dateWithinThisYear = new Date(
+  new Date().setFullYear(new Date().getFullYear() - 1)
+).getTime();
+
+const startDate = new Date(
+  dateWithinThisYear + Math.random() * (Date.now() - dateWithinThisYear)
+);
+
+const MOCK_PROPOSALS = [
```
brickpop: We should be using the interface to type check the mock data

javascript/modules/client/src/client-erc20.ts
```diff
@@ -233,3 +260,79 @@ function unwrapWithdrawParams(
...
+    endDate: new Date(startDate.getTime() + 7200000),
+    createdAt: new Date(dateWithinThisYear),
+
+    title: "New Founding for Lorex Lab SubDao",
+    summary:
+      "As most community members know, Aragon has strived to deploy its products to more cost-efficient blockchain networks to facilitate interaction.",
+    proposal: "<h1>This is the super important proposal body<h1>",
+    resources: [{ url: "https://example.com", description: "Example" }],
+
+    executed: false,
```
brickpop: Isn't this redundant, given the `status`?

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: Copy pasta probably

javascript/modules/client/src/internal/interfaces/common.ts
```diff
@@ -41,3 +41,40 @@ export type GasFeeEstimation = {
   average: bigint;
   max: bigint;
 };
+
+/**
+ * Contains the base structure of a proposal
+ */
+export interface IProposal {
```
brickpop: - IMO this should be a strict type, not a flexible interface
- We should use it in the code
- Let's make to mirror the fields defined here correspond to the the logic above

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -133,3 +159,28 @@ export enum ProposalCreationSteps {
...
+  nay?: number;
+  abstain?: number;
+
+  open: boolean;
+  participationRequiredPct: number;
+  supportRequiredPct: number;
+  votingPower: number;
+
+  voters: { id: string; voterState: VoteOptions; weight: number }[];
+}
```
brickpop: As stated above, I would recommend using `types` for strict data and do `... = Proposal & { ... }` 

Also, adapt the modifications suggested before in the code

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -133,3 +159,28 @@ export enum ProposalCreationSteps {
 export type ProposalCreationStepValue =
   | { key: ProposalCreationSteps.CREATING; txHash: string }
   | { key: ProposalCreationSteps.DONE; proposalId: string };
+
+// PROPOSAL RETRIEVAL
+
+export interface IErc20VotingProposal extends IProposal {
```
brickpop:
```suggestion
export interface IErc20Proposal extends IProposal {
```

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -133,3 +159,28 @@ export enum ProposalCreationSteps {
...
+
+  open: boolean;
+  participationRequiredPct: number;
+  supportRequiredPct: number;
+  votingPower: number;
+
+  voters: { id: string; voterState: VoteOptions; weight: number }[];
+}
+
+export type DaoToken = {
```
brickpop:
```suggestion
export type Erc20Token = {
```

Plus renaming where applicable

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: No, they're two different fields. This one is from the ERC20Vote entity.

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: `open` and `executed` both are somewhat redundant. That said I think they should still be included for convenience because they provide a way to check the open and executed status of the proposal much quicker than say something like the following:
```javascript
  if(proposal.status === ProposalStatus.ACTIVE || proposal.status === ProposalStatus.SUCCEEDED) {
     // proposal is open
  }
```

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: Yes to both. Worth renaming perhaps if that wasn't clear enough

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: See above comment concerning `open`

javascript/modules/client/README.md
Fabricevladimir: Yup

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: Good point

javascript/modules/client/src/client-erc20.ts
Fabricevladimir: ~Apparently the voting power is used to give weight to the votes and is useful especially when more than one token can participate to the vote. So essentially voting power of the token itself.~

This is the total amount of token at the moment of proposal creation.

brickpop (approved):

PR #47, opened 2022-07-26
Feature: Add GraphQL client

josemarinas:
## Description

Add GraphQL client to the SDK and test it with `getTransfers` function

Task: [APP-363](https://aragonassociation.atlassian.net/browse/APP-363)

## Type of change

## Checklist:


javascript/modules/client/src/client.ts
```diff
@@ -32,6 +32,8 @@ import { pack } from "@ethersproject/solidity";
 import { strip0x } from "@aragon/sdk-common";
 import { erc20ContractAbi } from "./internal/abi/erc20";
 import { Signer } from "@ethersproject/abstract-signer";
+// import { getENS } from "get-ens";
```
brickpop: Why this commented line? Ethers JS providers do resolve names BTW

javascript/modules/client/src/client.ts
```diff
@@ -330,13 +332,21 @@ export class Client extends ClientCore implements IClient {
...
       throw new Error("Invalid DAO address or ENS");
     }
 
+    const dao = await this.subgraph.getClient()
+      .request(
+        getDaoByAddress,
+        { address: daoAddressOrEns }
+      )
+
+    console.log(JSON.stringify(dao))
```
brickpop: :stop_sign:  :warning:

brickpop: Why this code if it is not being used yet?

javascript/modules/client/src/client.ts
```diff
@@ -404,78 +414,38 @@ export class Client extends ClientCore implements IClient {
...
+      throw new Error("Empty DAO address or ENS");
+    }
+    const provider = this.web3.getProvider()
+    if (!provider) {
+      throw new Error("No provider");
+    }
+    let address = daoAddressOrEns
+    if (!isAddress(daoAddressOrEns)) {
+      const recoveredAddress = await provider.resolveName(daoAddressOrEns)
+      if (!recoveredAddress) {
```
```suggestion
      const resolvedAddress = await provider.resolveName(daoAddressOrEns)
      if (!resolvedAddress) {
```

javascript/modules/client/src/client.ts
```diff
@@ -404,78 +414,38 @@ export class Client extends ClientCore implements IClient {
   }
 
   private async _getTransfers(
-    daoAddressOrEns: string,
-  ): Promise<IAssetTransfers> {
+    daoAddressOrEns: string
+  ): Promise<any> {
+    // TODO: typing on return
+
```
brickpop: IMO, we should keep the types, unless there's a reason. We've made a "promise" to the SDK consumers so far. I'd keep the mock data until we can return real results, and even when we do.... only change the return type if strictly needed.

javascript/modules/client/src/context.ts
```diff
@@ -251,10 +254,23 @@ export class Context {
    *
    * @public
    */
-  get ipfs(): IpfsClient[] | undefined {
+   get ipfs(): IpfsClient[] | undefined {
```
```suggestion
  get ipfs(): IpfsClient[] | undefined {
```

javascript/modules/client/src/internal/core.ts
```diff
@@ -12,6 +12,8 @@ import { Context } from "../context";
 import { GasFeeEstimation } from "./interfaces/common";
 import { Random } from "@aragon/sdk-common";
 import { Client as IpfsClient } from "@aragon/sdk-ipfs";
+import { GraphQLClient } from "graphql-request";
+import { status } from "./queries";
```
brickpop: Could we rename the dependency file as `graphql-queries`?

javascript/modules/client/src/internal/core.ts
```diff
@@ -237,5 +245,21 @@ export abstract class ClientCore implements IClientCore {
     },
   };
 
-  graphql: IClientGraphQLCore = {};
+  subgraph: IClientGraphQLCore = {
+    getClient: () => {
+      if (!this._subgraph) {
+        throw new Error("GraphQl client not inicialized")
```
brickpop:
```suggestion
        throw new Error("GraphQl client not initialized");
```

javascript/modules/client/src/internal/core.ts
```diff
@@ -237,5 +245,21 @@ export abstract class ClientCore implements IClientCore {
...
+      }
+      return this._subgraph
+    },
+    isUp: async ():Promise<boolean> => {
+      return this.subgraph.getClient().request(status)
+        .then((res) => {
+          if (res._meta && res._meta.deployment) {
```
brickpop:
```suggestion
          if (res._meta?.deployment) {
```

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -14,7 +14,9 @@ export interface IClient extends IClientCore {
       tokenAddresses: string[],
     ) => Promise<AssetBalance[]>;
     /** Retrieves the list of transfers from or to the given DAO, by default, ETH, DAI, USDC and USDT on Mainnet*/
-    getTransfers: (daoAddressOrEns: string) => Promise<IAssetTransfers>;
+    // TODO
+    // return typing
+    getTransfers: (daoAddressOrEns: string, searchParams: ISearchParams) => Promise<any>;
```
brickpop: Why adding a regression here? The UI already has data to work with, removing the type is not going to help, at least until we start dealing with real data.

brickpop: The searchParams parameter is fine, maybe we could make the parameter optional?

javascript/modules/client/src/internal/interfaces/core.ts
```diff
@@ -32,10 +33,12 @@ export interface IClientIpfsCore {
 }
 export interface IClientGraphQLCore {
   // Add here
+  getClient: () => GraphQLClient
```
brickpop, line 35: I understand that, unlike Web3 endpoints or IPFS clients... we only have one GraphQL endpoint? We should check if more than one could be supplied/rotated in case of error

javascript/modules/client/src/internal/queries/dao.ts
```diff
@@ -0,0 +1,53 @@
+import { gql } from "graphql-request";
+
+export const getDaoByAddress = gql`
```
brickpop: I would rename all of these queries into something that states that this is a GraphQL query, not a regular function:

examples:  `getDaoByAddressQuery`, `QueryGetDaoTransfersByAddres` , etc

javascript/modules/client/src/internal/queries/dao.ts
```diff
@@ -0,0 +1,53 @@
...
+    dao(id: $address){
+      id
+      name
+      metadata
+      createdAt
+    }
+  }
+`
+
+export const getDaoTransfersByAddress = gql`
```
brickpop: Is is possible to somehow parameterize the return types here? Or we should define types at the output returning function?

javascript/modules/client/src/internal/queries/withdraw.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from "graphql-request";
+
+export const searchWithdraw = gql`
```
brickpop: I think we need some higher level renamings here...  It should be `withdrawals` or `getWithdrawals`... filters/search is assumed

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -349,6 +350,48 @@ describe("Client", () => {
       ).toBe("7");
     });
   });
+  describe("GraphQL Client", () => {
+    it("Should create an invalid subgraph client", async () => {
```
brickpop:
```suggestion
    it("Should detect an invalid subgraph endpoint", async () => {
```

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -349,6 +350,48 @@ describe("Client", () => {
...
+      const isUp = await client.subgraph.isUp()
+      expect(isUp).toBe(false);
+    })
+    it("Should create an valid subgraph client", async () => {
+      const ctx = new Context(contextParamsLocalChain);
+      const client = new Client(ctx)
+      const isUp = await client.subgraph.isUp()
+      expect(isUp).toBe(true);
+    })
+    // it("Should get a dao with a specific address", async () => {
```
brickpop:
```suggestion
    test.todo("Should get a DAO's metadata with a specific address");
    // it("Should get a DAO's metadata from a given address", async () => {
```

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -349,6 +350,48 @@ describe("Client", () => {
...
+    // })
+    it("Should get the transfers of a dao", async () => {
+      const ctx = new Context(contextParamsLocalChain);
+      const client = new Client(ctx)
+      const daoAddress = '0x04d9a0f3f7cf5f9f1220775d48478adfacceff61'
+      const res = await client.methods.getTransfers(daoAddress)
+      expect(Array.isArray(res.deposits)).toBe(true)
+      expect(Array.isArray(res.withdraws)).toBe(true)
+    })
+    it("Should return null when trying the get the transfers of a dao that does not exist", async () => {
```
brickpop:
```suggestion
    it("Should return null when getting the transfers of a DAO that does not exist", async () => {
```

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -349,6 +350,48 @@ describe("Client", () => {
...
+      const ctx = new Context(contextParamsLocalChain);
+      const client = new Client(ctx)
+      const res = await client.methods.getTransfers(contextParamsLocalChain.dao)
+      expect(res).toBe(null)
```
brickpop: IMO, this should return an empty array, not null.

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -349,6 +350,48 @@ describe("Client", () => {
...
+    })
+    it("Should get the transfers of a dao", async () => {
```
brickpop:
```suggestion
    it("Should fail if the given ENS is invalid", async () => {
```

brickpop:
@josemarinas tests don't seem to be passing. And also, we should rebase to include the last changes on `develop`

brickpop:
Also @josemarinas, both the context and the Client's should support having multiple `subgraph` clients, and eventually `shiftClient()` if something goes wrong

javascript/modules/client/src/client.ts
josemarinas: Forgot to delete it

javascript/modules/client/src/client.ts
josemarinas: Forgot to comment it. I'll comment it and add a comment

javascript/modules/client/src/client.ts
josemarinas: :+1:

javascript/modules/client/src/internal/core.ts
josemarinas: Sure :+1:

javascript/modules/client/src/internal/interfaces/client.ts
josemarinas: Fabrice proposed a `SerarchParams` interface i the last commit, using the graphql like search params

```ts
export type DaoQueryOptions = {
  sortBy?: DaoSortBy;
  sortDirection?: "asc" | "desc";
  skip?: number;
  limit?: number;
};
```

Maybe changing this to something like this would make more sense, let me know what you think: 

```ts
export interface ISearchParams {
  limit?: number
  skip?: number
  sortDirection?: "asc" | "desc"
}

export interface ITransferSearchParams  extends ISearchParams{
  sortBy?: TransferSortBy
  addressOrEns?: string
}

export interface IDaoSearchParams  extends ISearchParams{
  sortBy?: DaoSortBy
  addressOrEns?: string
}
```
@brickpop

javascript/modules/client/src/internal/queries/dao.ts
josemarinas: Ok, makes sense, I'll update them with the new naming

javascript/modules/client/src/internal/queries/dao.ts
josemarinas: I searched a bit about this and I didn't find a way of typing the results in the same query, we should do it in the request

Something like this should do it

```ts
this.subgraph.getClient().request(QueryStuff)
.then((res:Stuff[]) => {
  return res
})
```
@brickpop

javascript/modules/client/src/internal/interfaces/core.ts
josemarinas: I didn't think that this could be a problem since it only had one 1 url, but since the subgraph hosted service is being deprecated, I don't know if we will deploy multiple endpoints so i'll add the functionality

javascript/modules/client/src/internal/queries/withdraw.ts
josemarinas: Changed, `searchWithdraw` to `QueryGetWithdrawals` and `searchDeposit` to `QueryGetDeposits`

Let me know if you are ok with the renaming @brickpop

javascript/modules/client/src/client.ts
```diff
@@ -327,7 +327,7 @@ export class Client extends ClientCore implements IClient {
 
   //// PRIVATE METHODS METADATA
 
-  private _getMetadata(daoAddressOrEns: string): Promise<DaoMetadata> {
+  private async _getMetadata(daoAddressOrEns: string): Promise<DaoMetadata> {
```
brickpop: Why async without an `await`?

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -35,6 +35,11 @@ export interface IClient extends IClientCore {
   };
 }
 
+export interface ISearchParams{
```
brickpop, line 38: Is this about search params? or rather about listing/pagination?

javascript/modules/client/src/internal/interfaces/core.ts
```diff
@@ -31,11 +32,14 @@ export interface IClientIpfsCore {
...
+  shiftClient: () => void
+  isUp: () => Promise<boolean>
+  ensureOnline: () => Promise<void>;
 }
 
 export interface IClientCore {
   web3: IClientWeb3Core;
   ipfs: IClientIpfsCore;
-  graphql: IClientGraphQLCore;
+  subgraph: IClientGraphQLCore;
```
brickpop: General question: Should this be about subgraph or about GraphQL?
Subgraph is the specific service we use not, but GraphQL is the general technology being used.
I guess all appearences should speak about `graphql` and similar

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -441,31 +507,31 @@ async function createLegacyDao(params: ContextParams) {
     })
 }
 
-export interface ICreateDaoERC20Voting {
+interface ICreateDaoERC20Voting {
```
brickpop, line 510: Are these interfaces defined on the testing? Can't they be extracted from the source logic itself?

javascript/modules/client/test/unit/context-erc20.test.ts
```diff
@@ -28,6 +28,7 @@ describe("ContextErc20 instances", () => {
       web3Providers: web3endpoints.working,
       gasFeeEstimationFactor: 0.1,
       pluginAddress: "0x2345",
+      subgraphURLs: []
```
brickpop: Same as above about graphql

javascript/modules/client/test/unit/context.test.ts
```diff
@@ -27,6 +27,7 @@ describe("Context instances", () => {
       daoFactoryAddress: "0x1234",
       web3Providers: web3endpoints.working,
       gasFeeEstimationFactor: 0.1,
+      subgraphURLs: []
```
brickpop: Same

brickpop (commented):
Thank you for the changes. I still have another comment:
- Could we add some usage example (readme) at least until there is a real implementation? I don't see how query parameters are being passed for example

javascript/modules/client/src/internal/interfaces/client.ts
josemarinas: For now is for listing, but I assume that at some point we will need to filter by other parameters, for example, status for the voting proposals. We can change it to something like `IPaginationParams` or `IListingParams` if you prefer

javascript/modules/client/src/client.ts
josemarinas: Forgot to delete it after modifying the contents of the function

javascript/modules/client/src/internal/interfaces/core.ts
josemarinas: This change was to unify the codebase because in some places it was used as `graphql` and in others as `subgraph`. About the naming is weird because in some parts of the code `subgraph` made more sense and in others `graphql`, so in the end i just picked one, but I guess `graphql` makes more sense in general so I'll change it

javascript/modules/client/test/integration/client.test.ts
josemarinas: Nope, these interfaces are temporary because they are using the old contracts structure to create a dao, just to test the deposit. Once the new contracts are deployed, I'll delete them

josemarinas:
> Thank you for the changes. I still have another comment:
> 
>     * Could we add some usage example (readme) at least until there is a real implementation? I don't see how query parameters are being passed for example

There is an example with QueryParams and `getMetadataMany`, I'm not sure if you refer to this

javascript/modules/client/test/integration/client.test.ts
brickpop: Ok, then let's add some clear `// @ TODO` comments, so that it's clear that these types are meant to be removed in short

javascript/modules/client/src/internal/interfaces/core.ts
brickpop: :+1:

javascript/modules/client/src/internal/interfaces/client.ts
brickpop: If "filter" sounds more reasonable to you, let's use it, but search is so specific... maybe we just want to list DAO's, with no filter at all

javascript/modules/client/test/integration/client.test.ts
josemarinas: There is already a big comment regarding this

javascript/modules/client/src/internal/interfaces/client.ts
josemarinas: :+1:

javascript/modules/client/src/internal/interfaces/client.ts
brickpop: The proposal below makes sense, if so I'd propose different namings:
- `IPagination` with `limit`, `skip` and `sortDirection`
- Anything extending it, named like `ITransferQueryParams`, `IDaoQueryParams`, ... `xxxQueryParams`

javascript/modules/client/src/internal/queries/dao.ts
brickpop: Okk, as long as the SDK's results are typed, we're all good 👍

javascript/modules/client/src/internal/queries/withdraw.ts
brickpop: I'd personally use `QueryWithdrawals` and `QueryDeposits`, but if the underlying graphql query is named as `GetXXX` then, it's fine :+1:

brickpop (approved):

PR #51, opened 2022-08-16
Fix: Polish sdk mocks

josemarinas:
## Description

Fixes and polishing of the sdk mock branch

Task: [APP-7878](https://aragonassociation.atlassian.net/browse/APP-878)

## Type of change

## Checklist:

### DIFF
--- a/javascript/modules/client/README.md
+++ b/javascript/modules/client/README.md
@@ -251,7 +251,7 @@ console.log(daos)
     };
     plugins: [
       {
-        id: "allowList-voting.plugin.dao.eth",
+        id: "addressList-voting.plugin.dao.eth",
         instanceAddress: "0x12345..."
       }
     ]
@@ -459,8 +459,8 @@ const pluginInitParams2: IErc20PluginInstall = {
     ]
   }
 }
-const erc20InstallEntry1 = ClientErc20.encoding.installEntry(pluginInitParams1)
-const erc20InstallEntry2 = ClientErc20.encoding.installEntry(pluginInitParams2)
+const erc20InstallPluginItem1 = ClientErc20.encoding.getPluginInstallItem(pluginInitParams1)
+const erc20InstallPluginItem2 = ClientErc20.encoding.getPluginInstallItem(pluginInitParams2)
 
 const createParams: ICreateParams = {
   metadata: {
@@ -473,7 +473,7 @@ const createParams: ICreateParams = {
     }]
   },
   ensSubdomain: "my-org", // my-org.dao.eth
-  plugins: [erc20InstallEntry1, erc20InstallEntry2]
+  plugins: [erc20InstallPluginItem1, erc20InstallPluginItem2]
 }
 
 // gas estimation
@@ -501,7 +501,7 @@ for await (const step of steps) {
 ### Creating an ERC20 proposal
 
 ```ts
-import { ClientErc20, Context, ContextPlugin, ICreateProposal, ProposalCreationSteps, VoteValues } from "@aragon/sdk-client";
+import { ClientErc20, Context, ContextPlugin, ICreateProposalParams, ProposalCreationSteps, VoteValues } from "@aragon/sdk-client";
 import { Wallet } from 'ethers'
 
 // Create a simple context
@@ -512,7 +512,8 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientErc20(contextPlugin)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
   metadata: {
     title: "Test Proposal",
     summary: "This is a short description",
@@ -559,7 +560,7 @@ for await (const step of steps) {
 ### Creating an ERC20 proposal with an action
 
 ```ts
-import { ClientErc20, Context, ContextPlugin, ICreateProposal, ProposalCreationSteps, VoteValues, IProposalSettings } from "@aragon/sdk-client";
+import { ClientErc20, Context, ContextPlugin, ICreateProposalParams, ProposalCreationSteps, VoteValues, IProposalSettings } from "@aragon/sdk-client";
 import { Wallet } from 'ethers'
 
 // Create a simple context
@@ -577,9 +578,10 @@ const configActionPrarms: IProposalSettings = {
   minTurnout: 0.5 // 50%
 }
 
-const configAction = client.encoding.setPluginConfigAction(configActionPrarms)
+const configAction = client.encoding.updatePluginSettingsAction(configActionPrarms)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
   metadata: {
     title: "Test Proposal",
     summary: "This is a short description",
@@ -637,9 +639,13 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientErc20(contextPlugin)
 
-const proposalId = "0x12345...0x1"
+const voteParams: IVoteProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
+  proposalId: '0x1234567890123456789012345678901234567890',
+  vote: VoteValues.YES
+}
 
-const steps = client.methods.voteProposal(proposalId, VoteValues.YES)
+const steps = client.methods.voteProposal(voteParams)
 for await (const step of steps) {
   try {
     switch (step.key) {
@@ -901,7 +907,7 @@ const pluginInitParams: IAddressListPluginInstall = {
   ]
 };
 
-const addressListInstallEntry = ClientAddressList.encoding.installEntry(pluginInitParams)
+const addressListInstallPluginItem = ClientAddressList.encoding.getPluginInstallItem(pluginInitParams)
 
 const createParams: ICreateParams = {
   metadata: {
@@ -914,7 +920,7 @@ const createParams: ICreateParams = {
     }]
   },
   ensSubdomain: "my-org", // my-org.dao.eth
-  plugins: [addressListInstallEntry]
+  plugins: [addressListInstallPluginItem]
 }
 
 // gas estimation
@@ -941,7 +947,7 @@ for await (const step of steps) {
 
 ### Creating a address list proposal
 ```ts
-import { ClientAddressList, Context, ContextPlugin, ICreateProposal, ProposalCreationSteps, VoteValues } from "@aragon/sdk-client";
+import { ClientAddressList, Context, ContextPlugin, ICreateProposalParams, ProposalCreationSteps, VoteValues } from "@aragon/sdk-client";
 import { Wallet } from 'ethers'
 
 // Create a simple context
@@ -952,7 +958,8 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientAddressList(contextPlugin)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
   metadata: {
     title: "Test Proposal",
     summary: "This is a short description",
@@ -999,7 +1006,7 @@ for await (const step of steps) {
 ### Creating a address list proposal with an action
 
 ```ts
-import { ClientAddressList, Context, ContextPlugin, ICreateProposal, ProposalCreationSteps, VoteValues, IProposalSettings } from "@aragon/sdk-client";
+import { ClientAddressList, Context, ContextPlugin, ICreateProposalParams, ProposalCreationSteps, VoteValues, IProposalSettings } from "@aragon/sdk-client";
 import { Wallet } from 'ethers'
 
 // Create a simple context
@@ -1017,9 +1024,10 @@ const configActionPrarms: IProposalSettings = {
   minTurnout: 0.5 // 50%
 }
 
-const configAction = client.encoding.setPluginConfigAction(configActionPrarms)
+const configAction = client.encoding.updatePluginSettingsAction(configActionPrarms)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
   metadata: {
     title: "Test Proposal",
     summary: "This is a short description",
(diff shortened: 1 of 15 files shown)


javascript/modules/client/README.md
```diff
@@ -512,7 +512,8 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientErc20(contextPlugin)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
```
brickpop: Simpler? 

```suggestion
  pluginAddress: "0x123456789012345678901234567890123456789012",
```

javascript/modules/client/README.md
```diff
@@ -637,9 +639,13 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientErc20(contextPlugin)
 
-const proposalId = "0x12345...0x1"
+const voteParams: IVoteProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
```
```suggestion
  pluginAddress: "0x123456789012345678901234567890123456789012",
```

javascript/modules/client/README.md
```diff
@@ -952,7 +958,8 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context, pluginAd
 // Create an ERC20 client
 const client = new ClientAddressList(contextPlugin)
 
-const proposalParams: ICreateProposal = {
+const proposalParams: ICreateProposalParams = {
+  pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
```
brickpop: Same

javascript/modules/client/src/client.ts
```diff
@@ -121,6 +121,7 @@ export class Client extends ClientCore implements IClient {
   estimation = {
     create: (params: ICreateParams) => this._estimateCreation(params),
     deposit: (params: IDepositParams) => this._estimateDeposit(params),
+    increaseAllowance: (params: IDepositParams) => this._estimateIncreaseAllowance(params),
```
brickpop: I would repurpose this in terms of `updateAllowance`

javascript/modules/client/src/client.ts
```diff
@@ -483,6 +495,7 @@ export class Client extends ClientCore implements IClient {
 // @ts-ignore  TODO: Remove this comment
 function unwrapCreateDaoParams(
   params: ICreateParams,
+  metadataIpfsUri: string
 ): [DAOFactory.DAOConfigStruct, DAOFactory.VoteConfigStruct, string, string] {
   // TODO: Serialize plugin params into a buffer
```
brickpop: The goal of this function should be to receive developer friendly parameters and rearrange them in terms that are smart contract friendly. The IPFS payload should be part of the developer submitted payload. Does `ICreateParams` include the metadata object? At some point the JSON string should be uploaded to IPFS, but this is outside of this function's scope.
Let's carefully think about that when we implement the final functionality

javascript/modules/client/src/context.ts
```diff
@@ -71,12 +71,8 @@ export class Context {
...
         contextParams.gasFeeEstimationFactor,
       ),
-      ipfs: contextParams.ipfsNodes.map((config) =>
-        new IpfsClient(config.url, config.headers)
-      ),
-      graphql: contextParams.graphqlNodes.map((url) =>
-        new GraphQLClient(url)
-      ),
+      ipfs: this.useIpfs(contextParams.ipfsNodes),
+      graphql: this.useGraphql(contextParams.graphqlNodes),
```
brickpop: I'm a bit confused. I would have sworn that `useWeb3Providers` was basically a setter, and that we were using it to update the Context state. But apparently it is just a digest helper?
useIpfs and useGraphql feels odd too, I would have an internal `resolveIpfs`, `resolveWeb3Providers` and `resolveGraphql()` function doing this, and then `useXXX` doing the setting

```suggestion
      ipfs: contextParams.ipfsNodes,
      graphql: contextParams.graphqlNodes,
```

brickpop: Adapt according to the above

javascript/modules/client/src/context.ts
```diff
@@ -107,14 +103,10 @@ export class Context {
...
-      this.state.ipfs = contextParams.ipfsNodes.map((config) =>
-        new IpfsClient(config.url, config.headers)
-      );
+      this.state.ipfs = this.useIpfs(contextParams.ipfsNodes)
     }
     if (contextParams.graphqlNodes?.length) {
-      this.state.graphql = contextParams.graphqlNodes.map((url) =>
-        new GraphQLClient(url)
-      );
+      this.state.graphql = this.useGraphql(contextParams.graphqlNodes)
```
brickpop: `this.useIpfs(...)`  and `this.useGraphql()` would feel best if repurposed as setters

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -58,7 +58,7 @@ function unwrapErc20InitParams(params: IErc20PluginInstall): [string, string, Bi
   ]
 }
 
-export function encodeActionSetPluginConfig(params: IProposalSettings): Uint8Array {
+export function encodeActionSetPluginConfig(params: IPluginSettings): Uint8Array {
```
brickpop: Better like this?

```suggestion
export function encodeSetPluginSettingsAction(params: IPluginSettings): Uint8Array {
```

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -67,7 +67,7 @@ export function encodeActionSetPluginConfig(params: IProposalSettings): Uint8Arr
   return hexToBytes(strip0x(hexBytes));
 }
 
-function unwrapSetPluginConfig(params: IProposalSettings): [BigNumber, BigNumber, BigNumber] {
+function unwrapSetPluginConfig(params: IPluginSettings): [BigNumber, BigNumber, BigNumber] {
```
brickpop:
```suggestion
function unwrapUpdatePluginSettings(params: IPluginSettings): [BigNumber, BigNumber, BigNumber] {
```

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -36,6 +36,7 @@ export interface IClient extends IClientCore {
   estimation: {
     create: (params: ICreateParams) => Promise<GasFeeEstimation>;
     deposit: (params: IDepositParams) => Promise<GasFeeEstimation>;
+    increaseAllowance: (params: IDepositParams) => Promise<GasFeeEstimation>;
```
brickpop: Same as above

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -74,19 +80,29 @@ export interface IProposalSettings {
   minDuration: number;
 }
 
-export interface ICreateProposal {
+export interface ICreateProposalParams {
+  pluginInstanceAddress: string
```
brickpop:
```suggestion
  pluginAddress: string
```

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -74,19 +80,29 @@ export interface IProposalSettings {
...
   executeOnPass?: boolean;
   creatorVote?: VoteValues;
 }
+export interface IVoteProposalParams {
+  pluginInstanceAddress: string
+  vote: VoteValues;
+  proposalId: string;
+}
+export interface IExecuteProposalParams {
+  pluginInstanceAddress: string
```
brickpop:
```suggestion
  pluginAddress: string
  vote: VoteValues;
  proposalId: string;
}
export interface IExecuteProposalParams {
  pluginAddress: string
```

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -153,7 +155,8 @@ describe("Client", () => {
       const context = new ContextPlugin(contextParamsLocalChain)
       const client = new ClientAddressList(context)
 
-      const proposalParams: ICreateProposal = {
+      const proposalParams: ICreateProposalParams = {
+        pluginInstanceAddress: '0x123456789012345678901234567890123456789012',
```
brickpop: Same

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -197,7 +200,8 @@ describe("Client", () => {
           reference: 'test'
         })
 
-      const proposalParams: ICreateProposal = {
+      const proposalParams: ICreateProposalParams = {
+        pluginInstanceAddress: '0x123456789012345678901234567890123456789012',
```
brickpop: Same

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -242,10 +246,13 @@ describe("Client", () => {
       const context = new ContextPlugin(contextParamsLocalChain)
       const client = new ClientAddressList(context)
 
-      const estimation = await client.estimation.voteProposal(
-        '0x1234567890123456789012345678901234567890',
-        VoteValues.YES
-      )
+      const voteParams: IVoteProposalParams = {
+        pluginInstanceAddress: "0x123456789012345678901234567890123456789012",
```
brickpop: Same

javascript/modules/client/src/client.ts
josemarinas: Yes, `ICreateParams` include the metadata as an Object, the idea is to upload this object to IPFS and here we introduce the received `cid`

javascript/modules/client/src/internal/encoding/plugins.ts
josemarinas: I would say `encodeUpdatePluginSettingsAction` is a better name since the function that calls this function is `_buildUpdatePluginSettingsAction`

javascript/modules/client/src/context.ts
josemarinas: Cannot do this, both graphql and ipfs typing is either a `GraphqlClient[]` or a `IpfsClient[]`, and the `contextParams.ipfsNodes` contains an object with the url and the headers. The `useIpfs` goes through this list creating the clients

javascript/modules/client/src/client.ts
brickpop: Ok, but then... is this function called for handling the response? or for receiving the user data and eventually transforming it for the blockchain? If it's the second case, then I guess the function should be called `wrapXXX` instead? 

I'm not clear at all what's the purpose of the function at this point, stringifying the received object should not be this function's job... at least someone else serializing and uploading to IPFS before, then wrapping the params here

javascript/modules/client/src/context.ts
brickpop: Ping

brickpop: My point is that something like `useIpfs()` or `useGraphql()` sounds like a setter. But here is being used as a helper that transforms some input into some output. 

I would have:
- `useWeb3()`, `useIpfs()` and `useGraphql()` as the setters to update the state
- These would receive the context param types and internally use something like `digestWeb3Context()` or similar to do the "helper" part of transforming stuff

However, this may make more sense in the scope of the client than on the context one? WDYT?

javascript/modules/client/src/context.ts
```diff
@@ -144,6 +136,37 @@ export class Context {
     }
   }
 
+  useIpfs(
```
brickpop: Same as above, this should be an internal non-class-related function with a different name

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -74,19 +80,29 @@ export interface IProposalSettings {
...
+export interface IExecuteProposalParams {
+  pluginAddress: string
+  proposalId: string;
+}
 
 // Factory init params
 
 export type IErc20PluginInstall = {
-  proposals: IProposalSettings;
+  proposals: IPluginSettings;
```
brickpop: Since the plugin settings can grow beyond the token scope, would you feel like `settings: IPluginSettings` makes more sense now?

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -105,7 +121,7 @@ type NewTokenParams = {
 
 export type IAddressListPluginInstall = {
   addresses: string[],
-  proposals: IProposalSettings
+  proposals: IPluginSettings
```
brickpop: Since the plugin settings can grow beyond the token scope, would you feel like `settings: IPluginSettings` makes more sense now?

brickpop (approved):

javascript/modules/client/src/client.ts
josemarinas: Yes, this is exactly the idea behind this, you already uploaded the metadata to ipfs and already have a `cid`, now you have this and a bunch of other input parameters that you want to use to call a function on the SCs, so you call this function, and you receive an array with the exact format that you will need to call the function in the `core-contracts-ethers` package.

About the `unwrap` naming, is the same naming that we've been using in the rest of the code for functions that have the exact same purpose but in other context like `unwrapWithdrawParams`, `unwrapErc20InitParams`, `unwrapErc20InitParams`...

javascript/modules/client/src/context.ts
josemarinas: I believe it's a matter of naming, as you say, the `use...` naming maybe does not make much sense for these functions, maybe we can change it to something like `get...Clients` or something more clear since this functions transform `urls` into `clients`.

I then believe that a function for setting the `web3` or `ipfs` doer not make much sense, we already have that with the `set` function that receives a `Partial` of the `ContextParams` this would imply 2 ways of doing the exact same thing

javascript/modules/client/src/internal/interfaces/plugins.ts
josemarinas: Yes, now it makes more sense in general

javascript/modules/client/src/context.ts
josemarinas: see comment below

javascript/modules/client/src/client.ts
brickpop: So, regargind the IPFS data, I would suggest that this function receives the CiD as opposed to the full object then. Because it could get serialized in different ways (with spaces, without spaces, etc) and create a different CiD than the intended one

javascript/modules/client/src/context.ts
brickpop: Makes sense :+1: 
So, can we rename these data transforming functions into something like `digestIpfsContext(...)`, `digestWeb3Context(...)` and move them as private helpers at the bottom of the file?

javascript/modules/client/src/internal/interfaces/plugins.ts
brickpop: Let's go for it

javascript/modules/client/src/internal/interfaces/plugins.ts
brickpop: Let's also go for it

javascript/modules/client/src/client.ts
```diff
@@ -496,8 +509,7 @@ function unwrapCreateDaoParams(
   return [
     {
       name: params.ensSubdomain,
-      // TODO: Return the IPFS URI (ipfs://<cid>)
-      metadata: JSON.stringify(params.metadata)
+      metadata: metadataCid
```
```suggestion
      metadata: "ipfs://" + metadataCid
```

javascript/modules/client/src/internal/interfaces/plugins.ts
brickpop:
```suggestion
  settings: IPluginSettings;
```

javascript/modules/client/src/internal/interfaces/plugins.ts
brickpop:
```suggestion
  settings: IPluginSettings
```

PR #61, opened 2022-08-25
Feature: Add decoders for all the existing actions

josemarinas:
## Description

Add decoders for the following actions
- **Client**
  - Withdraw
  - UpdateMetadata

- **ERC20**
  - UpdatePluginSettings

- **Address List**
  - UpdatePluginSettings

Task: [APP-929](https://aragonassociation.atlassian.net/browse/APP-929)

## Type of change

## Checklist:


javascript/modules/client/src/client-erc20.ts
```diff
@@ -284,6 +294,11 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
     }
   }
 
+  private _decodeUpdatePluginSettingsAction(data: Uint8Array): IPluginSettings {
+    const settings = decodeUpdatePluginSettingsAction(data)
+    return settings
+  }
```
brickpop: Why don't we simply call this function above? This private function adds nothing

javascript/modules/client/src/client.ts
```diff
@@ -181,6 +181,28 @@ export class Client extends ClientCore implements IClient {
...
+
+    /**
+     * Decodes a dao metadata from an encoded update metadata action
+     *
+     * @param {Uint8Array} data
+     * @return {*}  {Promise<IMetadata>}
+     * @memberof Client
+     */
+    updateMetadataAction: (data: Uint8Array): Promise<IMetadata> =>
+      this._decodeMetadataAction(data),
```
```suggestion
      this._decodeUpdateMetadataAction(data),
```

javascript/modules/client/src/client.ts
```diff
@@ -587,13 +601,32 @@ export class Client extends ClientCore implements IClient {
...
     return {
       to: address,
       value: BigInt(0),
       data: encodeUpdateMetadataAction(cid)
     }
   }
+
+  private _decodeWithdrawAction(data: Uint8Array): IWithdrawParams {
+    return decodeWithdrawActionData(data)
+  }
```
brickpop: Why don't we simply call the final function?

javascript/modules/client/src/client.ts
```diff
@@ -587,13 +601,32 @@ export class Client extends ClientCore implements IClient {
...
+
+  private async _decodeMetadataAction(data: Uint8Array): Promise<IMetadata> {
+    const cid = decodeUpdateMetadataAction(data)
+    try {
+      const stringMetadata = await this.ipfs.fetchString(cid)
+      return JSON.parse(stringMetadata)
+    } catch {
+      throw new Error("Error reading data from IPFS")
+    }
+  }
```
brickpop: Ok, maybe here we need to confirm with @Fabricevladimir whether it makes sense from the UI to receive the actual metadata or rather, the function being invoked and the individual parameters being passed to it

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -429,6 +435,65 @@ describe("Client", () => {
...
+  describe('Action decoders', () => {
+    it("Should decode an encoded withdraw action", async () => {
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+
+      const withdrawParams: IWithdrawParams = {
+        recipientAddress: '0x1234567890123456789012345678901234567890',
+        amount: BigInt(10),
+        reference: 'test',
+        tokenAddress: "0x1234567890098765432112345678900987654321"
```
brickpop: Do we also have the case without an ERC20 token?

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -13,6 +14,25 @@ export function encodeWithdrawActionData(params: IWithdrawParams): Uint8Array {
   return hexToBytes((strip0x(hexBytes)));
 }
 
+export function decodeWithdrawActionData(data: Uint8Array): IWithdrawParams {
+  const daoInterface = DAO__factory.createInterface();
+  // get hex bytes
+  const result: Result = daoInterface.decodeFunctionData("withdraw", data);
+  // Strip 0x => cast to ASCII => encode in Uint8Array
```
brickpop: Are we doing it?

javascript/modules/client/README.md
```diff
@@ -1371,6 +1371,48 @@ const daoAddress = "0x1234567890123456789012345678901234567890"
 const withdrawAction = await client.encoding.withdrawAction(daoAddress, withdrawParams)
 console.log(withdrawAction)
 ```
+### Upadte Metadata Raw
```
```suggestion
### Update Metadata Raw
```

javascript/modules/client/README.md
```diff
@@ -1411,6 +1453,138 @@ const configAction = client.encoding.updatePluginSettingsAction(configActionPrar
...
 ```
 
+## Action decoders
+
+### Decode Withdraw Action
+```ts
+import { Client, Context, IWithdrawParams } from "@aragon/sdk-client";
+const context: Context = new Context(params)
+const client: Client = new Client(context)
+const data: Uint8Array = [12, ..., 56]
```
```suggestion
const data: Uint8Array = new Uint8Array([12, ..., 56])
```

javascript/modules/client/src/client-addressList.ts
```diff
@@ -335,4 +356,17 @@ export class ClientAddressList extends ClientCore implements IClientAddressList
...
+    try {
+      const func = getFunctionFragment(data)
+      return {
+        id: func.format("minimal"),
+        functionName: func.name,
+        hash: "0x" + bytesToHex(data).substring(0, 8)
+      }
+    } catch {
+      return null
+    }
```
brickpop: Do we need the try catch block? Won't it return `null` if not found, itself?

javascript/modules/client/src/client.ts
```diff
@@ -576,7 +621,30 @@ export class Client extends ClientCore implements IClient {
...
+    // check ens
+    let address = daoAddreessOrEns
+    if (!isAddress(daoAddreessOrEns)) {
+      const resolvedAddress = await this.web3.getSigner()?.resolveName(daoAddreessOrEns)
+      if (!resolvedAddress) {
+        throw new Error("invalid ens")
+      }
+      address = resolvedAddress
+    }
+    const ipfsRegex = /^ipfs:\/\/(Qm[1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})/gm
```
brickpop: Shouldn't this have a line end symbol in the end `$`? Otherwise, the `split()` below may catch extra garbage being given after the CiD

javascript/modules/client/src/client.ts
```diff
@@ -587,13 +655,48 @@ export class Client extends ClientCore implements IClient {
       address = resolvedAddress
     }
     // upload metadata to IPFS
-    const cid = "0x1234567890123456789012345678901234567890"
+    let cid: string
+    try {
+      cid = await this.ipfs.add(JSON.stringify(params))
+    } catch {
+      throw new Error("Error uploading data to IPFS")
```
```suggestion
      throw new Error("Could not pin the metadata on IPFS")
```

javascript/modules/client/src/client.ts
```diff
@@ -576,7 +621,30 @@ export class Client extends ClientCore implements IClient {
...
+      const resolvedAddress = await this.web3.getSigner()?.resolveName(daoAddreessOrEns)
+      if (!resolvedAddress) {
+        throw new Error("invalid ens")
+      }
+      address = resolvedAddress
+    }
+    const ipfsRegex = /^ipfs:\/\/(Qm[1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})/gm
+    if (!ipfsRegex.test(ipfsUri)) {
+      throw new Error("Invalid ipfs uri")
+    }
```
brickpop: We could be using `URL` to parse the IPFS URI, and then get the `hostname`, which would only contain the CiD

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -26,7 +51,21 @@ function unwrapWithdrawParams(
...
-  const args = hexToBytes(ipfsUri);
+  const args = new TextEncoder().encode(ipfsUri)
   const hexBytes = daoInterface.encodeFunctionData("setMetadata", [args])
   return hexToBytes((strip0x(hexBytes)));
 }
+
+export function decodeUpdateMetadataAction(data: Uint8Array): string {
+  const daoInterface = DAO__factory.createInterface();
+  const hexBytes = bytesToHex(data, true)
+  // @ts-ignore
```
brickpop: Why?

javascript/modules/client/src/internal/encoding/common.ts
```diff
@@ -0,0 +1,10 @@
+import { bytesToHex } from "@aragon/sdk-common";
+import { FunctionFragment } from "@ethersproject/abi";
+import { Interface } from 'ethers/lib/utils'
+import { AVAILABLE_FUNCTIONS_ARRAY } from "../constants/encoding";
+
+export function getFunctionFragment(data: Uint8Array): FunctionFragment {
```
brickpop: What fragment? The parameters segment?

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -22,7 +23,7 @@ function unwrapAddressListInitParams(params: IAddressListPluginInstall): [string
     AddressZero,
     BigNumber.from(Math.round(params.settings.minTurnout * 100)),
     BigNumber.from(Math.round(params.settings.minSupport * 100)),
```
brickpop: Shouldn't we be using the already merged precision helper?

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -53,7 +54,7 @@ function unwrapErc20InitParams(params: IErc20PluginInstall): [string, string, Bi
...
```
brickpop: Same everywhere else

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -71,6 +72,27 @@ function unwrapUpdatePluginSettings(params: IPluginSettings): [BigNumber, BigNum
...
+    throw new Error("The received action is different from the expected action")
+  }
+  const result = votingInterface.decodeFunctionData("changeVoteConfig", data);
+  return wrapUpdatePluginSettings(result)
+}
+
+function wrapUpdatePluginSettings(result: Result): IPluginSettings {
+  return {
+    minTurnout: result[0].toNumber() / 100,
+    minSupport: result[1].toNumber() / 100,
```
brickpop: Shouldn't we be using the already merged precision helper?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -380,65 +401,260 @@ describe("Client", () => {
...
+      const updateMetadataAction = await client.encoding.updateMetadataAction(
+        "0x1234567890123456789012345678901234567890",
+        params,
+      );
+
+      expect(() => client.decoding.withdrawAction(updateMetadataAction.data))
+        .toThrow("The received action is different from the expected action");
+    });
+
+    it("Should try to decode a non valid action return an error", async () => {
```
brickpop:
```suggestion
    it("Should try to decode a invalid action and return an error", async () => {
```

javascript/modules/client/src/internal/encoding/plugins.ts
brickpop: Still

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -26,7 +53,32 @@ function unwrapWithdrawParams(
 
 export function encodeUpdateMetadataAction(ipfsUri: string): Uint8Array {
   const daoInterface = DAO__factory.createInterface();
-  const args = hexToBytes(ipfsUri);
-  const hexBytes = daoInterface.encodeFunctionData("setMetadata", [args])
-  return hexToBytes((strip0x(hexBytes)));
+  const args = new TextEncoder().encode(ipfsUri);
+  const hexBytes = daoInterface.encodeFunctionData("setMetadata", [args]);
```
brickpop, line 64: Why do we need to encode the IPFS URI as bytes? Can't it just be the string itself?

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -26,7 +53,32 @@ function unwrapWithdrawParams(
...
+    );
+  }
+  const result = daoInterface.decodeFunctionData("setMetadata", data);
+  const bytes = hexToBytes(result[0]);
+  const cid = new TextDecoder().decode(bytes);
+  const ipfsRegex =
+    /^Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/;
+  if (!ipfsRegex.test(cid)) {
+    throw new Error(
+      "The received cid is not valid",
```
brickpop: The metadata URL defined on the DAO is invalid

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -26,7 +53,32 @@ function unwrapWithdrawParams(
...
+export function decodeUpdateMetadataAction(data: Uint8Array): string {
+  const daoInterface = DAO__factory.createInterface();
+  const hexBytes = bytesToHex(data, true);
+  const receivedFunction = daoInterface.getFunction(
+    hexBytes.substring(0, 10) as any,
+  );
+  const expectedFunction = daoInterface.getFunction("setMetadata");
+  if (receivedFunction.name !== expectedFunction.name) {
+    throw new Error(
+      "The received action is different from the expected action",
```
brickpop:
```suggestion
      "The received action is different from the expected one",
```

javascript/modules/client/src/internal/constants/encoding.ts
```diff
@@ -0,0 +1,5 @@
+export const AVAILABLE_FUNCTIONS_ARRAY: string [] = [
```
brickpop:
```suggestion
export const AVAILABLE_FUNCTION_SIGNATURES: string [] = [
```

javascript/modules/client/src/internal/encoding/client.ts
josemarinas: Nope, the contracts receive an array of bytes

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -36,29 +58,33 @@ export function encodeErc20ActionInit(params: IErc20PluginInstall): Uint8Array {
   return hexToBytes(strip0x(hexBytes));
 }
 
-function unwrapErc20InitParams(params: IErc20PluginInstall): [string, string, BigNumber, BigNumber, BigNumber, string] {
+function unwrapErc20InitParams(
+  params: IErc20PluginInstall,
+): [string, string, BigNumber, BigNumber, BigNumber, string] {
```
brickpop, line 71: Is there a way to automatically bind this to an automated Typescript check?

Something like `func(): ReturnType<typeof MyFunction>`...?
Maybe for another PR

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -31,7 +38,20 @@ export interface IClient extends IClientCore {
...
+    ) => Promise<DaoAction>;
+    updateMetadataAction: (
+      daoAddressOrEns: string,
+      params: IMetadata,
+    ) => Promise<DaoAction>;
+  };
+  decoding: {
+    withdrawAction: (data: Uint8Array) => IWithdrawParams;
+    updateMetadataRawAction: (data: Uint8Array) => string;
+    updateMetadataAction: (data: Uint8Array) => Promise<IMetadata>;
```
brickpop, line 53: `getInterface()` can return null. Shouldn't these 3 other functions return null too? If there's an error we should always be having the same behavior.

brickpop (commented):
- Are we returning proposals with actions that can be decoded?
- Do we have encoders/decoders for installing a new plugin?
- Rename getInterface() to findInterface()

javascript/modules/client/src/internal/interfaces/client.ts
brickpop: Let's leave these 3 methods as they are and only rename `getInterface()` into `findInterface()`

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -380,65 +401,260 @@ describe("Client", () => {
...
+      };
+      const updateMetadataAction = await client.encoding.updateMetadataAction(
+        "0x1234567890123456789012345678901234567890",
+        params,
+      );
+      const recoveredIpfsUri: string = await client.decoding
+        .updateMetadataRawAction(updateMetadataAction.data);
+      const ipfsRegex =
+        /^ipfs:\/\/(Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,}))$/;
+      expect(ipfsRegex.test(recoveredIpfsUri)).toBe(true);
```
brickpop, line 537: I would an assertion on the CiD to ensure that the same data gets hashed in the same way

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -380,65 +401,260 @@ describe("Client", () => {
...
+      expect(iface?.id).toBe("function setMetadata(bytes)");
+      expect(iface?.functionName).toBe("setMetadata");
+      expect(iface?.hash).toBe("0xee57e36f");
+    });
+
+    it("Should try to get the function of an invalid data and return null", async () => {
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+      const data = new Uint8Array([11, 22, 22, 33, 33, 33]);
+      const iface = client.decoding.getInterface(data);
```
brickpop: findInterface()

javascript/modules/client/src/internal/constants/encoding.ts
```diff
@@ -0,0 +1,5 @@
+export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
+  "function changeVoteConfig(uint64,uint64,uint64)",
+  "function withdraw(address,address,uint256,string)",
+  "function setMetadata(bytes)",
+];
```
brickpop: Could these values be automatically obtained from the existing contract interfaces and function ID's ?

josemarinas: Yes, we can do something like this for each function:

```ts
const daoInterface = DAO__factory.createInterface();
const withdrawFunction = daoInterface.getFunction("withdraw");
withdrawFunction.format('minimal')
```

Does not make much sense because we are hardcoding the name of the function anyway, so if the function name changes we will need to change it

javascript/modules/client/src/internal/encoding/plugins.ts
josemarinas: I don't think we can do this, this is the fragment of code that makes use of this function:

```ts
  const erc20votingInterface = ERC20Voting__factory.createInterface();
  const args = unwrapErc20InitParams(params);
  // get hex bytes
  const hexBytes = erc20votingInterface.encodeFunctionData("initialize", args);
```
we will need to get the typing of the second argument of the `encodeFunctionData` function

javascript/modules/client/src/internal/constants/encoding.ts
brickpop: Right, but if one of the parameters changes, we will get the updated hash, for free. And if the function no longer exists, the call will throw an Error and we will immediately realize it. Also, for free

javascript/modules/client/src/internal/encoding/plugins.ts
brickpop: Well, technically speaking this is easy:

```ts
type MyArgs = Parameters<typeof (a: string, b: number) => { return a+b; }>;
type MyContractCallArgs = MyArgs[1]
```

The problem is that most likely, the type of the `args` parameter is generic, because we need to pass `"initialize"`, which is a dynamic value not known at compile time.

Isn't there an explicit function exported by the factory class? Otherwise, `encodeFunctionData` may not help

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -532,7 +532,10 @@ describe("Client", () => {
         .updateMetadataRawAction(updateMetadataAction.data);
       const ipfsRegex =
         /^ipfs:\/\/(Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,}))$/;
+
+      const expectedCid = await client.ipfs.add(JSON.stringify(params));
       expect(ipfsRegex.test(recoveredIpfsUri)).toBe(true);
+      expect(recoveredIpfsUri).toBe("ipfs://" + expectedCid);
```
brickpop: Testing the codebase with a parallel value that comes from the same underlying stack is a redundant test.
I would suggest to print this exact hash, hardcode it here, and let the test tell us if ever the same metadata produces a different value

javascript/modules/client/src/internal/encoding/plugins.ts
josemarinas: I don't think so, the nearest thing is:
```ts
const params = erc20votingInterface.functions["initialize(address,address,uint64,uint64,uint64,address)"].inputs
```
but this returns a `ParamType[]` which is not super useful

javascript/modules/client/src/internal/constants/encoding.ts
josemarinas: So, should we get this values when the client is created and store them in the context like we do with the `ipfsNodes`, `web3providers` but instead of receiving it from the user do it automatically?

javascript/modules/client/src/internal/encoding/plugins.ts
brickpop: Ok, then let's forget it

javascript/modules/client/src/internal/encoding/plugins.ts
```diff
@@ -19,6 +19,17 @@ import {
   IPluginSettings,
 } from "../interfaces/plugins";
 
+import { Interface } from "ethers/lib/utils";
```
brickpop: Could we make sure that we're using the scoped package? `@ethersproject/...`  instead of `ethers`

Otherwise, we should have installed the full `ethers` NPM package, which is quite large.
The same applies everywhere else

brickpop (approved):

PR #73, opened 2022-09-12
Feature: Add encoders/decoders for grant, revoke and freeze actions

josemarinas:
## Description

Add both the encode and decode functions for the following actions:
- Grant Permission
- Revoke Permission
- Freeze Permission

Task: [APP-945](https://aragonassociation.atlassian.net/browse/APP-945)

## Type of change

## Checklist:


javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+const grantParams = client.decoding.grantAction(
+  data
+);
+console.log(grantParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "0x1234567890...", // DaoPermission
```
brickpop: Shouldn't this be `"UPGRADE_PERMISSION"` ?

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+const revokeParams = client.decoding.revokeAction(
+  data
+);
+console.log(revokeParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "0x1234567890...", // DaoPermission
```
brickpop: Same

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+const data: Uint8Array =  new Uint8Array([12, ..., 56])
+
+const freezeParams = client.decoding.freezeAction(
+  data
+);
+console.log(freezeParams);
+/*
+{
+  where: "0x1234567890...",
+  permission: "0x1234567890...", // DaoPermission
```
brickpop: Same

javascript/modules/client/src/client.ts
```diff
@@ -179,6 +188,33 @@ export class Client extends ClientCore implements IClient {
   };
 
   encoding = {
+    /**
+     * Computes the payload to be given when creating a proposal that grants a permission in a dao
```
```suggestion
     * Computes the payload to be given when creating a proposal that grants a permission within a DAO
```

javascript/modules/client/src/client.ts
```diff
@@ -179,6 +188,33 @@ export class Client extends ClientCore implements IClient {
...
+     *
+     * @param {IPermissionParams} params
+     * @return {*}  {Promise<DaoAction>}
+     * @memberof Client
+     */
+    grantAction: (params: IPermissionParams): Promise<DaoAction> =>
+      this._buildGrantAction(params),
+    /**
+     * Computes the payload to be given when creating a proposal that revokes a permission in a dao
```
```suggestion
     * Computes the payload to be given when creating a proposal that revokes a permission within a DAO
```

javascript/modules/client/src/client.ts
```diff
@@ -179,6 +188,33 @@ export class Client extends ClientCore implements IClient {
...
+     *
+     * @param {IPermissionParams} params
+     * @return {*}  {Promise<DaoAction>}
+     * @memberof Client
+     */
+    revokeAction: (params: IPermissionParams): Promise<DaoAction> =>
+      this._buildRevokeAction(params),
+    /**
+     * Computes the payload to be given when creating a proposal that freezes a permission in a dao
```
```suggestion
     * Computes the payload to be given when creating a proposal that freezes a permission within a DAO
```

javascript/modules/client/src/client.ts
```diff
@@ -179,6 +188,33 @@ export class Client extends ClientCore implements IClient {
...
 
   encoding = {
+    /**
+     * Computes the payload to be given when creating a proposal that grants a permission in a dao
+     *
+     * @param {IPermissionParams} params
+     * @return {*}  {Promise<DaoAction>}
+     * @memberof Client
+     */
+    grantAction: (params: IPermissionParams): Promise<DaoAction> =>
```
brickpop: Why a Promise?

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
     return Promise.resolve({ deposits, withdrawals });
   }
 
+  private async _buildGrantAction(
+    params: IPermissionParams,
+  ): Promise<DaoAction> {
+    let where = params.where;
+    let who = params.who;
```
```suggestion
    let { where, who } = params;
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+    const signer = this.web3.getSigner();
+    if (!signer) {
+      throw new Error("A signer is needed");
+    }
+    if (!isAddress(params.where)) {
```
```suggestion
    if (!isAddress(where)) {
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+      where = await resolveEns(signer, params.where);
```
```suggestion
      where = await resolveEns(signer, where);
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+    }
+
+    if (!isAddress(params.where)) {
+      who = await resolveEns(signer, params.who);
```
```suggestion
    if (!isAddress(who)) {
      who = await resolveEns(signer, who);
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+    }
```
brickpop: Same as above. Who is not properly checked

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+    params: IFreezeParams,
+  ): Promise<DaoAction> {
+    let where = params.where;
+    const signer = this.web3.getSigner();
+    if (!signer) {
+      throw new Error("A signer is needed");
+    }
+    if (!isAddress(params.where)) {
+      where = await resolveEns(signer, params.where);
+    }
```
brickpop: Same as above. Plus, error checks should always go first

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -4,13 +4,131 @@ import { BigNumber } from "@ethersproject/bignumber";
...
+    hexBytes.substring(0, 10) as any,
+  );
+  const expectedFunction = daoInterface.getFunction("grant");
+  if (receivedFunction.name !== expectedFunction.name) {
+    throw new Error(
+      "The received action is different from the expected one",
+    );
+  }
+  const result = daoInterface.decodeFunctionData("grant", data);
+  return wrapPermissionParams(result);
```
brickpop, line 98: Shouldn't we do a loop here, to find which of the knows PERMISSION ID's is been requested?

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -4,13 +4,131 @@ import { BigNumber } from "@ethersproject/bignumber";
...
+    hexBytes.substring(0, 10) as any,
+  );
+  const expectedFunction = daoInterface.getFunction("revoke");
+  if (receivedFunction.name !== expectedFunction.name) {
+    throw new Error(
+      "The received action is different from the expected one",
+    );
+  }
+  const result = daoInterface.decodeFunctionData("revoke", data);
+  return wrapPermissionParams(result);
```
brickpop, line 126: Same as above

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -4,13 +4,131 @@ import { BigNumber } from "@ethersproject/bignumber";
...
+}
+
+export function encodeGrantActionData(
+  params: IPermissionParams,
+): Uint8Array {
+  const daoInterface = DAO__factory.createInterface();
+  const args = unwrapPermissionParams(params);
+  // get hex bytes
+  const hexBytes = daoInterface.encodeFunctionData("grant", args);
+  return hexToBytes(strip0x(hexBytes));
```
brickpop, line 80: Are we receiving the permission name? or the ID?

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
     return Promise.resolve({ deposits, withdrawals });
   }
 
+  private async _buildGrantAction(
```
brickpop: If the reason for the function to be `async` is the ENS resolution, I would suggest the SDK only accepted `addresses` and have a sync function.

This could be misleading and confusing, because the proposal creator could be defining `my-wallet.eth`, resolving to `0x1234...` upon creation, then be updated to point to `0x5678...`. Which is the correct address to give permissions to?

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+        {
+          who,
+          where,
+          permission: params.permission,
+        },
+      ),
+    };
+  }
+
+  private async _buildRevokeAction(
```
brickpop: Non async

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +693,88 @@ export class Client extends ClientCore implements IClient {
...
+        {
+          who,
+          where,
+          permission: params.permission,
+        },
+      ),
+    };
+  }
+
+  private async _buildFreezeAction(
```
brickpop: Non async

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -82,6 +87,54 @@ export interface IWithdrawParams {
...
+  where: string;
+  who: string;
+  permission: DaoPermissions;
+}
+export interface IFreezeParams {
+  where: string;
+  permission: DaoPermissions;
+}
+
+export enum DaoPermissions {
```
brickpop: Could we create an object with the permissions computed, instead of hardcoded?

If malicious mutability of values is a concern, we can export a Class with static getters:
```ts
export class DaoPermissions {
  get UPGRADE_PERMISSION_ID() {
    return keccak256("...");
  }
  ...
}
````

File: javascript/modules/client/src/internal/interfaces/client.ts [hunk 21]
```diff
@@ -82,6 +87,54 @@ export interface IWithdrawParams {
...
+  UPGRADE_PERMISSION =
+    "0xd5a14b802580ed4dd25c37f4587d65235454a7d1c553c5226e9ae8a88921c15d",
+  SET_METADATA_PERMISSION =
```
brickpop: The hashed values are referred to as permission "ID". 
- `UPGRADE_PERMISSION_ID`

javascript/modules/client/src/internal/utils/common.ts
```diff
@@ -0,0 +1,14 @@
+import { Signer } from "@ethersproject/abstract-signer";
+
+export async function resolveEns(
```
brickpop: Do we really need this?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,57 @@ describe("Client", () => {
...
+        where: "0x1234567890123456789012345678901234567890",
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const grantAction = await client.encoding.grantAction(
+        grantParams,
+      );
+
+      expect(typeof grantAction).toBe("object");
+      expect(grantAction.data).toBeInstanceOf(Uint8Array);
```
brickpop: Why don't we assert on the `to`, `value` and `data` hex stringified value?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,57 @@ describe("Client", () => {
...
+        where: "0x1234567890123456789012345678901234567890",
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const revokeAction = await client.encoding.revokeAction(
+        revokeParams,
+      );
+
+      expect(typeof revokeAction).toBe("object");
+      expect(revokeAction.data).toBeInstanceOf(Uint8Array);
```
brickpop: Same as above

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,57 @@ describe("Client", () => {
...
+        where: "0x1234567890123456789012345678901234567890",
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const freezeAction = await client.encoding.freezeAction(
+        permissionParams,
+      );
+
+      expect(typeof freezeAction).toBe("object");
+      expect(freezeAction.data).toBeInstanceOf(Uint8Array);
```
brickpop: Same as above

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,57 @@ describe("Client", () => {
...
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const grantAction = await client.encoding.grantAction(
+        grantParams,
+      );
+
+      expect(typeof grantAction).toBe("object");
+      expect(grantAction.data).toBeInstanceOf(Uint8Array);
+    });
```
brickpop, line 466: Why don't we check that different values produce different results, accordingly?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,6 +504,70 @@ describe("Client", () => {
...
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const grantAction = await client.encoding.grantAction(
+        grantParams,
+      );
+      const decodedGrantParams: IPermissionParams = client.decoding
+        .grantAction(grantAction.data);
+
+      expect(decodedGrantParams.permission).toBe(grantParams.permission);
```
brickpop: This should not be a hex string at all (permission ID). This should be the human readable permission name.

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,6 +504,70 @@ describe("Client", () => {
...
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const revokeAction = await client.encoding.revokeAction(
+        revokeParams,
+      );
+      const decodedRevokeParams: IPermissionParams = client.decoding
+        .revokeAction(revokeAction.data);
+
+      expect(decodedRevokeParams.permission).toBe(revokeParams.permission);
```
brickpop: Same

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,6 +504,70 @@ describe("Client", () => {
...
+        permission: DaoPermissions.UPGRADE_PERMISSION,
+      };
+
+      const freezeAction = await client.encoding.freezeAction(
+        freezeParams,
+      );
+      const decodedFreezeParams: IFreezeParams = client.decoding
+        .freezeAction(freezeAction.data);
+
+      expect(decodedFreezeParams.permission).toBe(freezeParams.permission);
```
brickpop: Same

javascript/modules/client/README.md
josemarinas: It is, but the `console.log` will show it like this, but in the code the type will be `daoPermission` and if you try to equal it to `daoPermission.UPGRADE_PERMISSION` it will return `true`

javascript/modules/client/src/client.ts
josemarinas: You need to wait for the name resolution in case we are using an ENS domain

javascript/modules/client/src/internal/interfaces/client.ts
josemarinas: Yes, this will imply that we will need to make the permission of type string, the return of the getter, which forces us to validate that the permission is in a list of permissions which implies having a map between permissions and permission hash, which is what we were trying to avoid with the class

javascript/modules/client/README.md
```diff
@@ -1540,6 +1540,101 @@ which needs to be encoded in a low level format.
...
+  Client,
+  Context,
+  Permissions,
+  IFreezeParams,
+} from "@aragon/sdk-client";
+
+const context: Context = new Context(params);
+const client: Client = new Client(context);
+
+const freezeParams: IFreezePArams = {
```
```suggestion
const freezeParams: IFreezeParams = {
```

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+const grantParams = client.decoding.grantAction(
+  data
+);
+console.log(grantParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
+  }
```
```suggestion
}
```

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+const revokeParams = client.decoding.revokeAction(
+  data
+);
+console.log(revokeParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
+  }
```
```suggestion
}
```

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+const freezeParams = client.decoding.freezeAction(
+  data
+);
+console.log(freezeParams);
+/*
+{
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
+  }
```
```suggestion
}
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
   }
 
+  private _buildGrantAction(
+    params: IPermissionParams,
+  ): DaoAction {
+    const signer = this.web3.getSigner();
+    if (!signer) {
+      throw new Error("A signer is needed");
+    }
+    let { where, who } = params;
```
```suggestion
    const { where, who } = params;
```

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
+    if (!isAddress(where)) {
+      throw new Error("Invalid address");
+    }
+    if (!isAddress(who)) {
+      throw new Error("Invalid address");
+    }
+    return {
+      to: where,
```
brickpop: I think the `to` address always needs to be the DAO, where the DB is stored

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
+    if (!signer) {
+      throw new Error("A signer is needed");
+    }
+    let { where, who } = params;
+    if (!isAddress(where)) {
+      throw new Error("Invalid address");
+    }
+    if (!isAddress(who)) {
+      throw new Error("Invalid address");
+    }
```
brickpop: If the error message is identical, let's merge the two conditions in one.
Otherwise, let's return different error messages.
Also, let's group checks by `if / else if` instead of separate if's

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
+      throw new Error("A signer is needed");
+    }
+    let { where, who } = params;
+    if (!isAddress(where)) {
+      throw new Error("Invalid address");
+    }
+
+    if (!isAddress(who)) {
+      throw new Error("Invalid address");
+    }
```
brickpop: Same as above

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
+    return {
+      to: where,
```
brickpop: Same

javascript/modules/client/src/client.ts
```diff
@@ -630,6 +692,85 @@ export class Client extends ClientCore implements IClient {
...
+        {
+          who,
+          where,
+          permission: params.permission,
+        },
+      ),
+    };
+  }
+
+  private _buildFreezeAction(
```
brickpop, line 754: Same as 👆

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -4,13 +4,139 @@ import { BigNumber } from "@ethersproject/bignumber";
...
+}
+function wrapFreezeParams(
+  result: Result,
+): IFreezeParams {
+  return {
+    where: result[0],
+    permission:
+      // @ts-ignore
+      Object.keys(PermissionIds).find((k) => PermissionIds[k] === result[1])
+        .replace("_ID", "") || "",
```
brickpop:
```suggestion
        .replace(/_ID$/, "") || "",
```

brickpop: `find()` could possibly be null, so we may remove the `ts-ignore` by using `?.replace`

brickpop: We might also use a mapping PermissionId => Permission and Permission => PermissionId, but this solution works as well

javascript/modules/client/src/internal/encoding/client.ts
```diff
@@ -4,13 +4,139 @@ import { BigNumber } from "@ethersproject/bignumber";
...
+  ];
+}
+function wrapPermissionParams(
+  result: Result,
+): IPermissionParams {
+  return {
+    where: result[0],
+    who: result[1],
+    permission:
+      // @ts-ignore
```
brickpop: Same as above

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,122 @@ describe("Client", () => {
...
+      const grantAction = client.encoding.grantAction(
+        grantParams,
+      );
+      const grantAction2 = client.encoding.grantAction(
+        grantParams2,
+      );
+      const decoder = new TextDecoder();
+
+      expect(typeof grantAction).toBe("object");
+      expect(grantAction.to).toMatch(/^0x[A-Fa-f0-9]{40}$/i);
```
brickpop: If the next check is an exact match, there is no need to check a generic regex here

javascript/modules/client/test/integration/client.test.ts
brickpop: Could we check `bytesToHex(grantAction.data)` instead and compare it to a hardcoded hex string?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,122 @@ describe("Client", () => {
...
+        revokeParams,
+      );
+
+      const revokeAction2 = client.encoding.revokeAction(
+        revokeParams2,
+      );
+      const decoder = new TextDecoder();
+
+      expect(typeof revokeAction).toBe("object");
+      expect(revokeAction.to).toMatch(/^0x[A-Fa-f0-9]{40}$/i);
```
brickpop: Same

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +423,122 @@ describe("Client", () => {
...
+      const freezeAction = client.encoding.freezeAction(
+        freezeParams,
+      );
+      const freezeAction2 = client.encoding.freezeAction(
+        freezeParams2,
+      );
+      const decoder = new TextDecoder();
+
+      expect(typeof freezeAction).toBe("object");
+      expect(freezeAction.to).toMatch(/^0x[A-Fa-f0-9]{40}$/i);
```
brickpop: Same

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,10 +569,79 @@ describe("Client", () => {
...
-    it("Should decode an encoded raw withdraw action of an erc20 token", async () => {
+    it("Should decode an encoded grant action", () => {
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+
+      const grantParams: IPermissionParams = {
+        who: "0x1234567890123456789012345678901234567890",
+        where: "0x1234567890123456789012345678901234567890",
+        permission: Permissions.UPGRADE_PERMISSION,
+      };
```
brickpop: Could we have a test vector, with at least 2 different parameter sets being checked?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,10 +569,79 @@ describe("Client", () => {
...
+      expect(decodedGrantParams.where).toBe(
+        decodedGrantParams.where,
+      );
+      expect(decodedGrantParams.who).toBe(decodedGrantParams.who);
+    });
+    it("Should decode an encoded revoke action", () => {
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+
+      const revokeParams: IPermissionParams = {
```
brickpop: Same

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -450,10 +569,79 @@ describe("Client", () => {
...
+      expect(decodedRevokeParams.where).toBe(
+        decodedRevokeParams.where,
+      );
+      expect(decodedRevokeParams.who).toBe(decodedRevokeParams.who);
+    });
+    it("Should decode an encoded freeze action", () => {
       const context = new Context(contextParamsLocalChain);
       const client = new Client(context);
 
+      const freezeParams: IFreezeParams = {
```
brickpop: Same

javascript/modules/client/src/internal/encoding/client.ts
josemarinas: No, the error has nothing to do with this, it has to do with the permission constant
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'.ts(7053)
  ```

File: javascript/modules/client/src/client.ts (hunk 35 above)
Comment (josemarinas) [reply javascript/modules/client/src/client.ts]: From what I understood reading the contracts `where` is the dao address where the permissions are granted, `who` is the address of whom the permissions are granted to.

So here the `where` param will be the dao address. Is this incorrect?

Comment (brickpop) [reply javascript/modules/client/src/client.ts]: It doesn't have to be the DAO. The DAO's permissions database could contain rules affecting 100% external contracts. For example, a plugin could request a rule so that only helper `0x1234` was able to `MINT` on helper `0x2345`. None of which would be the DAO neither the Plugin

File: javascript/modules/client/src/internal/encoding/client.ts (hunk 40 above)
Comment (brickpop) [reply javascript/modules/client/src/internal/encoding/client.ts]: Uhm, then we could replace `{}` by something similar to `{} as { [k: string]: Permissions[string] }`   or even just `{} as { [k: string]: string }` ?

Comment (josemarinas) [reply javascript/modules/client/src/internal/encoding/client.ts]: `{} as { [k: string]: string }` did the trick, thank you :)

File: javascript/modules/client/src/client.ts (hunk 35 above)
Comment (josemarinas) [reply javascript/modules/client/src/client.ts]: Oh, ok, that makes sense, I'll add the dao address as a parameter then

File: javascript/modules/client/README.md [hunk 48]
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+const grantParams = client.decoding.grantAction(
+  data
+);
+console.log(grantParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
```
brickpop, line 1783: Could we include the `permissionId` as well? Could we also mention that `permission` may be null if the permission cannot be identified?

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+const revokeParams = client.decoding.revokeAction(
+  data
+);
+console.log(revokeParams);
+/*
+{
+  who: "0x1234567890...",
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
```
brickpop, line 1805: Same

javascript/modules/client/README.md
```diff
@@ -1663,6 +1758,74 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+const data: Uint8Array =  new Uint8Array([12, ..., 56])
+
+const freezeParams = client.decoding.freezeAction(
+  data
+);
+console.log(freezeParams);
+/*
+{
+  where: "0x1234567890...",
+  permission: "UPGRADE_PERMISSION",
```
brickpop, line 1826: Same

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -82,6 +89,44 @@ export interface IWithdrawParams {
...
+const Permissions = {
+  UPGRADE_PERMISSION: "UPGRADE_PERMISSION",
+  SET_METADATA_PERMISSION: "SET_METADATA_PERMISSION",
+  EXECUTE_PERMISSION: "EXECUTE_PERMISSION",
+  WITHDRAW_PERMISSION: "WITHDRAW_PERMISSION",
+  SET_SIGNATURE_VALIDATOR_PERMISSION: "SET_SIGNATURE_VALIDATOR_PERMISSION",
+  SET_TRUSTED_FORWARDER_PERMISSION: "SET_TRUSTED_FORWARDER_PERMISSION",
+  ROOT_PERMISSION: "ROOT_PERMISSION",
+  CREATE_VERSION_PERMISSION: "CREATE_VERSION_PERMISSION",
+  MULTIPLY_PERMISSION: "MULTIPLY_PERMISSION",
```
brickpop: Some of these permissions are there just for testing purposes (multiply, do something, etc). Could we group them aside of the real ones?
- Prefix them with something like `__TEST_` + name
- Remove them from the declaration and "add" them right after:  `Permissions.__TEST_XX = "TEST_PERMISSION";

brickpop: Or instead, index them in a separate list for testing, and if the Permission ID does not correspond to any of the known "real" permissions, then try on the testing list. But don't export the testing list

javascript/modules/client/test/integration/client.test.ts
brickpop: We aren't doing this yet

javascript/modules/client/test/integration/client.test.ts
brickpop: Not yet

javascript/modules/client/test/integration/client.test.ts
brickpop: Not yet

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -552,71 +552,91 @@ describe("Client", () => {
...
+        {
+          where: "0x0987654321098765432109876543210987654321",
+          permission: Permissions.EXECUTE_PERMISSION,
+        },
+      ];
+      for (let i = 0; i < paramsArray.length; i++) {
+        const params = paramsArray[i];
+        const action = client.encoding.freezeAction(daoAddress, params);
+        const decodedParams: IPermissionParams = client.decoding
+          .grantAction(action.data);
```
brickpop: Why are we decoding as a `grant` action?

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -552,71 +552,91 @@ describe("Client", () => {
...
+          who: "0x0987654321098765432109876543210987654321",
+          where: "0x0987654321098765432109876543210987654321",
+          permission: Permissions.EXECUTE_PERMISSION,
+        },
+      ];
+      for (let i = 0; i < paramsArray.length; i++) {
+        const params = paramsArray[i];
+        const action = client.encoding.revokeAction(daoAddress, params);
+        const decodedParams: IPermissionParams = client.decoding
+          .grantAction(action.data);
```
brickpop: Why are we decoding as a `grant` action?

javascript/modules/client/src/client.ts
```diff
@@ -253,7 +255,7 @@ export class Client extends ClientCore implements IClient {
      * @return {*}  {IPermissionParams}
      * @memberof Client
      */
-    grantAction: (data: Uint8Array): IPermissionParams =>
+    grantAction: (data: Uint8Array): IPermissionResponse =>
```
brickpop: Response makes sense if there was a request, I'm not sure `{ who, where, permission, ... }` is a response to something

brickpop: How about `...DecodedParams`?

josemarinas: Seems fine to me :+1:

javascript/modules/client/README.md
```diff
@@ -1663,6 +1764,78 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+### Decode action grant permission
+
+```ts
+import { Client, Context, IPermissionParams, Permissions } from "@aragon/sdk-client";
+
+const context: Context = new Context(params);
+const client: Client = new Client(context);
+
+const data: Uint8Array =  new Uint8Array([12, ..., 56])
+const daoAddress = "0x1234567890123456789012345678901234567890"
```
brickpop: What's the use of `daoAddress` here?

josemarinas: Didn't notice it was a decoding, my bad

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -82,6 +89,47 @@ export interface IWithdrawParams {
...
+export interface IPermissionDecodedParams extends IPermissionParams {
+  permissionId: string;
+}
+export interface IFreezeParams {
+  where: string;
+  permission: string;
+}
+export interface IFreezeDecodedParams extends IFreezeParams {
+  permissionId: string;
+}
```
brickpop:
```suggestion
interface IPermissionParamsBase {
  where: string;
  who: string;
  permission: string;
}
interface IPermissionDecodedParamsBase extends IPermissionParamsBase {
  permissionId: string;
}

export interface IGrantPermissionParams extends IPermissionParamsBase {};
export interface IRevokePermissionParams extends IPermissionParamsBase {};

export interface IGrantPermissionDecodedParams extends IPermissionDecodedParamsBase {}
export interface IRevokePermissionDecodedParams extends IPermissionDecodedParamsBase {}

export interface IFreezePermissionParams {
  where: string;
  permission: string;
}
export interface IFreezePermissionDecodedParams extends IFreezePermission {
  permissionId: string;
}
```

brickpop: And update their usage

javascript/modules/client/test/integration/client.test.ts
```diff
@@ -420,6 +427,98 @@ describe("Client", () => {
...
+        {
+          who: "0x1234567890123456789012345678901234567890",
+          where: "0x1234567890123456789012345678901234567890",
+          permission: Permissions.UPGRADE_PERMISSION,
+        },
+      ];
+      let actions: DaoAction[] = [];
+      for (let i = 0; i < paramsArray.length; i++) {
+        const params = paramsArray[i];
+        actions.push(client.encoding.grantAction(daoAddress, params));
```
brickpop: Let's make sure that `daoAddress` doesn't have the same address `0x1234...` as many other test variables

javascript/modules/client/README.md
```diff
@@ -1663,6 +1764,77 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+```ts
+import { Client, Context, IGrantPermissionDecodedParams } from "@aragon/sdk-client";
+
+const context: Context = new Context(params);
+const client: Client = new Client(context);
+
+const data: Uint8Array =  new Uint8Array([12, ..., 56])
+
+const grantParams:IGrantPermissionDecodedParams = client.decoding.grantAction(
```
brickpop: Spacing, formatting. Still, not sure this typing is essential here

javascript/modules/client/README.md
```diff
@@ -1663,6 +1764,77 @@ const configAction = client.encoding.updatePluginSettingsAction(
...
+
+```ts
+import { Client, Context, IRevokePermissionDecodedParams } from "@aragon/sdk-client";
+
+const context: Context = new Context(params);
+const client: Client = new Client(context);
+
+const data: Uint8Array =  new Uint8Array([12, ..., 56])
+
+const revokeParams:IRevokePermissionDecodedParams = client.decoding.revokeAction(
```
brickpop: Same

brickpop (approved):

PR #74, opened 2022-09-15
Feature: Implement real data in ipfs and subgraph calls

josemarinas:
## Description

The aim of this PR is to connect the sdk with IPFS and subgraph, for this the sdk should call IPFS  whenever is possible and call subgraph for the following functions:
#### Client
- getDAo
- getDaos
#### ERC20 Client
- getProposal
- getProposals
- getSettings
#### AddressList Client
- getProposal
- getProposals
- getSettings

Task: [APP-926](https://aragonassociation.atlassian.net/browse/APP-926)

## Type of change

## Checklist:


javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+      const startDate = new Date(
+        parseInt(erc20VotingProposal.startDate) * 1000,
+      );
+      const endDate = new Date(parseInt(erc20VotingProposal.endDate) * 1000);
+      const creationDate = new Date(
+        parseInt(erc20VotingProposal.createdAt) * 1000,
+      );
+      let usedVotingWeight: bigint = BigInt(0);
+      erc20VotingProposal.voters.forEach(
+        (voter: { voter: { id: string }; vote: string; weight: string }) => {
```
brickpop: Could we factor this type out?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+          usedVotingWeight = usedVotingWeight + BigInt(voter.weight);
```
brickpop:
```suggestion
          usedVotingWeight += BigInt(voter.weight);
```

brickpop: I would either use a plain `for` loop, or do a `reduce()` instead of a `forEach()`

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+        },
+      );
+      // get status
+      const status = getProposalStatus(
```
brickpop:
```suggestion
      const status = computeProposalStatus(
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+              to: action.to,
+              value: BigInt(action.value),
+            };
+          },
+        ),
+        status,
+        result: {
+          yes: BigInt(erc20VotingProposal.yea),
+          no: BigInt(erc20VotingProposal.nay),
+          abstain: BigInt(erc20VotingProposal.abstain),
```
brickpop: Are we receiving subgraph data as numbers? or as strings? Big numbers will not be representable through a numeric graphql value

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+          //   BigInt(erc20VotingProposal.participationRequiredPct),
+          //   2,
+          // ),
+          // TODO DELETE ME
+          minSupport: parseFloat(
+            formatEther(erc20VotingProposal.supportRequiredPct),
+          ),
+          minTurnout: parseFloat(
+            formatEther(erc20VotingProposal.participationRequiredPct),
+          ),
```
brickpop: Aren't these percent values? We have the helper for encoding/decoding float's as bigints. Not sure why Ethers formatting is needed here

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+            parseInt(erc20VotingProposal.startDate),
+        },
+        token: {
+          address: erc20VotingProposal.pkg.token.id,
+          symbol: erc20VotingProposal.pkg.token.symbol,
+          name: erc20VotingProposal.pkg.token.name,
+          decimals: parseInt(erc20VotingProposal.pkg.token.decimals),
+        },
+        usedVotingWeight,
+        totalVotingWeight: BigInt(erc20VotingProposal.votingPower),
```
brickpop: Same, is the source a string?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+          (voter: { voter: { id: string }; vote: string; weight: string }) => {
+            let vote;
+            switch (voter.vote) {
+              case "Yea":
+                vote = VoteValues.YES;
+                break;
+              case "Nay":
+                vote = VoteValues.NO;
+                break;
+              case "Abstain":
```
brickpop: I think the string values changed? Otherwise, let's factor this in an enum so that we only need to change things once, whenever subgraph is updated to use "yes" "no", etc

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,21 +432,140 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+                vote = VoteValues.ABSTAIN;
+                break;
+            }
+            return {
+              address: voter.voter.id,
+              vote,
+              weight: BigInt(voter.weight),
+            };
```
brickpop: Is the full voter list returned here? Doesn't pagination apply?

javascript/modules/client/src/client.ts
```diff
@@ -366,6 +372,14 @@ export class Client extends ClientCore implements IClient {
       signer,
     );
 
+    // @ts-ignore
+    let cid: string;
```
```suggestion
    let cid = "";
```

javascript/modules/client/src/client.ts
```diff
@@ -366,6 +372,14 @@ export class Client extends ClientCore implements IClient {
...
+    try {
+      cid = await this.ipfs.add(JSON.stringify(params.metadata));
+    } catch {
+      throw new Error("Could not pin the metadata on IPFS");
```
brickpop, line 400: We might create and use custom errors in sdk/common

brickpop, line 400: We might create and use custom errors in sdk/common

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
-  private _getDao(daoAddressOrEns: string): Promise<DaoDetails> {
-    // TODO: Implement actual fetch logic using subgraph.
-
-    if (!daoAddressOrEns) {
-      throw new Error("Invalid DAO address or ENS");
+  private async _getDao(daoAddressOrEns: string): Promise<DaoDetails | null> {
+    if (!isAddress(daoAddressOrEns)) {
+      // TODO
+      // ens resolve
+      throw new Error("Invalid address");
```
brickpop: Can't we easily resolve it here?

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+        return null;
+      }
+      // const stringMetadata = await this.ipfs.fetchString(dao.metadata);
+      // TODO
+      // this is a temporal fix and should be changed by the line above
+      // but the current daos in subgraph dont have a proper metadata
+      const stringMetadata = await this.ipfs.fetchString(
+        "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr",
+      );
+      const metadata = JSON.parse(stringMetadata);
```
brickpop, line 647: We already have a task for that, but still good to have

```suggestion
      // TODO: Parse and validate schema
      const metadata = JSON.parse(stringMetadata);
```

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+          name: metadata.name,
+          description: metadata.description,
+          avatar: metadata.avatar || null,
+          links: metadata.links,
+        },
+        creationDate: new Date(parseInt(dao.createdAt) * 1000),
+        // TODO update when new subgraph schema is deployed
+        plugins: dao.packages.map(
+          (
+            pkg: { pkg: { id: string; __typename: string } },
```
brickpop: Could we factor this type out locally?

javascript/modules/client/src/client.ts
brickpop, line 647: We already have a task for that, but still good to have

```suggestion
      // TODO: Parse and validate schema
      const metadata = JSON.parse(stringMetadata);
```

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+          ): InstalledPluginListItem => {
+            return {
+              instanceAddress: pkg.pkg.id,
+              id: pkg.pkg.__typename,
```
brickpop: The plugin ID should be a bytes 32 value. Is `__typename` that?

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+      const client = this.graphql.getClient();
+      const { daos } = await client.request(QueryDaos, {
+        limit,
+        skip,
+        direction,
+        sortBy,
+      });
+      for (let i = 0; i < daos.length; i++) {
+        const dao = daos[i];
+        // const stringMetadata = await this.ipfs.fetchString(dao.metadata);
```
brickpop: Handling the array in serial mode is going to be quite slow. Could we turn this into a `Promise.all(daos.map(dao => {...}))` operation?

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+        // TODO
+        // this is a temporal fix and should be changed by the line above
+        // but the current daos in subgraph dont have a proper metadata
+        const stringMetadata = await this.ipfs.fetchString(
+          "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr",
+        );
+        const metadata = JSON.parse(stringMetadata);
```
```suggestion
        // TODO: validate model
        const metadata = JSON.parse(stringMetadata);
```

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+          address: dao.id,
+          ensDomain: dao.name,
+          metadata: {
+            name: metadata.name,
+            avatar: metadata.avatar || null,
+          },
+          // TODO update when new subgraph schema is deployed
+          plugins: dao.packages.map(
+            (
+              pkg: { pkg: { id: string; __typename: string } },
```
brickpop: Factor type out locally

javascript/modules/client/src/client.ts
```suggestion
        // TODO: validate model
        const metadata = JSON.parse(stringMetadata);
```

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+            ): InstalledPluginListItem => {
+              return {
+                instanceAddress: pkg.pkg.id,
+                id: pkg.pkg.__typename,
```
brickpop: Same question as above

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+    daoAddress: string,
     _tokenAddresses: string[],
-  ): Promise<AssetBalance[]> {
-    // TODO: Implement actual fetch logic using subgraph.
-    // Note: it would be nice if the client could be instantiated with dao identifier
-
-    if (!daoIdentifier) {
-      throw new Error("Invalid DAO address or ENS");
+  ): Promise<AssetBalance[] | null> {
+    // TODO check address
```
brickpop, line 697:
```suggestion
    if (tokenAddresses.includes(addr => !isAddress(addr))) {
      throw new Error("Invalid token address");
    }
```

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+    try {
+      await this.graphql.ensureOnline();
+      const client = this.graphql.getClient();
+      const { dao } = await client.request(QueryBalances, {
+        address: daoAddress,
+      });
+      if (!dao) {
+        return null;
+      }
+      if (dao.balances.length === 0) {
```
brickpop: What if we are being asked for a specific token, not part of the default set?

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+          type: "erc20",
+          address: balance.token.id,
+          name: balance.token.name,
+          symbol: balance.token.symbol,
+          decimals: parseInt(balance.token.decimals),
+          balance: BigInt(balance.balance),
+          lastUpdate: new Date(parseInt(balance.lastUpdated) * 1000),
+        });
+      }
+      return balances;
```
brickpop: If we're not handling the extra tokens here, do we have a task to address that in subgraph and then the SDK? Yet, a TODO comment here?

javascript/modules/client/src/client.ts
```diff
@@ -579,50 +593,152 @@ export class Client extends ClientCore implements IClient {
...
+    } catch (err) {
+      throw new Error("Cannot fetch the balance data from GraphQL");
     }
-
-    const AssetBalances: AssetBalance[] = assetList.map((token) => ({
-      ...token,
-    }));
-
-    return Promise.resolve(AssetBalances);
```
brickpop, line 625: Have we done a cleanup of all the existing mock data generators?

javascript/modules/client/src/client-erc20.ts
josemarinas: `string` or `null` in some cases, here the null case is not taken into account, but in the next version it will be

javascript/modules/client/src/client-erc20.ts
josemarinas: Yes, there is a big TODO explaining why this, but basically till now the `minTurnout` and `minSupport` were stored as `BigInts` with 18 numbers of precession. Right now we are storing them with a 2 digit precession, and our encoders-decoders have a max of 15 numbers of precision, so we will need to use the `formatEtther` until we start using the new contracts and storing them with 2 digit precision

javascript/modules/client/src/client-erc20.ts
josemarinas: Yes, subgraph always send strings on big numbers

javascript/modules/client/src/client-erc20.ts
josemarinas: Yes, It returns the full list, we can paginate this if you want

javascript/modules/client/src/client.ts
josemarinas: Yes I've been thinking on this, but would create a separate task to change all the error messages to the errors in common

javascript/modules/client/src/client.ts
josemarinas: Yes, but I was getting some errors on testing and didn't want to spend too much time fixing them, I'll enable ens for all the functions once I finish connecting them to subgraph

javascript/modules/client/src/client.ts
josemarinas: No, I forgot to make a Todo here, you are right this should be the address of the plugin or the ens name

javascript/modules/client/src/client-erc20.ts
josemarinas: Strings

javascript/modules/client/src/client-erc20.ts
josemarinas: Subgraph, at this moment, has a mix of different ways of formatting this values so until we have a standard multiple nonsense parse will happen depending on the situation

brickpop: Ok, so this requires subgraph to be refactored to only store standard numbers, and not bigints? If so, I can create a task

javascript/modules/client/src/client-erc20.ts
brickpop: mm ok, I'm more concerned about subgraph not exploding, rather than the SDK itself 🤔 
Another task maybe to consider on Subgraph?

javascript/modules/client/src/client.ts
brickpop: Was that the reason for the ts-ignore?

javascript/modules/client/src/client.ts
brickpop: Well, if the PR was unrelated to this feature... it might make sense, but if we're adding the functionality now, I don't see why adding technical debt would be a gain :)

javascript/modules/client/src/client.ts
brickpop:
```suggestion
      // TODO: Try resolving an ENS domain
      throw new Error("Invalid address");
```

javascript/modules/client/src/client.ts
brickpop: I couldn't see the removal of the unused mock generator functions

javascript/modules/client/src/client-addressList.ts
```diff
@@ -307,53 +322,222 @@ export class ClientAddressList extends ClientCore
...
+        parseInt(whitelistProposal.createdAt) * 1000,
+      );
+      // get status
+      const status = computeProposalStatus(
+        startDate,
+        endDate,
+        whitelistProposal.executed,
+        whitelistProposal
+          .yea,
+        whitelistProposal.nay,
```
brickpop: Isn't it easier to simply pass `whitelistProposal` and do the checks internally?

javascript/modules/client/src/client-addressList.ts
```diff
@@ -307,53 +322,222 @@ export class ClientAddressList extends ClientCore
...
+          ),
+          duration: parseInt(whitelistProposal.endDate) -
+            parseInt(whitelistProposal.startDate),
+        },
+        totalVotingWeight: parseInt(whitelistProposal.votingPower),
+        votes: whitelistProposal.voters.map(
+          (voter: { voter: { id: string }; vote: string; weight: string }) => {
+            let vote;
+            switch (voter.vote) {
+              case "Yea":
```
brickpop: ~~Could these 3 hardcoded strings be factored out somewhere else? Like `SUBGRAPH_YES_VALUE` and similar, in the file we have for subgraph queries~~

Can we also apply `SubgraphVoteValuesMap.get(voter.vote),` here?

javascript/modules/client/src/client-addressList.ts
```diff
@@ -307,53 +322,222 @@ export class ClientAddressList extends ClientCore
...
-        BigInt(proposal.result.yes),
-        BigInt(proposal.result.no),
+    let where = {};
+    let address = daoAddressOrEns;
+    if (address) {
+      if (!isAddress(address)) {
+        await this.web3.ensureOnline()
+        const provider = this.web3.getProvider();
+        if (!provider) {
+          throw new RequiredProviderError();
```
brickpop:
```suggestion
          throw new NoProviderError();
```

javascript/modules/client/src/client-addressList.ts
```diff
@@ -307,53 +322,222 @@ export class ClientAddressList extends ClientCore
...
+          skip,
+          direction,
+          sortBy,
+        },
       );
-      proposals.push(proposal);
+      await this.ipfs.ensureOnline();
+      return Promise.all(
+        whitelistProposals.map(
+          async (proposal: any): Promise<AddressListProposalListItem> => {
```
brickpop: Can't we type `proposal`?

javascript/modules/client/src/client-addressList.ts
```diff
@@ -307,53 +322,222 @@ export class ClientAddressList extends ClientCore
...
+            );
+            const endDate = new Date(parseInt(proposal.endDate) * 1000);
+            // get status
+            const status = computeProposalStatus(
+              startDate,
+              endDate,
+              proposal.executed,
+              proposal
+                .yea,
+              proposal.nay,
```
brickpop: Same as before, we might could receive just a proposal

javascript/modules/client/src/client-erc20.ts
```diff
@@ -461,13 +480,12 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+      for (let i = 0; i < erc20VotingProposal.voters.length; i++) {
+        const voter = erc20VotingProposal.voters[i];
+        usedVotingWeight += BigInt(voter.weight);
+      }
       // get status
-      const status = getProposalStatus(
+      const status = computeProposalStatus(
         startDate,
         endDate,
         erc20VotingProposal.executed,
```
brickpop: Same as before

javascript/modules/client/src/client-erc20.ts
```diff
@@ -504,9 +522,15 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
-          abstain: BigInt(erc20VotingProposal.abstain),
+          yes: erc20VotingProposal.yea
+            ? BigInt(erc20VotingProposal.yea)
+            : BigInt(0),
+          no: erc20VotingProposal.yea
+            ? BigInt(erc20VotingProposal.nay)
+            : BigInt(0),
+          abstain: erc20VotingProposal.yea
+            ? BigInt(erc20VotingProposal.abstain)
+            : BigInt(0),
```
brickpop:
```suggestion
          no: erc20VotingProposal.nay
            ? BigInt(erc20VotingProposal.nay)
            : BigInt(0),
          abstain: erc20VotingProposal.abstain
            ? BigInt(erc20VotingProposal.abstain)
            : BigInt(0),
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -542,81 +566,163 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+            );
+            const endDate = new Date(parseInt(proposal.endDate) * 1000);
+            // get status
+            const status = computeProposalStatus(
+              startDate,
+              endDate,
+              proposal.executed,
+              proposal
+                .yea,
+              proposal.nay,
```
brickpop: Same as above

javascript/modules/client/src/client.ts
```diff
@@ -629,18 +652,24 @@ export class Client extends ClientCore implements IClient {
         // TODO update when new subgraph schema is deployed
         plugins: dao.packages.map(
           (
-            pkg: { pkg: { id: string; __typename: string } },
+            plugin: SubgraphPlugin,
```
```suggestion
            plugin: SubgraphDaoPluginItem,
```

javascript/modules/client/src/client.ts
```diff
@@ -629,18 +652,24 @@ export class Client extends ClientCore implements IClient {
...
+              // plugins, however, this will eventually change
+              id: plugin.pkg.__typename + ".plugin.dao.eth",
               version: "0.0.1",
             };
           },
         ),
       };
     } catch (err) {
-      throw new Error("Cannot fetch the dao details data from GraphQL");
+      throw new GraphQLError("dao");
```
```suggestion
      throw new GraphQLError("DAO");
```

javascript/modules/client/src/client.ts
```diff
@@ -680,64 +708,81 @@ export class Client extends ClientCore implements IClient {
           // TODO update when new subgraph schema is deployed
           plugins: dao.packages.map(
             (
-              pkg: { pkg: { id: string; __typename: string } },
+              plugin: SubgraphPlugin,
```
brickpop: Same

javascript/modules/client/src/client.ts
```diff
@@ -680,64 +708,81 @@ export class Client extends ClientCore implements IClient {
...
                 version: "0.0.1",
               };
             },
           ),
-        });
-      }
-      return daosDetails;
+        };
+      }));
+      // return daosDetails;
```
brickpop: Should this line be removed then?

javascript/modules/client/src/client.ts
```diff
@@ -680,64 +708,81 @@ export class Client extends ClientCore implements IClient {
...
     } catch (err) {
-      throw new Error("Cannot fetch the dao details data from GraphQL");
+      throw new GraphQLError("dao");
```
```suggestion
      throw new GraphQLError("DAO");
```

javascript/modules/client/src/client.ts
```diff
@@ -680,64 +708,81 @@ export class Client extends ClientCore implements IClient {
...
       }
-      const balances: AssetBalance[] = [];
-      for (let i = 0; i < dao.balances.length; i++) {
-        const balance = dao.balances[i];
-        if (balance.token.symbol === "ETH") {
-          balances.push({
-            type: "native",
+      // TODO
+      // handle other tokens that are not ERC20 or eth
+      await this.ipfs.ensureOnline();
```
brickpop: Why do we need IPFS ensure online here?

javascript/modules/client/src/internal/graphql-queries/proposal.ts
```diff
@@ -44,3 +44,89 @@ query erc20VotingProposal($proposalId: ID!) {
   }
 }
 `;
+//, $sortBy: ERC20VotingProposal_orderBy!
+//, orderBy: $sortBy
```
brickpop: Is this going to help in the future?

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -274,7 +274,11 @@ export interface IDaoQueryParams extends IPagination {
...
 export enum DaoSortBy {
-  CREATED_AT = 'created_at',
-  NAME = 'name',
-  POPULARITY = 'totalProposals', // currently defined as number of proposals
+  CREATED_AT = "created_at",
+  NAME = "name",
+  POPULARITY = "totalProposals", // currently defined as number of proposals
+}
+
+export interface SubgraphPlugin {
```
brickpop: See the name suggested above

javascript/modules/client/src/internal/utils/plugins.ts
```diff
@@ -1,18 +1,27 @@
...
-    return ProposalStatus.SUCCEEDED
+    return ProposalStatus.SUCCEEDED;
   } else {
-    return ProposalStatus.DEFEATED
+    return ProposalStatus.DEFEATED;
   }
 }
 
+export function isProposalId(propoosalId: string): boolean {
+  const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
```
brickpop, line 46: Is the `0x1234_0x2345` format what we receive from subgraph?

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -36,7 +39,6 @@ const IPFS_API_KEY = process.env.IPFS_API_KEY ||
 const web3endpoints = {
   working: [
     "https://mainnet.infura.io/v3/94d2e8caf1bc4c4884af830d96f927ca",
-    "https://cloudflare-eth.com/",
```
brickpop: Don't we have another working one? Otherwise, `shiftNode` tests are less representative

javascript/modules/client/test/integration/client-erc20.test.ts
```diff
@@ -37,7 +40,6 @@ const IPFS_API_KEY = process.env.IPFS_API_KEY ||
 const web3endpoints = {
   working: [
     "https://mainnet.infura.io/v3/94d2e8caf1bc4c4884af830d96f927ca",
-    "https://cloudflare-eth.com/",
```
brickpop: Same

javascript/modules/client/test/integration/client-erc20.test.ts
```diff
@@ -582,31 +584,106 @@ describe("Client", () => {
...
+      const client = new ClientErc20(context);
+      const limit = 5;
+      const address = "0xn0tv4l1d";
+      const params: IProposalQueryParams = {
+        limit,
+        sortBy: ProposalSortBy.CREATED_AT,
+        direction: SortDirection.ASC,
+        daoAddressOrEns: address,
+      };
+      await expect(() => client.methods.getProposals(params)).rejects.toThrow(
```
brickpop, line 668: Ignorant question, is this extra rejects needed?

```suggestion
      await expect(() => client.methods.getProposals(params)).toThrow(
```

javascript/modules/common/src/errors.ts
```diff
@@ -8,3 +8,28 @@ export class UnsupportedProtocolError extends Error {
...
+  constructor() {
+    super("Invalid address");
+  }
+}
+export class InvalidProposalIdError extends Error {
+  constructor() {
+    super("Invalid proposal ID");
+  }
+}
+export class RequiredProviderError extends Error {
```
brickpop:
```suggestion
export class NoProviderError extends Error {
```

javascript/modules/client/src/client-addressList.ts
josemarinas: yes, forgot to replicate it here, my bad

javascript/modules/client/src/client.ts
josemarinas: If we don't do that, the calls inside the `promise.all` will call `ensureOnline` and since it is concurrent and there is only one context it never manages to get a working IPFS node, so I call the `ensureOnline` to make sure that we already have a working IPFS node so each `fetchString` does not start iterating over the nodes

javascript/modules/client/src/internal/utils/plugins.ts
josemarinas: yes, the proposal id is composed this way `daoAddtess_proposalIndex`, so for example, the proposal number 2 of the dao `0x1234...` will be `ox1234..._0x2`

javascript/modules/client/test/integration/client-erc20.test.ts
josemarinas: Yes, if it is a promise, you need the `rejects`.

javascript/modules/client/src/client.ts
```diff
@@ -682,52 +660,31 @@ export class Client extends ClientCore implements IClient {
...
+          // const stringMetadata = await this.ipfs.fetchString(dao.metadata);
+          // TODO
+          // this is a temporal fix and should be changed by the line above
+          // but the current daos in subgraph dont have a proper metadata
+          const stringMetadata = await this.ipfs.fetchString(
+            "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr",
+          );
+          const metadata = JSON.parse(stringMetadata);
+          return toDaoListItem(dao, metadata);
+        }),
```
brickpop: Turning this into raw promises (no async) would be better for the final bundle size. Same on the rest of Promise.all() where there's no `if/else`


```suggestion
        daos.map((dao: ISubgraphDaoListItem): Promise<DaoListItem> => {
          // const stringMetadata = await this.ipfs.fetchString(dao.metadata);
          // TODO
          // this is a temporal fix and should be changed by the line above
          // but the current daos in subgraph dont have a proper metadata

          return this.ipfs.fetchString(
            "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr",
          ).then(stringMetadata => {
            const metadata = JSON.parse(stringMetadata);
            return toDaoListItem(dao, metadata);
          });
        }),
```

javascript/modules/client/src/client.ts
```diff
@@ -752,34 +709,21 @@ export class Client extends ClientCore implements IClient {
...
-            address: balance.token.id,
-            name: balance.token.name,
-            symbol: balance.token.symbol,
-            decimals: parseInt(balance.token.decimals),
-            balance: BigInt(balance.balance),
-            lastUpdate: new Date(parseInt(balance.lastUpdated) * 1000),
-          };
-        }),
+        balances.map(
+          async (balance: ISubgraphBalance): Promise<AssetBalance> =>
```
brickpop: Why async?

```suggestion
          (balance: ISubgraphBalance): Promise<AssetBalance> =>
```

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -246,15 +246,17 @@ export type InstalledPluginListItem = {
   version: string;
 };
 
+export type DaoMetadata = {
```
brickpop: Didn't we already have such a type before?

javascript/modules/client/src/internal/utils/plugins.ts
brickpop: Ok, then we need a task in Subgraph to make it be just the proposal's ID... otherwise URL's are going to be too long

javascript/modules/client/src/internal/utils/plugins.ts
```diff
@@ -25,3 +49,241 @@ export function isProposalId(propoosalId: string): boolean {
...
+  const startDate = new Date(
+    parseInt(proposal.startDate) * 1000,
+  );
+  const endDate = new Date(parseInt(proposal.endDate) * 1000);
+  const creationDate = new Date(
+    parseInt(proposal.createdAt) * 1000,
+  );
+  let usedVotingWeight: bigint = BigInt(0);
+  for (let i = 0; i < proposal.voters.length; i++) {
+    const voter = proposal.voters[i];
```
brickpop:
```suggestion
  for (const voter of proposal.voters) {
```

javascript/modules/client/src/client.ts
brickpop: But does `toAssetBalance()` need to access IPFS?

javascript/modules/client/src/internal/interfaces/client.ts
josemarinas: Yes, I didn't saw it

javascript/modules/common/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ TEMPLATE:
 -->
 
 ## [UPCOMING] 
+### Added
+- New errors
```
brickpop: Hehehe no please :pray: 

```suggestion
- New error types
```

javascript/modules/client/src/client-erc20.ts
```diff
@@ -434,74 +452,149 @@ export class ClientErc20 extends ClientCore implements IClientErc20 {
...
+        {
+          address: pluginAddress,
+        },
+      );
+      if (!erc20VotingPackage) {
+        return null;
+      }
+      return {
+        minDuration: parseInt(erc20VotingPackage.minDuration),
+        minSupport: parseFloat(
```
brickpop, line 570:
```suggestion
        // TODO: use decodeRatio() when ready
        minSupport: parseFloat(
```

javascript/modules/client/src/internal/utils/plugins.ts
```diff
@@ -1,18 +1,288 @@
...
+import { formatEther } from "@ethersproject/units";
 
-
-export function getProposalStatus(startDate: Date, endDate: Date, executed: boolean, yes?: bigint, no?: bigint): ProposalStatus {
+export function computeProposalStatus(
+  proposal:
+    | ISubgraphErc20Proposal
+    | ISubgraphAddressListProposal
+    | ISubgraphErc20ProposalListItem
+    | ISubgraphAddressListProposalListItem,
```
brickpop: Maybe better using a common interface? Otherwise we will be adding types all the time

brickpop (approved):

PR #76, opened 2022-09-21
App 1024 sdk split client core inner components into separate classes

mathewmeconry:
## Description

Splits core class into multiple classes: web3, ipfs and graphql

Task: APP-1024

## Type of change

## Checklist:


javascript/modules/client/src/internal/modules/graphql.ts
```diff
@@ -0,0 +1,82 @@
...
+    public async isUp(): Promise<boolean> {
+        try {
+            const res = await this.getClient().request(QueryStatus)
+            if (res._meta?.deployment) {
+                return true
+            }
+            return false
+        } catch (e) {
+            return false
+        }
```
brickpop: Turning this into an `async` function creates a bigger bundle and overhead. If there is no specific need for `async/await` (if/else clauses, for await loops, etc) I'd suggest keeping it as it was.

javascript/modules/client/src/internal/modules/ipfs.ts
```diff
@@ -0,0 +1,92 @@
...
+    /** Returns `true` if the current client is on line */
+    public async isUp(): Promise<boolean> {
+        if (!this._ipfs?.length) return Promise.resolve(false);
+
+        try {
+            await this.getClient().nodeInfo()
+            return true
+        } catch (e) {
+            return false
+        }
```
brickpop: Same as before, an `async/await` function has a heavier footprint, is harder to debug and with the given logic could be kept as a plain function returning a chain of promises

javascript/modules/client/src/internal/modules/ipfs.ts
```diff
@@ -0,0 +1,92 @@
...
+            const res = await client.add(input)
+            return res.hash
+        } catch (e) {
+            throw new Error("Could not upload data: " + (e?.message || ""));
+        }
+    }
+
+    public async fetchBytes(cid: string): Promise<Uint8Array | undefined> {
+        const client = await this.getOnlineClient();
+        return client.cat(cid);
```
brickpop: Same as before with `async`

javascript/modules/client/src/internal/modules/web3.ts
```diff
@@ -0,0 +1,172 @@
...
+        if (!provider) return Promise.reject(new Error("No provider"));
+
+        return provider
+            .getNetwork()
+            .then(() => true)
+            .catch(() => false);
+    }
+
+    /**
+     * Returns a contract instance at the given address
```
brickpop: Could we leave `ensureOnline()` next to `isUp()` as it was recently?

brickpop (approved):

PR #78, opened 2022-09-27
Fix: PluginContext `fromContext` function

josemarinas:
## Description

- Fix `fromContext` function in the `ContextPlugin` class
- Removes `pluginAddress` property since it is no longer useful
- Fix typo in plugin name

Task: [APP-1038](https://aragonassociation.atlassian.net/browse/APP-1038)

## Type of change

## Checklist:

### DIFF
--- a/javascript/modules/client/package.json
+++ b/javascript/modules/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.8.1-alpha",
+  "version": "0.9.0-alpha",
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/sdk-client.esm.js",
--- a/javascript/modules/client/package.json
+++ b/javascript/modules/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.8.0-alpha",
+  "version": "0.8.1-alpha",
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/sdk-client.esm.js",
--- a/javascript/modules/client/test/integration/client-addressList.test.ts
+++ b/javascript/modules/client/test/integration/client-addressList.test.ts
@@ -6,8 +6,9 @@ import { Wallet } from "@ethersproject/wallet";
 import {
   Client,
   ClientAddressList,
+  Context,
+  ContextParams,
   ContextPlugin,
-  ContextPluginParams,
   SortDirection,
 } from "../../src";
 import * as ganacheSetup from "../../../../helpers/ganache-setup";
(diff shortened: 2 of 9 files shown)


javascript/modules/client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.8.0-alpha",
+  "version": "0.8.1-alpha",
```
brickpop: These changes feel like a breaking change, not a fix. Since we are in alpha stage, I would suggest bumping to 0.9.0-alpha

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -76,32 +77,29 @@ const grapqhlEndpoints = {
 const TEST_WALLET =
   "0xdf57089febbacf7ba0bc227dafbffa9fc08a93fdc68e1e42411a14efcf23656e";
 
-const contextParams: ContextPluginParams = {
+const contextParams: ContextParams = {
```
brickpop, line 80: Did we also update the readme?

javascript/modules/client/test/integration/client-addressList.test.ts
```diff
@@ -132,8 +130,9 @@ describe("Client", () => {
 
   describe("Client instances", () => {
     it("Should create a working client", async () => {
-      const ctx = new ContextPlugin(contextParams);
-      const client = new ClientAddressList(ctx);
+      const ctx = new Context(contextParams);
+      const ctxPlugin = ContextPlugin.fromContext(ctx)
+      const client = new ClientAddressList(ctxPlugin);
```
brickpop, line 135: Maybe this is more future proof, just a random question: if the current plugin context has no extra state now... does it make sense to keep a separate class? Probably it does, so that when future settings appear we don't need to resplit again

javascript/modules/client/test/integration/client-erc20.test.ts
```diff
@@ -78,31 +79,28 @@ const grapqhlEndpoints = {
 const TEST_WALLET =
   "0xdf57089febbacf7ba0bc227dafbffa9fc08a93fdc68e1e42411a14efcf23656e";
 
-const contextParams: ContextPluginParams = {
+const contextParams: ContextParams = {
```
brickpop, line 82: Same with the readme

javascript/modules/client/test/integration/client-addressList.test.ts
josemarinas: No, but the readme was already correct :man_facepalming:

javascript/modules/client/test/integration/client-addressList.test.ts
josemarinas: Yes, this was the exact thought process behind it, maybe it does not make sense atm, but if we want to add properties that are unique to the plugins in the future we can do it without having to split them up again. If in the end we don't add any property, we can get rid of it and use the normal `Context`

brickpop (approved):

PR #87, opened 2022-10-11
Feature: Generate `README` automatically from the examples folder 

josemarinas:
## Description

- Create an example/ folder
- Each folder contains files named like 1-context.ts, 2-client.ts, 3-dao-creation.ts, … sorted by logical order
- Each file contains a /** ... */ block with the contents to extract as the markdown text/content
- Below, there is a plain working example
- A Readme template contains the non-generated contents, with a placeholder to be replaced by the example data
- A script extracts the contents and generates the final readme (yarn run or npm run)
- Move the existing readme contents into examples/

Task: [APP-1023](https://aragonassociation.atlassian.net/browse/APP-1023)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


javascript/modules/client/README.generated.md
```diff
@@ -0,0 +1,422 @@
+
+## Depositing ERC20 tokens to a DAO
```
brickpop: Shouldn't this file be in `gitignore`? We only care about the final readme, not about an interim artifact

javascript/modules/client/examples/client-addressList/02-client.ts
```diff
@@ -0,0 +1,8 @@
+/* Header
+## Client Plugin AddressList
+
+*/
+
+/* Code */
+console.log("hello")
+/* - */
```
brickpop: I would suggest a more generic approach:
- Everything inside a `/* MARKDOWN  ...   */`  block is treated as raw markdown
- Anything else, is treated as Typescript

This allows us to have text and code in any arbitrary shape that we like, and we don't enfoce the first thing being the Header.

And this way we don't need the "code start" and "code end" tokens, which are easy to forget.

javascript/modules/client/examples/client/08-get-transfers.ts
```diff
@@ -0,0 +1,76 @@
...
+  limit: 10, // optional
+  skip: 0, // optional
+  direction: SortDirection.ASC, // optional
+  type: TransferType.DEPOSIT, // optional
+};
+const transfers: Transfer[] | null = await client.methods.getTransfers(params);
+console.log(transfers);
+/* - */
+/* Result
+[
```
brickpop: Couldn't the result token... be simply a TS comment and that's it?

javascript/modules/client/examples/constants.ts
```diff
@@ -0,0 +1,25 @@
+import { Wallet } from "@ethersproject/wallet";
```
brickpop: What's the purpose of this file? constants or context?

javascript/modules/client/README.generated.md
josemarinas: Yes, this is going to be renamed to `README.md` I left it there for you to see the result of the automation

javascript/modules/client/examples/constants.ts
josemarinas: The purpose is to not redefine the context all the time, we can call it `context.ts` if you prefer this naming

javascript/modules/client/examples/client-addressList/02-client.ts
josemarinas: Ok :+1:, makes sense

javascript/modules/client/README.generated.md
brickpop: mmm I wouldn't do so. The way I see it, the main README should have the general explanation about the library, installation, etc... and then link to a separate file called `examples.md` or similar with all the generated code

javascript/modules/client/examples/constants.ts
brickpop: if that makes sense, sure thing

javascript/modules/client/examples/00-client/00-installation.ts
```diff
@@ -0,0 +1,68 @@
...
+`@aragon/sdk-client` provides easy access to the high level interactions to be
+made with an Aragon DAO. It consists of three different components:
+
+- General-purpose DAO client
+- Custom clients for specific DAO plugins
+- Context for holding inheritable configuration
+
+Contributors: See [development](#development) below
+
+# Installation
```
brickpop: As said before, I would have this on the `README.md` file, as a static file. And then, have an `examples.md` file or even an examples index file, linking to each `examples/context.md`, etc

javascript/modules/client/package.json
```diff
@@ -22,7 +22,8 @@
     "lint": "tsdx lint",
     "prepare": "tsdx build",
     "size": "size-limit",
-    "analyze": "size-limit --why"
+    "analyze": "size-limit --why",
+    "readme": "node ./scripts/generate-readme.js ./examples ./README.generated.md"
```
brickpop:
```suggestion
    "readme": "node ./scripts/generate-readme.js ./examples ./examples.md"
```

javascript/modules/client/scripts/generate-readme.js
```diff
@@ -0,0 +1,78 @@
+const fs = require("fs");
+const { argv } = require("process");
+
+generateReadme();
+
+async function generateReadme() {
+  const path = argv[2];
+  const outputFile = argv[3];
+  const mdRegex = /(\/\* MARKDOWN)(.|\n)*?(\*\/)/;
```
brickpop:
```suggestion
  if(!path || !outputFile) {
    console.error("Usage:  node generate-readme.js <examples-folder> <output-file.md>);
    process.exit(1);
  }
  const mdRegex = /(\/\* MARKDOWN)(.|\n)*?(\*\/)/;
```

javascript/modules/client/scripts/generate-readme.js
```diff
@@ -0,0 +1,78 @@
...
+          data.toString(),
+          mdRegex,
+          ["/* MARKDOWN", "*/"],
+          true
+        );
+        const code = data.toString().replace(mdRegex, "");
+        let readmeStr = md;
+        if (code && code !== "\n") {
+          readmeStr = md + `\n\`\`\`ts\n` + code + `\n\`\`\`\n`;
+        }
```
brickpop: Does this support the case of a file having (1) markdown (2) code (3) more markdown (4) more code ?

I would suggest breaking this into simpler, more focused functions (doesn't need to be perfect though):
- buildExampleFile(path) => string
- parseFile(path) => tokens: Array<{ type: "code" | "markdown", content: string}>
- processTokens(tokens) => string

javascript/modules/client/README.md
```diff
@@ -21,2175 +21,7 @@ yarn add @aragon/sdk-client
...
-console.log(params)
-/*
-[
-  "0x12345...",
-  "0x56789...",
-  "0x13579...",
-]
-*/
-```
+[See examples file](./examples.md)
```
```suggestion
An exhaustive list of the SDK features and methods can be found on the **[examples.md](examples.md) file**.
```

javascript/modules/client/scripts/generate-readme.js
josemarinas: No, right now it does not support that, and with the aproach that we have rigth now we cannot support it, the only batch of stuff that has some kind of pattern of separator is the markdown
```ts
/* MARKDOWN
....
*/
```
We use the `mdRegex` to detect this pattern and get the section, and then we use the same `mdRegex` to delete this section and get the code. So if we have the pattern of having md -> code ->md -> code, we wouldn't be able to separate the firs section of code from the second one and generate an ordered markdown after that.

javascript/modules/client/README.md
```diff
@@ -21,2230 +21,7 @@ yarn add @aragon/sdk-client
...
-console.log(params)
-/*
-[
-  "0x12345...",
-  "0x56789...",
-  "0x13579...",
-]
-*/
-```
+An exhaustive list of the SDK features and methods can be found on the **[examples.md](examples.md) file**.
```
brickpop, line 24: I would also explain how to generate/update the examples file

javascript/modules/client/scripts/generate-examples-md.js
```diff
@@ -0,0 +1,105 @@
...
+  for (const path of paths) {
+    if (path.isDirectory()) {
+      const folderPath = rootPath + "/" + path.name;
+      let files = fs.readdirSync(folderPath);
+      files = prependString(folderPath + "/", files);
+      ret = [...ret, ...files];
+    }
+  }
+  return ret;
+}
```
brickpop: Couldn't we use `glob` or similar to expand all the files recursively? This will do fine with one folder level

javascript/modules/client/examples/context.ts
```diff
@@ -0,0 +1,25 @@
+import { Wallet } from "@ethersproject/wallet";
+import { ContextParams } from "@aragon/sdk-client";
+
+
+export const IPFS_API_KEY = "ipfs-api-key";
```
brickpop: Didn't we have any text/markdown here?

javascript/modules/client/scripts/generate-examples-md.js
```diff
@@ -0,0 +1,105 @@
+const fs = require("fs");
+const { argv } = require("process");
+const regex = /(\/\* MARKDOWN)(.|\n)*?(\*\/\n)/g;
+const removeStrings = ["/* MARKDOWN\n", "*/"];
+
+generateExamplesMd();
```
brickpop:
```suggestion
generateExamplesMd()
  .catch(err => {
    console.error(err);
    process.exit(1);
  });
```

javascript/modules/client/scripts/generate-examples-md.js
```diff
@@ -0,0 +1,105 @@
...
+      "Usage:  node generate-examples-md.js <examples-folder> <output-file.md>"
+    );
+    process.exit(1);
+  }
+  const files = getFileNames(path);
+  if (fs.existsSync(outputFile)) {
+    await fs.promises.rm(outputFile);
+  }
+  for (let i = 0; i < files.length; i++) {
+    const file = files[i];
```
brickpop:
```suggestion
  for (const file of files) {
```

javascript/modules/client/scripts/generate-examples-md.js
```diff
@@ -0,0 +1,105 @@
...
+    fs.readFile(file, async (err, data) => {
```
brickpop: This is a problem, because the `for` loop and the `readFile` functions are synchronous, while the inner callback is async. We could be continuing the execution flow before `appendFile` has been processed in all callbacks. 

I think we should build an array of promises and then do `Promise.all()`  on it.

javascript/modules/client/examples/context.ts
josemarinas: Nope :(, but it only has the `contextParams` that are imported in the rest of the examples

javascript/modules/client/scripts/generate-examples-md.js
josemarinas: Not sure if this will break it in the same way as the `for of` loop

javascript/modules/client/examples.md
```diff
@@ -90,6 +90,117 @@ for await (const step of steps) {
...
+        break;
+    }
+  } catch (err) {
+    console.error(err);
+  }
+}
+
+```
+### Depositing ERC20 tokens to a DAO
+
```
brickpop: Can we make sure that the 3 ticke are placed next to the code and not next to the markdown? Or at least, leave some extra new line?

brickpop: I would also advice performing a `.trim()` on the processed contents and then add the extra new lines to create at least one empty line between blocks

brickpop (approved):

PR #89, opened 2022-10-17
Feature: Add allowance estimation

josemarinas:
## Description

Add allowance estimation function

Task: [APP-1034](https://aragonassociation.atlassian.net/browse/APP-1034)

## Type of change

## Checklist:

### DIFF
--- a/javascript/modules/client/src/client.ts
+++ b/javascript/modules/client/src/client.ts
@@ -37,7 +37,6 @@ import { BigNumber } from "@ethersproject/bignumber";
 import { AddressZero } from "@ethersproject/constants";
 import {
   Contract,
-  ContractFactory,
   ContractReceipt,
   ContractTransaction,
 } from "@ethersproject/contracts";
@@ -56,6 +55,7 @@ import {
   InvalidAddressOrEnsError,
   NoProviderError,
   NoSignerError,
+  NoTokenAddress,
   Random,
   strip0x,
 } from "@aragon/sdk-common";
@@ -533,7 +533,9 @@ export class Client extends ClientCore implements IClient {
   _estimateDeposit(params: IDepositParams) {
     const signer = this.web3.getConnectedSigner();
     if (!signer) {
-      throw new Error("A signer is needed for estimating the gas cost");
+      throw new NoSignerError();
+    } else if (!signer.provider) {
+      throw new NoProviderError();
     }
 
     const [daoAddress, amount, tokenAddress, reference] = unwrapDepositParams(
@@ -546,9 +548,6 @@ export class Client extends ClientCore implements IClient {
     if (tokenAddress === AddressZero) {
       override.value = amount;
     }
-
-    // TODO: If the approved ERC20 amount is not enough,
-    // estimate the cose of increasing the allowance
 
     return daoInstance.estimateGas
       .deposit(tokenAddress, amount, reference, override)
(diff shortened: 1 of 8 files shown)


javascript/modules/client/src/client.ts
```diff
@@ -546,27 +549,48 @@ export class Client extends ClientCore implements IClient {
...
-
     return daoInstance.estimateGas
       .deposit(tokenAddress, amount, reference, override)
       .then((gasLimit) => {
         return this.web3.getApproximateGasFee(gasLimit.toBigInt());
       });
   }
 
-  _estimateUpdateAllowance(_params: IDepositParams) {
+  async _estimateUpdateAllowance(params: IDepositParams) {
```
brickpop, line 560: Why async?

javascript/modules/client/src/client.ts
```diff
@@ -546,27 +549,48 @@ export class Client extends ClientCore implements IClient {
...
+    //   const provider = this.web3.getProvider();
+    //   if (!provider) {
+    //     throw new NoProviderError();
+    //   }
+    //   const resolvedAddress = await provider.resolveName(daoAddreess);
+    //   if (!resolvedAddress) {
+    //     throw new InvalidAddressOrEnsError();
+    //   }
+    //   daoAddreess = resolvedAddress;
+    // }
```
brickpop: Why this comment? Can't we refactor this somewhere else and leave clean code here?

javascript/modules/common/src/errors.ts
```diff
@@ -43,3 +43,9 @@ export class UnexpectedActionError extends Error {
     super("The received action is different from the expected one");
   }
 }
+
```
brickpop, line 46: Did NoSignerError() already exist?

javascript/modules/client/src/client.ts
josemarinas: This comment was more of a doubt if we wanted ens resolution in this function, if we don't want it I'll remove the comment and the `async`, if we want to resolve it, I'll just uncomment

javascript/modules/client/src/client.ts
josemarinas: see below

javascript/modules/common/src/errors.ts
josemarinas: Yes, I just added an error for when the token address is not specified when trying to update the allowance because you need the token address

javascript/modules/client/src/client.ts
```diff
@@ -546,27 +549,48 @@ export class Client extends ClientCore implements IClient {
       override.value = amount;
     }
 
-    // TODO: If the approved ERC20 amount is not enough,
-    // estimate the cose of increasing the allowance
-
     return daoInstance.estimateGas
```
brickpop, line 553: Why are we removing the TODO without implementing it?

javascript/modules/client/src/client.ts
brickpop: ok, but then why don't we uncomment this part and have the feature?

javascript/modules/client/src/client.ts
josemarinas: I believe this comment was before we knew the UI will be dividing the deposit in two steps, so since we now have the estimate of increasing allowance i believe it does not make sense but let me know what you think

javascript/modules/client/src/client.ts
josemarinas: I wasn't sure that we wanted this feature here because we don't have it in `deposit` or `estimation.deposit` but I'll uncomment it and add a `TODO` to update the `deposit` and `estimation.deposit` later

javascript/modules/client/src/client.ts
brickpop: The point is that the SDK should make sense to anyone, not just to what the UI is deciding to do in UX terms. If we only leave this functionality, we should have two clearly distinct functions where names make it obvious that only one part is estimated. But personally, I prefer an SDK where the exact same thing that will be executed in real life is the one that we will estimate

javascript/modules/client/src/client.ts
```diff
@@ -565,20 +565,21 @@ export class Client extends ClientCore implements IClient {
...
-    //   }
-    //   const resolvedAddress = await provider.resolveName(daoAddreess);
-    //   if (!resolvedAddress) {
-    //     throw new InvalidAddressOrEnsError();
-    //   }
-    //   daoAddreess = resolvedAddress;
-    // }
+    
+    // resolve ens
+    let daoAddreess = params.daoAddress;
```
brickpop: If we leave this piece of code, shouldn't `daoAddress` be used below this block?

javascript/modules/client/src/client.ts
brickpop: In general, we should be able to handle ENS names everywhere, as the intended way to interact with DAO's will be with URL's where the parameter is a domain name, not a hex string (ideally)

javascript/modules/client/src/client.ts
josemarinas: I mean, for me the way it is right now is pretty clear and does not need changes:
- `estimation.deposit` -> estimates the gas fees for making a deposit
- `estimation.updateAllowance` -> estimates the gas fees for updating the allowance

brickpop: Ok, then I would say: Let's make sure that the JSDoc clearly states that `estimation.deposit()` does not include a potential update of the allowance.

javascript/modules/client/src/internal/interfaces/client.ts
```diff
@@ -160,6 +160,8 @@ export type DaoCreationStepValue =
 // DEPOSIT
 
 export interface IDepositParams {
+  // TODO
+  // update naming to daoAddressOrEns
   daoAddress: string;
```
brickpop: Why don't we change it now, then? The UI is only going to get more coupled than it is now

brickpop (approved):

brickpop (approved):

PR #90, opened 2022-10-18
Feature: Split clients into different classes

josemarinas:
## Description
Splits clients into different subclasses, each client now has a `modules` folder with 4 files inside:

- methods.ts
- estimation.ts
- encoders.ts
- decoders.ts

The tests are also divided accordingly and now use a `constants.ts` file to get all constants

Task: [APP-1025](https://aragonassociation.atlassian.net/browse/APP-1025)

## Type of change

## Checklist:


javascript/modules/client/src/internal/utils/client.ts
```diff
@@ -16,6 +23,55 @@ import {
   TransferType,
 } from "../interfaces/client";
 
+export function unwrapCreateDaoParams(
```
brickpop, line 26: Wasn't this renamed in a different way?  Wrap/Unwrap was gone

javascript/modules/client/src/client-addressList.ts
```diff
@@ -1,66 +1,19 @@
...
-import { isAddress } from "@ethersproject/address";
-import {
-  QueryAddressListProposal,
-  QueryAddressListProposals,
-} from "./internal/graphql-queries/proposal";
-import { QueryAddressListPluginSettings } from "./internal/graphql-queries/settings";
+import { IClientAddressListMethodsModule } from "./modules-addressList/methods";
+import { IClientAddressListEncodingModule } from "./modules-addressList/encoding";
+import { IClientAddressListDecodingModule } from "./modules-addressList/decoding";
+import { IClientAddressListEstimationModule } from "./modules-addressList/estimation";
```
brickpop: I would structure files and code differently. I would have `src/internal/<name>/<module>.ts` or similar

javascript/modules/client/src/client-addressList.ts
```diff
@@ -70,173 +23,25 @@ const PLUGIN_ID = "0x1234567890123456789012345678901234567890";
...
+  public encoding: IClientAddressListEncoding;
+  public decoding: IClientAddressListDecoding;
+  public estimation: IClientAddressListEstimation;
+
   constructor(context: ContextPlugin) {
     super(context);
+    this.methods = new IClientAddressListMethodsModule(context);
+    this.encoding = new IClientAddressListEncodingModule(context);
+    this.decoding = new IClientAddressListDecodingModule(context);
+    this.estimation = new IClientAddressListEstimationModule(context);
```
brickpop:
```suggestion
    this.methods = new IClientAddressListMethods(context);
    this.encoding = new IClientAddressListEncoding(context);
    this.decoding = new IClientAddressListDecoding(context);
    this.estimation = new IClientAddressListEstimation(context);
```

javascript/modules/client/src/client-addressList.ts
```diff
@@ -70,173 +23,25 @@ const PLUGIN_ID = "0x1234567890123456789012345678901234567890";
...
     /**
      * Computes the parameters to be given when creating the DAO,
      * so that the plugin is configured
      *
-     * @param {IErc20PluginInstall} params
-     * @return {*}  {FactoryInitParams}
+     * @param {IAddressListPluginInstall} params
+     * @return {*}  {IPluginInstallItem}
      * @memberof ClientAddressList
      */
```
brickpop, line 55: Why not importing the static bit from the module file as well?

javascript/modules/client/src/client-erc20.ts
```diff
@@ -1,703 +1,52 @@
...
-import {
-  QueryErc20Proposal,
-  QueryErc20Proposals,
-} from "./internal/graphql-queries/proposal";
-import { formatEther } from "@ethersproject/units";
-import { QueryErc20PluginSettings } from "./internal/graphql-queries/settings";
-import { QueryToken } from "./internal/graphql-queries/token";
+import { IClientErc20EncodingModule } from "./modules-erc20/encoding";
+import { IClientErc20DecodingModule } from "./modules-erc20/decoding";
+import { IClientErc20EstimationModule } from "./modules-erc20/estimation";
```
brickpop: Do we need `Module` in the name? same as above

javascript/modules/client/src/client-erc20.ts
```diff
@@ -1,703 +1,52 @@
...
+  public encoding: IClientErc20Encoding;
+  public decoding: IClientErc20Decoding;
+  public estimation: IClientErc20Estimation;
+
   constructor(context: ContextPlugin) {
     super(context);
+    this.methods = new IClientErc20MethodsModule(context);
+    this.encoding = new IClientErc20EncodingModule(context);
+    this.decoding = new IClientErc20DecodingModule(context);
+    this.estimation = new IClientErc20EstimationModule(context);
```
brickpop: Could this be readonly? I think that if it is assiged on the constructor, we could have as an immutable property. Otherwise, having a getter might help too.

javascript/modules/client/src/client-erc20.ts
```diff
@@ -1,703 +1,52 @@
...
      * @memberof ClientErc20
      */
     getPluginInstallItem: (params: IErc20PluginInstall): IPluginInstallItem => {
       return {
-        // id: this._pluginAddress,
         id: PLUGIN_ID,
         data: encodeErc20ActionInit(params),
       };
     },
   };
```
brickpop: Same as before, why not factored out too?

javascript/modules/client/src/client.ts
```diff
@@ -1,947 +1,31 @@
...
+  public encoding: IClientEncoding;
+  public decoding: IClientDecoding;
+  public estimation: IClientEstimation;
+
+  constructor(context: Context) {
+    super(context);
+    this.methods = new IClientMethodsModule(context);
+    this.encoding = new IClientEncodingModule(context);
+    this.decoding = new IClientDecodingModule(context);
+    this.estimation = new IClientEstimationModule(context);
```
brickpop: Same as above

javascript/modules/client/src/internal/interfaces/plugins.ts
```diff
@@ -13,91 +13,114 @@ import { BigNumber } from "@ethersproject/bignumber";
...
-    voteProposal: (params: IVoteProposalParams) => Promise<GasFeeEstimation>;
-    executeProposal: (
-      params: IExecuteProposalParams,
-    ) => Promise<GasFeeEstimation>;
-  };
+export interface IClientAddressList {
+  methods: IClientAddressListMethods;
+  encoding: IClientAddressListEncoding;
+  decoding: IClientAddressListDecoding;
+  estimation: IClientAddressListEstimation;
```
brickpop, line 123: I'm of the impression that we are typing the same thing several times. Maybe it's needed but we're typing these fields here and also typing them on the class itself.

javascript/modules/client/src/internal/utils/client.ts
```diff
@@ -16,6 +23,55 @@ import {
...
+    // {
+    //   receivers: params.mintConfig.map((receiver) => receiver.address),
+    //   amounts: params.mintConfig.map((receiver) => receiver.balance),
+    // },
+    pluginDataBytes,
+    AddressZero, // TODO: Remove when the new contract version is available
+  ];
+}
+
+export function  unwrapDepositParams(
```
brickpop, line 64: Same

javascript/modules/client/src/internal/utils/plugins.ts
```diff
@@ -18,6 +19,22 @@ import {
   VoteValues,
 } from "../interfaces/plugins";
 import { formatEther } from "@ethersproject/units";
+import { IDAO } from "@aragon/core-contracts-ethers";
+
+export function unwrapProposalParams(
```
brickpop, line 24: Unwrap?

javascript/modules/client/src/modules-addressList/decoding.ts
```diff
@@ -0,0 +1,71 @@
...
+  decodeUpdatePluginSettingsAction,
+  getFunctionFragment,
+} from "../internal/encoding/plugins";
+import { IInterfaceParams } from "../internal/interfaces/common";
+import {
+  IClientAddressListDecoding,
+  IPluginSettings,
+} from "../internal/interfaces/plugins";
+
+export class IClientAddressListDecodingModule extends ClientCore
```
brickpop: I suggest not adding "Module"

javascript/modules/client/src/modules-addressList/encoding.ts
```diff
@@ -0,0 +1,96 @@
...
+  encodeUpdatePluginSettingsAction,
+} from "../internal/encoding/plugins";
+import { DaoAction } from "../internal/interfaces/common";
+import {
+  IClientAddressListEncoding,
+  IPluginSettings,
+} from "../internal/interfaces/plugins";
+import { isAddress } from "@ethersproject/address";
+
+export class IClientAddressListEncodingModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-addressList/estimation.ts
```diff
@@ -0,0 +1,81 @@
...
+import { ClientCore } from "../internal/core";
+import { GasFeeEstimation } from "../internal/interfaces/common";
+import {
+  IClientAddressListEstimation,
+  ICreateProposalParams,
+  IExecuteProposalParams,
+  IVoteProposalParams,
+} from "../internal/interfaces/plugins";
+
+export class IClientAddressListEstimationModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-addressList/methods.ts
```diff
@@ -0,0 +1,350 @@
...
+} from "../internal/utils/plugins";
+import { isAddress } from "@ethersproject/address";
+import {
+  QueryAddressListPluginSettings,
+  QueryAddressListProposal,
+  QueryAddressListProposals,
+} from "../internal/graphql-queries";
+import { SortDirection } from "../internal/interfaces/common";
+
+export class IClientAddressListMethodsModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-erc20/decoding.ts
```diff
@@ -0,0 +1,60 @@
...
+  getFunctionFragment,
+} from "../internal/encoding/plugins";
+import { IInterfaceParams } from "../internal/interfaces/common";
+import {
+  IClientErc20Decoding,
+  IMintTokenParams,
+  IPluginSettings,
+} from "../internal/interfaces/plugins";
+
+export class IClientErc20DecodingModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-erc20/encoding.ts
```diff
@@ -0,0 +1,65 @@
...
+} from "../internal/encoding/plugins";
+import { DaoAction } from "../internal/interfaces/common";
+import {
+  IClientErc20Encoding,
+  IMintTokenParams,
+  IPluginSettings,
+} from "../internal/interfaces/plugins";
+import { isAddress } from "@ethersproject/address";
+
+export class IClientErc20EncodingModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-erc20/estimation.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ClientCore } from "../internal/core";
+import { GasFeeEstimation } from "../internal/interfaces/common";
+import {
+  IClientErc20Estimation,
+  ICreateProposalParams,
+  IExecuteProposalParams,
+  IVoteProposalParams,
+} from "../internal/interfaces/plugins";
+
+export class IClientErc20EstimationModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules-erc20/methods.ts
```diff
@@ -0,0 +1,403 @@
...
+import {
+  computeProposalStatusFilter,
+  isProposalId,
+  toErc20Proposal,
+  toErc20ProposalListItem,
+} from "../internal/utils/plugins";
+import { SortDirection } from "../internal/interfaces/common";
+import { delay } from "../internal/temp-mock";
+
+export class IClientErc20MethodsModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules/decoding.ts
```diff
@@ -0,0 +1,112 @@
...
+  IClientDecoding,
+  IFreezePermissionDecodedParams,
+  IGrantPermissionDecodedParams,
+  IMetadata,
+  IRevokePermissionDecodedParams,
+  IWithdrawParams,
+} from "../internal/interfaces/client";
+import { IInterfaceParams } from "../internal/interfaces/common";
+
+export class IClientDecodingModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules/encoding.ts
```diff
@@ -0,0 +1,186 @@
...
+import { isAddress } from "@ethersproject/address";
+import {
+  encodeFreezeAction,
+  encodeGrantActionData,
+  encodeRevokeActionData,
+  encodeUpdateMetadataAction,
+  encodeWithdrawActionData,
+} from "../internal/encoding/client";
+
+export class IClientEncodingModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules/estimation.ts
```diff
@@ -0,0 +1,91 @@
...
+import { ClientCore } from "../internal/core";
+import {
+  IClientEstimation,
+  ICreateParams,
+  IDepositParams,
+} from "../internal/interfaces/client";
+import { GasFeeEstimation } from "../internal/interfaces/common";
+import { unwrapDepositParams } from "../internal/utils/client";
+
+export class IClientEstimationModule extends ClientCore
```
brickpop: Same

javascript/modules/client/src/modules/methods.ts
```diff
@@ -0,0 +1,466 @@
...
+  toDaoDetails,
+  toDaoListItem,
+  toTransfer,
+  unwrapDepositParams,
+} from "../internal/utils/client";
+import { isAddress } from "@ethersproject/address";
+import { Context } from "../context";
+import { ClientCore } from "../internal/core";
+
+export class IClientMethodsModule extends ClientCore implements IClientMethods {
```
brickpop: Same

javascript/modules/client/test/integration/client-addressList-modules/decoding.test.ts
```diff
@@ -0,0 +1,127 @@
+// @ts-ignore
+declare const describe, it, expect;
+
+import {
+  ClientAddressList,
```
brickpop: Could we also group tests by `component/<module>`?

javascript/modules/client/src/internal/utils/client.ts
josemarinas: Nope, this was renamed only on the `internal/encoding` and `internal/encoding` files. We can use the same strategy to rename them if you want and leave them in the `utils` folder or put them in a new one. 

Let me know what do you think

javascript/modules/client/src/client-addressList.ts
josemarinas: So move the client ones to `internal/modules` where the `ipfs`, `web3`, and `graphql` modules live and the other ones to `internal/modules-addressList`?

javascript/modules/client/src/client-addressList.ts
josemarinas: I was just following the naming that we used in the `ipfs`, `web3` and `graphql` modules which is `IpfsModule` `Web3Module` and `GraphqlModule` respectively. 

I would definitely drop the `I` because it is not an interface and does not make sense, `IClientAddressListMethodsModule` => `ClientAddressListMethodsModule` 

About the module at the end if we remove it I would remove it everywhere `class Web3Module` => `class Web3` otherwise we wouldn't be consistent with the naming

javascript/modules/client/src/client-addressList.ts
josemarinas: Because this function is meant be called without initializing the plugin client `ClientAddressList.encoding.getPluginInstallItem()`, if we move it to inside the module we will need to create the client to call it.

The other solution would be creating a `AddressListEncodingStaticModule` just for this function, I thought that this was a bit too much for just one function and that it made sense to leave it there without creating a module specifically for it, but we can go the other way if you prefer it

javascript/modules/client/src/client-erc20.ts
josemarinas: See above

javascript/modules/client/src/internal/interfaces/plugins.ts
josemarinas: Yeah, we can skip typing them in the class, and it will infer the typing

javascript/modules/client/src/client-erc20.ts
josemarinas: Yes I we can have them as 
```ts
readonly methods
readonly encoding
readonly decoding
readonly estimation
```
and won't affect the usage

javascript/modules/client/src/client-erc20.ts
josemarinas: see above

javascript/modules/client/src/internal/utils/plugins.ts
josemarinas: same as above

javascript/modules/client/src/modules-addressList/decoding.ts
josemarinas: same as above

javascript/modules/client/test/integration/client-addressList-modules/decoding.test.ts
josemarinas: Isn't that how are grouped at this moment?
```
- integration/client-modules
| ->  integration/client-modules/encoding.test.ts
| ->  integration/client-modules/decoding.test.ts
| ->  integration/client-modules/estimation.test.ts
| ->  integration/client-modules/methods.test.ts

and so on
```
What do you suggest? Something like this?

```
- integration/client
| ->  integration/client/encoding.test.ts
| ->  integration/client/decoding.test.ts
| ->  integration/client/estimation.test.ts
| ->  integration/client/methods.test.ts
| ->  integration/client/client.test.ts

and so on
```

javascript/modules/client/src/client-addressList.ts
brickpop: My point here is that if this is a static function, then we could have the implementation as a standalone function implemented alongside the module class. 

Something like:
```ts
import { getPluginInstallItem } from ".../modules/....";

class Client {
  // ...

  static encoding = {
    getPluginInstallItem: (...params) => getPluginInstallItem(...params)
  }
}
```

javascript/modules/client/src/internal/utils/client.ts
brickpop: OK, then let's not grow the scope of this PR and keep things as they are

javascript/modules/client/src/client-addressList.ts
brickpop: no, I was meaning something more like:
- `src/internal/client/index.ts`
- `src/internal/client/methods.ts`
- `src/internal/client/encoding.ts`
- ...
- `src/internal/client-erc20/index.ts`
- `src/internal/client-erc20/methods.ts`
- `src/internal/client-erc20/encoding.ts`
- ...

javascript/modules/client/src/client-addressList.ts
brickpop: Well, we should use interfaces for typing class methods, otherwise we're going to have issues if we type with classes.
But as per class names, they should definitely be without an `I`. 

Regarding the `Module` suffix, it is good that we have consistency, but GraphQL/IPFS/Web3 are a different kind of concept than "Encoding", etc. These are groups of functions of the same type, I don't think `Module` helps understanding much here, even more given the fact that this is internal code that external users should not be working with.

javascript/modules/client/src/client-erc20.ts
brickpop: Whatever that gives us 100% certainty that no external user will be able to reimplement stuff and front run the SDK. 

Is `readonly` a TS compile-time only thing? Or is it native pure JS that actually prevents writing to? If not, I would define a getter reading private properties behind a `Symbol`. But maybe this is scope for another PR

javascript/modules/client/src/internal/interfaces/plugins.ts
brickpop: As you see fit, no personal preference

javascript/modules/client/test/integration/client-addressList-modules/decoding.test.ts
brickpop: My point is about the path names with dashes, I would have folder hierarchies grouping things more than `client-addressList-modules`. 

Low priority comment, it's just about testing ergonomics

javascript/modules/client/src/internal/utils/client.ts
josemarinas: https://aragonassociation.atlassian.net/browse/APP-1138

javascript/modules/client/src/client-addressList.ts
```diff
@@ -1,640 +1,52 @@
...
   constructor(context: ContextPlugin) {
     super(context);
+    this.methods = new ClientAddressListMethods(context);
+    this.encoding = new ClientAddressListEncoding(context);
+    this.decoding = new ClientAddressListDecoding(context);
+    this.estimation = new ClientAddressListEstimation(context);
+    Object.freeze(this.methods);
+    Object.freeze(this.encoding);
+    Object.freeze(this.decoding);
+    Object.freeze(this.estimation);
```
brickpop: - Can't we use `readonly` to make the fields readonly? Or having this as a private variable and adding a getter
- IMO, the `Object.freeze` would make sense with the class prototype, to prevent external mutations of new instances as well

javascript/modules/client/src/internal/client-addressList/decoding.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Decodes a dao metadata from an encoded update metadata action
+   *
+   * @param {Uint8Array} data
+   * @return {*}  {IPluginSettings}
+   * @memberof ClientAddressListDecoding
+   */
+  public updatePluginSettingsAction(data: Uint8Array): IPluginSettings {
+    return decodeUpdatePluginSettingsAction(data);
```
brickpop, line 30: Maybe this goes beyond the scope of this PR, but now that this code is in a class with higher ergonomics, why don't we bring the internal code from `encoding/plugins` here instead of having all dummy/proxy class methods?

If this creates code duplication, then it's fine like it is

javascript/modules/client/src/internal/client-addressList/encoding.ts
```diff
@@ -0,0 +1,117 @@
...
+} from "../encoding/plugins";
+import { DaoAction, IPluginInstallItem } from "../interfaces/common";
+import {
+  IAddressListPluginInstall,
+  IClientAddressListEncoding,
+  IPluginSettings,
+} from "../interfaces/plugins";
+import { isAddress } from "@ethersproject/address";
+
+const PLUGIN_ID = "0x1234567890123456789012345678901234567890";
```
brickpop: This may be out of the scope, but this PLUGIN_ID could be defined in a central  `constants.ts` file, and imported here. Otherwise, getting visibility of these values will be hard

javascript/modules/client/src/internal/client-erc20/decoding.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  /**
+   * Decodes the mint token params from an encoded mint token action
+   *
+   * @param {Uint8Array} data
+   * @return {*}  {IMintTokenParams}
+   * @memberof ClientErc20Decoding
+   */
+  public mintTokenAction(data: Uint8Array): IMintTokenParams {
+    return decodeMintTokenAction(data);
```
brickpop, line 39: Same as before, the final code might be here if it is unique

javascript/modules/client/src/internal/client/decoding.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Decodes the freeze parameters from an encoded freeze action
+   *
+   * @param {Uint8Array} data
+   * @return {*}  {IFreezePermissionDecodedParams}
+   * @memberof ClientDecoding
+   */
+  public freezeAction(data: Uint8Array): IFreezePermissionDecodedParams {
+    return decodeFreezeActionData(data);
+  }
```
brickpop, line 56: Same

javascript/modules/client/test/integration/client-addressList/client-addressList.test.ts
```diff
@@ -0,0 +1,60 @@
+// @ts-ignore
```
brickpop: Isn't the name of the file repetitive? Not a big deal though

We could simply call it `client` or even `index`

javascript/modules/client/src/client-addressList.ts
josemarinas: I couldn't find the `readonly` property in the class [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) for JavaScript so I guess they are TS only, I'll change them to private with a getter

javascript/modules/client/src/internal/client-addressList/decoding.ts
josemarinas: Is going to create duplication in the plugins because some methods are shared `encodeUpdatePluginSettingsAction` and `decodeUpdatePluginSettingsAction` specifically. Maybe It would mad more sense once we divide the packages of the `client-erc20` and `client-addresslist`

brickpop: If you think that dividing these clients into independent packages can be a good solution, now is a good time to do it :+1: 

Also, in the case where the logic is the exact same, it makes sense to factor the common code out. But if they have, for example, different plugin parameters, then I would have such logic in place.

javascript/modules/client/src/client-addressList.ts
```diff
@@ -1,640 +1,61 @@
...
+    return this.privateMethods;
+  }
+  get encoding(): IClientAddressListEncoding {
+    return this.privateEncoding;
+  }
+  get decoding(): IClientAddressListDecoding {
+    return this.privateDecoding;
+  }
+  get estimation(): IClientAddressListEstimation {
+    return this.privateEstimation;
```
brickpop, line 44: Private variables can still be changed. The proposal below may be an overkill and still not add much protection, just for inspiration:

```suggestion

const privateMethods = Symbol("privateMethods");
const privateEncoding = Symbol("privateEncoding");
const privateDecoding = Symbol("privateDecoding");
const privateEstimation = Symbol("privateEstimation");

/**
 * Provider a generic client with high level methods to manage and interact an Address List Voting plugin installed in a DAO
 */
export class ClientAddressList extends ClientCore
  implements IClientAddressList {
  private [privateMethods]: IClientAddressListMethods = new ClientAddressListMethods(context);
  private [privateEncoding]: IClientAddressListEncoding = new ClientAddressListEncoding(context);
  private [privateDecoding]: IClientAddressListDecoding = new ClientAddressListDecoding(context);
  private [privateEstimation]: IClientAddressListEstimation = new ClientAddressListEstimation(context);

  constructor(context: ContextPlugin) {
    super(context);
  }
  get methods(): IClientAddressListMethods {
    return this[privateMethods];
  }
  get encoding(): IClientAddressListEncoding {
    return this[privateEncoding];
  }
  get decoding(): IClientAddressListDecoding {
    return this[privateDecoding];
  }
  get estimation(): IClientAddressListEstimation {
    return this[privateEstimation];
```

javascript/modules/client/src/internal/client-addressList/decoding.ts
josemarinas: Yes, maybe we can edit the [APP-1026](https://aragonassociation.atlassian.net/browse/APP-1026), because this was supposed to be to move files from address list and erc20 clients to a separate folder and increase a bit the scope of the task to make independent packages and address this kind of problems

javascript/modules/client/test/integration/client-addressList/decoding.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import {
+  ClientAddressList,
+  Context,
+  ContextPlugin,
+  IPluginSettings,
+} from "../../../src";
+import { contextParamsLocalChain } from "../constants";
+
+describe("Action decoders", () => {
```
brickpop: Shouldn't we add a higher level `describe("Client Address List")` to encapsulate these tests together?
Same everywhere

javascript/modules/client/test/integration/client-erc20/decoding.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import { ClientErc20, Context, ContextPlugin } from "../../../src";
+
+import {
+  IMintTokenParams,
+  IPluginSettings,
+} from "../../../src/internal/interfaces/plugins";
+import { bytesToHex } from "@aragon/sdk-common";
+import { contextParamsLocalChain } from "../constants";
+
+describe("Client ERC20 decoders", () => {
```
brickpop: Let's have a common parent for the 4 blocks

```suggestion
describe("Client ERC20", () => {
```

javascript/modules/client/test/integration/client-erc20/encoding.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import {
+  IErc20PluginInstall,
+  IMintTokenParams,
+  IPluginSettings,
+} from "../../../src/internal/interfaces/plugins";
+import { AddressZero } from "@ethersproject/constants";
+import { InvalidAddressError } from "@aragon/sdk-common";
+import { contextParamsLocalChain } from "../constants";
+describe("Client ERC20 decoders", () => {
```
brickpop:
```suggestion
describe("Client ERC20", () => {
```

javascript/modules/client/test/integration/client-erc20/estimation.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import {
+  ICreateProposalParams,
+  IExecuteProposalParams,
+  IVoteProposalParams,
+  VoteValues,
+} from "../../../src/internal/interfaces/plugins";
+import { contextParamsLocalChain } from "../constants";
+import * as ganacheSetup from "../../../../../helpers/ganache-setup";
+import * as deployContracts from "../../../../../helpers/deployContracts";
+describe("Client ERC20 decoders", () => {
```
brickpop:
```suggestion
describe("Client ERC20", () => {
```

javascript/modules/client/test/integration/client-erc20/index.test.ts
```diff
@@ -0,0 +1,59 @@
...
+declare const describe, it, expect;
+
+import { JsonRpcProvider } from "@ethersproject/providers";
+import { Wallet } from "@ethersproject/wallet";
+import { ClientErc20, Context, ContextPlugin } from "../../../src";
+import { Client as IpfsClient } from "@aragon/sdk-ipfs";
+import { GraphQLClient } from "graphql-request";
+import { contextParams, contextParamsFailing } from "../constants";
+
+describe("Client ERC20 decoders", () => {
```
brickpop:
```suggestion
describe("Client ERC20", () => {
```

javascript/modules/client/test/integration/client-erc20/methods.test.ts
```diff
@@ -0,0 +1,430 @@
...
+  contextParamsLocalChain,
+  TEST_DAO_ADDRESS,
+  TEST_ERC20_DAO_ADDRESS,
+  TEST_ERC20_PLUGIN_ADDRESS,
+  TEST_ERC20_PROPOSAL_ID,
+  TEST_INVALID_ADDRESS,
+  TEST_NON_EXISTING_ADDRESS,
+} from "../constants";
+
+describe("Client ERC20 decoders", () => {
```
brickpop:
```suggestion
describe("Client ERC20", () => {
```

brickpop (approved):

PR #91, opened 2022-10-25
Feature: Split plugin files into separate folders

josemarinas:
## Description

Refactor of the code splitting both plugin clients (`erc20` and `addressList`)
- Move `AddressListClient` files to `client/src/client-addressList`
- Move `ERC20Client` files to `client/src/client-erc20`
- Move common method between `addressListclient` and `erc20-client` to `plugins-common`

List of points to be resolved:
- `plugins-common` should be a new package npm page or should it be merged with `common`?
- The 3 clients share a dependency for both `Context` and `ClientCore`, since `ContextPlugin` is in `plugin-common` if we decide to merge `plugins-common` with `common` this will create a circular dependency between `@aragon/sdk-client` and `@aragon/sdk-common`. Possible solutions:
  - Create an independent package for `plugin-common`
  - Move `plugin-common`, `Context` and `ClientCore` to `@aragon/sdk-common`

Task: [APP-1026](https://aragonassociation.atlassian.net/browse/APP-1026)

## Type of change

## Checklist:


brickpop:
Replying to some of your comments:
> plugins-common should be a new package npm page or should it be merged with common?

It could make sense to have this as part of `@aragon/sdk-common` maybe?
This would also solve the other problem mentioned

javascript/modules/client/src/client-addressList/index.ts
```diff
@@ -0,0 +1 @@
+export * from './client-addressList'
```
brickpop: Couldn't this file simply hold the contents of client-addressList?

javascript/modules/client/src/client-addressList/internal/client-addressList/decoding.ts
```diff
@@ -1,17 +1,17 @@
 import { bytesToHex } from "@aragon/sdk-common";
```
brickpop: Isn't it a bit redundant to have the file on `client-addressList/.../client-addressList/*` ?

javascript/modules/client/src/client-addressList/internal/constant.ts
```diff
@@ -0,0 +1,3 @@
+// NOTE: This address needs to be set when the plugin has been published and the ID is known
```
brickpop: I would name it "constants" in plural

javascript/modules/client/src/client-erc20/client-erc20.ts
```diff
@@ -5,14 +5,14 @@ import {
   IClientErc20Estimation,
```
brickpop: Couldn't this file be named `client.ts`  now that it belongs to the `client-erc20` folder?

javascript/modules/client/src/client-erc20/internal/client-erc20/decoding.ts
```diff
@@ -1,17 +1,14 @@
 import { bytesToHex } from "@aragon/sdk-common";
```
brickpop: Same as above. Even if it belongs to `internal` I don't see why duplicating `client-erc20` in the path

javascript/modules/client/src/index.ts
```diff
@@ -1,8 +1,8 @@
 export * from "./client";
-export * from "./client-erc20";
-export * from "./client-addressList";
+export * from "./client-addressList/index";
+export * from "./client-erc20/index";
```
brickpop: `/index` is always inferred automatically. Unless there was a file called `addressList` in the parent folder, which would be an anti-pattern to avoid

javascript/modules/client/src/index.ts
```diff
@@ -40,23 +40,25 @@ export {
...
+// IMintTokenParams,
+// IPluginSettings,
+// IProposalQueryParams,
+// IProposalSettings,
+// IVoteProposalParams,
+// ProposalCreationSteps,
+// ProposalSortBy,
+// ProposalStatus,
+// VoteProposalStep,
+// VoteValues,
```
brickpop: Why would we want to keep this? Remove after the whole refactor is done

javascript/modules/client/src/internal/client/decoding.ts
```diff
@@ -8,7 +8,7 @@ import {
   decodeUpdateMetadataAction,
   decodeWithdrawActionData,
   getFunctionFragment,
-} from "../encoding/client";
+} from "../encoding";
```
brickpop: How are we importing "decodeXXX" functions from a file called `encoding`?

javascript/modules/client/src/internal/constants.ts
```diff
@@ -1,4 +1,9 @@
...
-// NOTE: This address needs to be set when the plugin has been published and the ID is known
-export const ADDRESSLIST_PLUGIN_ID = "0x1234567890123456789012345678901234567890";
+import { DAO__factory } from "@aragon/core-contracts-ethers";
+
+export const AVAILABLE_CLIENT_FUNCTION_SIGNATURES: string[] = [
+  DAO__factory.createInterface().getFunction("withdraw").format("minimal"),
+  DAO__factory.createInterface().getFunction("setMetadata").format("minimal"),
+  DAO__factory.createInterface().getFunction("revoke").format("minimal"),
+  DAO__factory.createInterface().getFunction("grant").format("minimal"),
+  DAO__factory.createInterface().getFunction("freeze").format("minimal"),
```
brickpop, line 8: Where have all these function ID's have been moved from? `constants.ts` might make less sense than something like `abi.ts`, `functions.ts` or similar... but it's just a suggestion

javascript/modules/client/src/internal/graphql-queries/index.ts
```diff
@@ -1,7 +1,5 @@
 export * from "./dao";
 export * from "./status";
 export * from "./balances";
-export * from "./proposal";
-export * from "./settings";
-export * from "./token";
+export * from "../../client-erc20/internal/graphql-queries/token";
```
brickpop: Exporting plugin-based queries like this on a PR called "split" might be a bit confusing

javascript/modules/client/src/plugin-common/encoding.ts
```diff
@@ -0,0 +1,64 @@
...
+    BigNumber.from(encodeRatio(params.minTurnout, 2)),
+    BigNumber.from(encodeRatio(params.minSupport, 2)),
+    BigNumber.from(params.minDuration),
+  ];
+}
+
+export function getFunctionFragment(data: Uint8Array): FunctionFragment {
+  const hexBytes = bytesToHex(data, true);
+  const inter = new Interface(AVAILABLE_PLUGIN_FUNCTION_SIGNATURES);
+  return inter.getFunction(hexBytes.substring(0, 10));
```
brickpop: Probably unrelated to the PR, but... doesn't this function potentially return `null`?

javascript/modules/client/src/plugin-common/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from "./interfaces";
+export * from "./context-plugin";
+export * from "./constants";
+export * from "./constants";
```
brickpop: Twice?

javascript/modules/client/test/integration/client-erc20/methods.test.ts
```diff
@@ -1,29 +1,10 @@
...
 declare const describe, it, beforeAll, afterAll, expect, test;
 
-import {
-  Client,
-  ClientErc20,
-  Context,
-  ContextPlugin,
-  SortDirection,
-} from "../../../src";
+import { Client, ClientErc20, Context, ContextPlugin, ExecuteProposalStep, ICanVoteParams, ICreateProposalParams, IExecuteProposalParams, IProposalQueryParams, IVoteProposalParams, ProposalCreationSteps, ProposalSortBy, ProposalStatus, SortDirection, VoteProposalStep, VoteValues } from "../../../src";
```
brickpop: Did `deno fmt` do this?

javascript/modules/client/src/internal/client/decoding.ts
josemarinas: Yea I does not make a lot of sense, this folder hast the functions to encode and decode de actions, maybe we can split it into `encoding` and `decoding`, let me know what do you think

javascript/modules/client/src/internal/constants.ts
josemarinas: They have been moved from `contsnts/encoding`, maybe we can rename it, but for me rn it makes sense, in the end it is a constant with the available functions.

javascript/modules/client/src/plugin-common/encoding.ts
josemarinas: No, It always returns a function fragment, I believe it throws if it does not find the function

javascript/modules/client/src/internal/client/decoding.ts
brickpop: it might make more sense logically, but it can also make things confusing. What do you think?

javascript/modules/client/src/client-addressList/internal/client/decoding.ts
```diff
@@ -1,17 +1,18 @@
 import { bytesToHex } from "@aragon/sdk-common";
```
brickpop: Do we need to have a path like `<plugin>/internal/client/decoding.ts`? In my opinion `/client` is already implied.... unless there were too many files on the `internal` folder of that plugin, I would flatten it

javascript/modules/client/src/client-addressList/internal/client/encoding.ts
```diff
@@ -1,20 +1,23 @@
 import { InvalidAddressError } from "@aragon/sdk-common";
```
brickpop: Same as above

javascript/modules/client/src/client-addressList/internal/encoding.ts
```diff
@@ -0,0 +1,106 @@
+import { WhitelistVoting__factory } from "@aragon/core-contracts-ethers";
```
brickpop: Why do we have `<plugin>/internal/encoding.ts` as well as `<plugin>/internal/client/encoding.ts`?  (applies to the rest)

javascript/modules/client/src/client-erc20/internal/client/decoding.ts
```diff
@@ -1,17 +1,15 @@
 import { bytesToHex } from "@aragon/sdk-common";
```
brickpop: Same as before, maybe it is OK to have `/client` as an extra folder level... but we already have `client-erc20` itself

javascript/modules/client/src/internal/client/decoding.ts
josemarinas: I would leave it as it is and if it grows too much, split it

javascript/modules/client/src/client-addressList/internal/client/decoding.ts
josemarinas: Yes thats a different thing, `/internal/client/encoding.ts` refers to the encoding class of the client. Maybe it's a bit weird due to the naming, but they are totally different thing and should be separated imo. Maybe we could reconsider the `modules` naming instead of client

brickpop: ok, how about naming the module like `plugin-erc20` and `plugin-addressList`?

brickpop (approved):

PR #93, opened 2022-11-02
Feature: Prevent malicious mutation of global class prototypes

josemarinas:
## Description

Freeze `this` and `<class>.proptotype` to make the classes not modifiable after instantiating them

Task: [APP-1133](https://aragonassociation.atlassian.net/browse/APP-1133)

## Type of change

## Checklist:

### DIFF
--- a/modules/client/src/client-common/core.ts
+++ b/modules/client/src/client-common/core.ts
@@ -9,28 +9,27 @@ import { Web3Module } from "./modules/web3";
 import { IPFSModule } from "./modules/ipfs";
 import { GraphqlModule } from "./modules/graphql";
 
+const web3Map = new Map<ClientCore, IClientWeb3Core>();
+const ipfsMap = new Map<ClientCore, IClientIpfsCore>();
+const graphqlMap = new Map<ClientCore, IClientGraphQLCore>();
 /**
  * Provides the low level foundation so that subclasses have ready-made access to Web3, IPFS and GraphQL primitives
  */
 export abstract class ClientCore implements IClientCore {
-  private privateWeb3: IClientWeb3Core;
-  private privateIpfs: IClientIpfsCore;
-  private privateGraphql: IClientGraphQLCore;
-
   constructor(context: Context) {
-    this.privateWeb3 = new Web3Module(context);
-    this.privateIpfs = new IPFSModule(context);
-    this.privateGraphql = new GraphqlModule(context);
+    web3Map.set(this, new Web3Module(context));
+    ipfsMap.set(this, new IPFSModule(context));
+    graphqlMap.set(this, new GraphqlModule(context));
     Object.freeze(ClientCore.prototype);
   }
 
   get web3(): IClientWeb3Core {
-    return this.privateWeb3;
+    return web3Map.get(this)!;
   }
   get ipfs(): IClientIpfsCore {
-    return this.privateIpfs;
+    return ipfsMap.get(this)!;
   }
   get graphql(): IClientGraphQLCore {
-    return this.privateGraphql;
+    return graphqlMap.get(this)!;
   }
 }
--- a/modules/client/src/client-common/core.ts
+++ b/modules/client/src/client-common/core.ts
@@ -13,13 +13,24 @@ import { GraphqlModule } from "./modules/graphql";
  * Provides the low level foundation so that subclasses have ready-made access to Web3, IPFS and GraphQL primitives
  */
 export abstract class ClientCore implements IClientCore {
-  public web3: IClientWeb3Core;
-  public ipfs: IClientIpfsCore;
-  public graphql: IClientGraphQLCore;
+  private privateWeb3: IClientWeb3Core;
+  private privateIpfs: IClientIpfsCore;
+  private privateGraphql: IClientGraphQLCore;
 
   constructor(context: Context) {
-    this.web3 = new Web3Module(context);
-    this.ipfs = new IPFSModule(context);
-    this.graphql = new GraphqlModule(context);
+    this.privateWeb3 = new Web3Module(context);
+    this.privateIpfs = new IPFSModule(context);
+    this.privateGraphql = new GraphqlModule(context);
+    Object.freeze(ClientCore.prototype);
+  }
+
+  get web3(): IClientWeb3Core {
+    return this.privateWeb3;
+  }
+  get ipfs(): IClientIpfsCore {
+    return this.privateIpfs;
+  }
+  get graphql(): IClientGraphQLCore {
+    return this.privateGraphql;
   }
 }
--- a/modules/client/src/client-common/modules/graphql.ts
+++ b/modules/client/src/client-common/modules/graphql.ts
@@ -12,6 +12,7 @@ export class GraphqlModule implements IClientGraphQLCore {
     if (context.graphql?.length) {
       this._graphql = context.graphql;
       this._graphqlIdx = Math.floor(Random.getFloat() * context.graphql.length);
+      Object.freeze(GraphqlModule.prototype);
     }
   }
 
--- a/modules/client/src/client-common/modules/graphql.ts
+++ b/modules/client/src/client-common/modules/graphql.ts
@@ -4,16 +4,25 @@ import { Context } from "../../client-common/context";
 import { QueryStatus } from "../graphql-queries";
 import { IClientGraphQLCore } from "../interfaces/core";
 
-export class GraphqlModule implements IClientGraphQLCore {
-  private _graphql: GraphQLClient[] = [];
-  private _graphqlIdx: number = -1;
+enum Keys {
+  Graphql,
+  GraphqlIdx,
+}
+const GraphqlModuleMap = new Map<Keys, any>([
+  [Keys.Graphql, [] as GraphQLClient[]],
+  [Keys.GraphqlIdx, -1],
+]);
 
+export class GraphqlModule implements IClientGraphQLCore {
   constructor(context: Context) {
     if (context.graphql?.length) {
-      this._graphql = context.graphql;
-      this._graphqlIdx = Math.floor(Random.getFloat() * context.graphql.length);
-      Object.freeze(GraphqlModule.prototype);
+      GraphqlModuleMap.set(Keys.Graphql, context.graphql);
+      GraphqlModuleMap.set(
+        Keys.Graphql,
+        Math.floor(Random.getFloat() * context.graphql.length),
+      );
     }
+    Object.freeze(GraphqlModule.prototype);
   }
 
   /**
@@ -22,23 +31,27 @@ export class GraphqlModule implements IClientGraphQLCore {
    * @returns {GraphQLClient}
    */
   public getClient(): GraphQLClient {
-    if (!this._graphql[this._graphqlIdx]) {
+    const graphql = GraphqlModuleMap.get(Keys.Graphql);
+    const graphqlIdx = GraphqlModuleMap.get(Keys.GraphqlIdx);
+    if (!graphql[graphqlIdx]) {
       throw new Error("No graphql endpoints available");
     }
-    return this._graphql[this._graphqlIdx];
+    return graphql[graphqlIdx];
   }
 
   /**
-   * Starts using the next available IPFS endpoint
+   * Starts using the next available Graphql endpoint
    * @returns {void}
    */
   public shiftClient(): void {
-    if (!this._graphql?.length) {
+    const graphql = GraphqlModuleMap.get(Keys.Graphql);
+    const graphqlIdx = GraphqlModuleMap.get(Keys.GraphqlIdx);
+    if (!graphql?.length) {
       throw new Error("No graphql endpoints available");
-    } else if (this._graphql?.length < 2) {
+    } else if (graphql?.length < 2) {
       throw new Error("No other endpoints");
     }
-    this._graphqlIdx = (this._graphqlIdx + 1) % this._graphql?.length;
+    GraphqlModuleMap.set(Keys.GraphqlIdx, (graphqlIdx + 1) % graphql?.length);
   }
 
   /**
@@ -64,11 +77,12 @@ export class GraphqlModule implements IClientGraphQLCore {
    * @returns {Promise<void>}
    */
   public async ensureOnline(): Promise<void> {
-    if (!this._graphql?.length) {
+    const graphql = GraphqlModuleMap.get(Keys.Graphql);
+    if (!graphql?.length) {
       return Promise.reject(new Error("graphql client is not initialized"));
     }
 
-    for (let i = 0; i < this._graphql?.length; i++) {
+    for (let i = 0; i < graphql?.length; i++) {
       if (await this.isUp()) return;
 
       this.shiftClient();
--- a/modules/client/src/client-common/modules/graphql.ts
+++ b/modules/client/src/client-common/modules/graphql.ts
@@ -4,39 +4,42 @@ import { Context } from "../../client-common/context";
 import { QueryStatus } from "../graphql-queries";
 import { IClientGraphQLCore } from "../interfaces/core";
 
-enum Keys {
-  Graphql,
-  GraphqlIdx,
-}
-const GraphqlModuleMap = new Map<Keys, any>([
-  [Keys.Graphql, [] as GraphQLClient[]],
-  [Keys.GraphqlIdx, -1],
-]);
+const GraphqlMap = new Map<GraphqlModule, GraphQLClient[]>();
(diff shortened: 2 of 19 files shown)


javascript/modules/client/src/client-common/modules/graphql.ts
```diff
@@ -12,6 +12,7 @@ export class GraphqlModule implements IClientGraphQLCore {
     if (context.graphql?.length) {
       this._graphql = context.graphql;
       this._graphqlIdx = Math.floor(Random.getFloat() * context.graphql.length);
+      Object.freeze(GraphqlModule.prototype);
```
brickpop: Why not freezing out of the `if` clause? if there's no client, someone could inject one

javascript/modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -11,6 +11,7 @@ export class IPFSModule implements IClientIpfsCore {
     if (context.ipfs?.length) {
       this._ipfs = context.ipfs;
       this._ipfsIdx = Math.floor(Random.getFloat() * context.ipfs.length);
+      Object.freeze(IPFSModule.prototype);
```
brickpop: Same

javascript/modules/client/src/client-common/modules/web3.ts
```diff
@@ -32,6 +32,7 @@ export class Web3Module implements IClientWeb3Core {
     if (context.gasFeeEstimationFactor) {
       this._gasFeeEstimationFactor = context.gasFeeEstimationFactor;
     }
+    Object.freeze(Web3Module.prototype);
```
brickpop: Freezing the prototype is fine, but I think someone could still mutate the signer or the provider. 

We could declare a local `Map` that was not exported, in which each class could store its own private components `privSigners.set(this, signer)`, then access them like `privSigners.get(this)`.

The downside is that this would take references and never release them

brickpop: This applies to IPFS and GraphQL as well

modules/client/src/client-common/core.ts
```diff
@@ -13,13 +13,24 @@ import { GraphqlModule } from "./modules/graphql";
...
+  private privateGraphql: IClientGraphQLCore;
 
   constructor(context: Context) {
-    this.web3 = new Web3Module(context);
-    this.ipfs = new IPFSModule(context);
-    this.graphql = new GraphqlModule(context);
+    this.privateWeb3 = new Web3Module(context);
+    this.privateIpfs = new IPFSModule(context);
+    this.privateGraphql = new GraphqlModule(context);
+    Object.freeze(ClientCore.prototype);
```
brickpop, line 23: Why not Object.freeze(this)  as well?

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -4,15 +4,25 @@ import { Context } from "../../client-common/context";
...
-  private _graphql: GraphQLClient[] = [];
-  private _graphqlIdx: number = -1;
+enum Keys {
+  Graphql,
+  GraphqlIdx,
+}
+const GraphqlModuleMap = new Map<Keys, any>([
+  [Keys.Graphql, [] as GraphQLClient[]],
+  [Keys.GraphqlIdx, -1],
+]);
```
brickpop: This will make that only one single class can use it. Any other will collide. Indexing should be made per-instance.

```ts
const _instanceMap = new Map<any, GraphQLClient>();

class GraphqlModule {
  private _graphqlIdx: number = -1;

  constructor() {
    if ( ...) {
     _instanceMap.set(this, context.graphql);
     // ...
    }
  }

  getClient() {
    // ...
    const clients = _instanceMap.get(this);
    return clients[this._graphqlIdx];
  }
}
```

brickpop: The index variable could also be within a map, just for simple reference

modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -3,48 +3,62 @@ import { Context } from "../../client-common/context";
...
 import { Client as IpfsClient } from "@aragon/sdk-ipfs";
 
+enum Keys {
+  IPFS,
+  IPFSIdx,
+}
+const IPFSModuleMap = new Map<Keys, any>([
+  [Keys.IPFS, [] as IpfsClient[]],
+  [Keys.IPFSIdx, -1],
+]);
```
brickpop: Same

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,76 @@ import { Signer } from "@ethersproject/abstract-signer";
...
+}
+const Web3ModuleMap = new Map<Keys, any>([
+  [Keys.DaoFactoryAddress, ""],
+  [Keys.GasFeeEstimationFactor, 1],
+  [Keys.Web3Providers, [] as JsonRpcProvider[]],
+  [Keys.GasFeeEstimationFactor, 1],
+  [Keys.Web3Idx, -1],
+  [Keys.Signer, undefined],
+]);
 export class Web3Module implements IClientWeb3Core {
```
brickpop: Same

modules/client/src/client-common/core.ts
josemarinas: We cannot do that because this class must be extendable, if we freeze `this` it will throw when trying to do `extends ClientCore`

brickpop, line 23: Why not `Object.freeze(this)`?

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -4,40 +4,55 @@ import { Context } from "../../client-common/context";
 import { QueryStatus } from "../graphql-queries";
 import { IClientGraphQLCore } from "../interfaces/core";
 
-export class GraphqlModule implements IClientGraphQLCore {
-  private _graphql: GraphQLClient[] = [];
-  private _graphqlIdx: number = -1;
+const GraphqlMap = new Map<GraphqlModule, GraphQLClient[]>();
+const GraphqlIdxMap = new Map<GraphqlModule, number>();
```
brickpop: I would name them in lower case, potentially even with an underscore. Otherwise, it seems like a Class

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -4,40 +4,55 @@ import { Context } from "../../client-common/context";
...
+      );
     }
+    Object.freeze(GraphqlModule.prototype);
+    Object.freeze(this);
   }
 
+  get graphql(): GraphQLClient[] {
+    return GraphqlMap.get(this) || [];
+  }
+  get graphqlIdx(): number {
```
brickpop: they might be `private`

brickpop: Since this is a GrapnQL focused class, we might call this something like `this.clients` or `this.endpoints`... as well as `this.clientIdx` or `this.idx` or `this.currentIdx`. Otherwise `graphql` means nothing.

modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -3,48 +3,61 @@ import { Context } from "../../client-common/context";
 import { IClientIpfsCore } from "../interfaces/core";
 import { Client as IpfsClient } from "@aragon/sdk-ipfs";
 
-export class IPFSModule implements IClientIpfsCore {
-  private _ipfs: IpfsClient[] = [];
-  private _ipfsIdx: number = -1;
+const IPFSMap = new Map<IPFSModule, IpfsClient[]>();
+const IPFSIdxMap = new Map<IPFSModule, number>();
```
brickpop: Same, lowercase. And maybe a more maningful name like `ipfsClientsMap` and `ipfsClientIdxMap`

modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -3,48 +3,61 @@ import { Context } from "../../client-common/context";
...
+
+  get ipfs(): IpfsClient[] {
+    return IPFSMap.get(this) || [];
+  }
+  get ipfsIdx(): number {
+    const idx = IPFSIdxMap.get(this);
+    if (idx === undefined) {
+      return -1;
     }
+    return idx;
```
brickpop: Same. `this.clients` and `this.clientIdx` or `this.currentIdx`

modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -3,48 +3,61 @@ import { Context } from "../../client-common/context";
...
-  private _ipfsIdx: number = -1;
+const IPFSMap = new Map<IPFSModule, IpfsClient[]>();
+const IPFSIdxMap = new Map<IPFSModule, number>();
 
+export class IPFSModule implements IClientIpfsCore {
   constructor(context: Context) {
     if (context.ipfs?.length) {
-      this._ipfs = context.ipfs;
-      this._ipfsIdx = Math.floor(Random.getFloat() * context.ipfs.length);
+      IPFSIdxMap.set(this, Math.floor(Random.getFloat() * context.ipfs.length));
```
brickpop:
```suggestion
      // Storing client data in the private module's scope to prevent external mutation
      IPFSIdxMap.set(this, Math.floor(Random.getFloat() * context.ipfs.length));
```

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -4,40 +4,55 @@ import { Context } from "../../client-common/context";
...
-  private _graphqlIdx: number = -1;
+const GraphqlMap = new Map<GraphqlModule, GraphQLClient[]>();
+const GraphqlIdxMap = new Map<GraphqlModule, number>();
 
+export class GraphqlModule implements IClientGraphQLCore {
   constructor(context: Context) {
     if (context.graphql?.length) {
-      this._graphql = context.graphql;
-      this._graphqlIdx = Math.floor(Random.getFloat() * context.graphql.length);
+      GraphqlMap.set(this, context.graphql);
```
brickpop:
```suggestion
      // Storing client data in the private module's scope to prevent external mutation
      GraphqlMap.set(this, context.graphql);
```

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,77 @@ import { Signer } from "@ethersproject/abstract-signer";
 import { GasFeeEstimation } from "../../client-common/interfaces/common";
 import { IClientWeb3Core } from "../interfaces/core";
 
+const DaoFactoryAddressMap = new Map<Web3Module, string>();
+const GasFeeEstimationFactorMap = new Map<Web3Module, number>();
+const Web3ProvidersMap = new Map<Web3Module, JsonRpcProvider[]>();
+const Web3IdxMap = new Map<Web3Module, number>();
+const SignerMap = new Map<Web3Module, Signer>();
```
brickpop: Same. Lowercase

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,77 @@ import { Signer } from "@ethersproject/abstract-signer";
 import { GasFeeEstimation } from "../../client-common/interfaces/common";
 import { IClientWeb3Core } from "../interfaces/core";
 
+const DaoFactoryAddressMap = new Map<Web3Module, string>();
+const GasFeeEstimationFactorMap = new Map<Web3Module, number>();
+const Web3ProvidersMap = new Map<Web3Module, JsonRpcProvider[]>();
+const Web3IdxMap = new Map<Web3Module, number>();
```
brickpop: `providerIdxMap`

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,77 @@ import { Signer } from "@ethersproject/abstract-signer";
 import { GasFeeEstimation } from "../../client-common/interfaces/common";
 import { IClientWeb3Core } from "../interfaces/core";
 
+const DaoFactoryAddressMap = new Map<Web3Module, string>();
+const GasFeeEstimationFactorMap = new Map<Web3Module, number>();
+const Web3ProvidersMap = new Map<Web3Module, JsonRpcProvider[]>();
```
brickpop: `providersMap`

modules/client/src/client-common/core.ts
josemarinas: If you freeze `this` you cannot extend the class afterwards and this will crash
```ts
export class MyClass extends ClientCore {
//...
}
```

brickpop: Then we also need to use a `Map` here, because the private field can still be accessed and modified from the outside.

The getters don't add anything here

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -4,40 +4,56 @@ import { Context } from "../../client-common/context";
...
   }
 
+  private get clients(): GraphQLClient[] {
+    return clientsMap.get(this) || [];
+  }
+  private get clientIdx(): number {
+    const idx = clientsIdxMap.get(this);
+    if (idx === undefined) {
+      return -1;
+    }
```
brickpop: Wouldn't it be easier to always set the `idx` to `-1` if there is no client in the constructor? This way, we wouldn't need to test for `undefined` every single time

brickpop: Same applies to the rest

modules/client/src/client-common/modules/ipfs.ts
```diff
@@ -3,48 +3,65 @@ import { Context } from "../../client-common/context";
...
+    // Storing client data in the private module's scope to prevent external mutation
     if (context.ipfs?.length) {
-      this._ipfs = context.ipfs;
-      this._ipfsIdx = Math.floor(Random.getFloat() * context.ipfs.length);
+      clientsIdxMap.set(
+        this,
+        Math.floor(Random.getFloat() * context.ipfs.length),
+      );
+      clientsMap.set(this, context.ipfs);
+    }
```
brickpop, line 19: Initializing `clientsIdxMap` with `-1` otherwise

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,78 @@ import { Signer } from "@ethersproject/abstract-signer";
 import { GasFeeEstimation } from "../../client-common/interfaces/common";
 import { IClientWeb3Core } from "../interfaces/core";
 
+const daoFactoryAddressMap = new Map<Web3Module, string>();
+const gasFeeEstimationFactorMap = new Map<Web3Module, number>();
+const providersMap = new Map<Web3Module, JsonRpcProvider[]>();
+const providersIdxMap = new Map<Web3Module, number>();
```
brickpop:
```suggestion
const providerIdxMap = new Map<Web3Module, number>();
```

modules/client/src/client-common/modules/web3.ts
```diff
@@ -6,56 +6,78 @@ import { Signer } from "@ethersproject/abstract-signer";
...
+  private get daoFactoryAddress(): string {
+    return daoFactoryAddressMap.get(this) || "";
+  }
+  private get gasFeeEstimationFactor(): number {
+    return gasFeeEstimationFactorMap.get(this) || 1;
+  }
+  private get web3Providers(): JsonRpcProvider[] {
+    return providersMap.get(this) || [];
+  }
+  private get web3Idx(): number {
```
brickpop: `providerIdx`? Only if you find it helpful

modules/client/src/client-common/modules/graphql.ts
josemarinas: The problem with that is that if you use `Map.get()` it will return `number || undefined` and if you try to use an oneliner in this return like `return providerIdxMap.get(this) || -1` it wont work because if the map contains a 0 it will equal to `false`

modules/client/src/client-common/core.ts
josemarinas: I don't think so, because we freeze the objects that extend this class, and this class is not used elsewhere

brickpop: yes, we freeze the inner objects on their scope, but not on the root `client` scope. So I could come and do `client["privateWeb3"] = new MyMaliciousWeb3()` without a problem

brickpop: If the only thing exported is the getters, and the getters wrap 3 private maps, then the components cannot be directly accessed

josemarinas: I believe it freezes everything, including the objects and functions in the ClientCore

brickpop (approved):
Creating a follow-up task to add `dispose()` and testing for this

PR #97, opened 2022-11-07
Feature: Use real contracts to create a DAO

josemarinas:
## Description

- Adapt current code to the new contracts
- Update local chain deploy script
  - New contracts
  - Ens
- Update Create dao methods
- Update tests

Task: [APP-507]()

## Type of change

## Checklist:


.github/workflows/javascript_test.yml
```diff
@@ -41,13 +41,11 @@ jobs:
...
       - name: Integration test
-        run: yarn test modules/client/test/integration --ci --coverage
+        run: yarn test modules/client/test/integration --ci --coverage --passWithNoTests
         env:
-          DAO_FACTORY: ${{ steps.contracts.outputs.dao }}
           IPFS_API_KEY: ${{ secrets.IPFS_API_KEY }}
 
       - name: Unit test
-        run: yarn test --testPathIgnorePatterns=modules/client/test/integration --ci --coverage
+        run: yarn test --testPathIgnorePatterns=modules/client/test/integration --ci --coverage  --passWithNoTests
```
brickpop, line 49: Do we still need this?

helpers/deployContracts.ts
```diff
@@ -1,31 +1,277 @@
...
-    .deploy(registry.address, token.address);
-
-  // send ETH to hardcoded wallet in tests
-  await owner.sendTransaction({
-    to: "0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199",
-    value: ethers.utils.parseEther("50.0"),
-  });
-
-  return dao;
+  const owner = provider.getSigner();
```
brickpop:
```suggestion
  const deployWallet = provider.getSigner();
```

helpers/deployContracts.ts
```diff
@@ -1,31 +1,277 @@
...
+}
+
+async function deployPlugin(
+  pluginRepoFactory: aragonContracts.PluginRepoFactory,
+  setupAddr: string,
+  name: string,
+  version: [BigNumberish, BigNumberish, BigNumberish],
+  owner: Signer
+) {
+  const repoaddr = await pluginRepoFactory.callStatic.createPluginRepoWithVersion(
```
brickpop: Will the address match 100% with the analogous transaction executed below below?

helpers/deployContracts.ts
```diff
@@ -1,31 +1,277 @@
...
+) {
+  let daoAddress = await daoFactory.callStatic.createDao(
+    daoSettings,
+    pluginSettings
+  );
+  await daoFactory.createDao(daoSettings, pluginSettings);
+  return daoAddress;
+}
+
+export async function createAllowlistDAO(deployment: Deployment, name: string) {
```
brickpop: Can we have the ERC20 counterpart? Otherwise, let's open a follow-up task

brickpop (commented):
Round 1, tomorrow more

.github/workflows/javascript_test.yml
mathewmeconry: yes there are uint tests

helpers/deployContracts.ts
mathewmeconry: renamed to `deployOwnerWallet` because it is also the owner of some contracts

helpers/deployContracts.ts
mathewmeconry: we can but it wasn't needed in the current tests so there is none in the helper functions.
I'll create the function anyways to test the ERC20Voting createProposal stuff.

helpers/deployContracts.ts
mathewmeconry: in the testing environment yes. because the transactions are not run in parallel

.github/workflows/javascript_test.yml
brickpop: But do we need empty tests at this point?

modules/client/src/addressList/internal/client/decoding.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from "../../../client-common";
 import { AVAILABLE_FUNCTION_SIGNATURES } from "../constants";
 import { IClientAddressListDecoding } from "../../interfaces";
-import { WhitelistVoting__factory } from "@aragon/core-contracts-ethers";
+import { AllowlistVoting__factory } from "@aragon/core-contracts-ethers";
```
brickpop, line 12: Here we have some mismatch, because the SDK calls this plugin AddressList whereas the solidity project calls it AllowList. We agreed to name it like the first in some of the coordinations channels, I guess we'll need a follow-up here

modules/client/src/client-common/context.ts
```diff
@@ -64,12 +64,14 @@ export class Context {
       network: contextParams.network,
       signer: contextParams.signer,
       daoFactoryAddress: contextParams.daoFactoryAddress,
+      daoRegistryAddress: contextParams.daoRegistryAddress,
+      pluginRepoRegistryAddress: contextParams.pluginRepoRegistryAddress,
```
brickpop, line 68: How about `pluginRegistryAddress` ?

modules/client/src/client-common/context.ts
```diff
@@ -163,6 +166,32 @@ export class Context {
...
+   * Getter for daoFactoryAddress property
+   *
+   * @var daoFactoryAddress
+   *
+   * @returns {string}
+   *
+   * @public
+   */
+  get pluginRepoRegistryAddress(): string | undefined {
+    return this.state.pluginRepoRegistryAddress;
```
brickpop, line 192: Similar as above

modules/client/src/client-common/context.ts
```diff
@@ -221,10 +250,10 @@ export class Context {
 
   private static resolveWeb3Providers(
     endpoints: string | JsonRpcProvider | (string | JsonRpcProvider)[],
-    network: Networkish,
+    network: Networkish
```
brickpop, line 253: These formatting glitches look like using a formatter other than `deno` :)

modules/client/src/client-common/interfaces/common.ts
```diff
@@ -14,7 +14,7 @@ export enum DaoRole {
  */
 export interface IPluginInstallItem {
   id: string; // ENS domain or address of the plugin's Repo
-  data: Uint8Array;
+  data: string;
```
brickpop: Why aren't we using a uint8 array anymore?

modules/client/src/client-common/interfaces/context.ts
```diff
@@ -11,6 +11,8 @@ type Web3ContextParams = {
   network: Networkish;
   signer?: Signer;
   daoFactoryAddress?: string;
+  daoRegistryAddress?: string;
+  pluginRepoRegistryAddress?: string;
```
brickpop, line 15: same

modules/client/src/client-common/interfaces/context.ts
```diff
@@ -21,17 +23,18 @@ type GraphQLContextParams = {
...
+  GraphQLContextParams;
 
 // Context state data
 
 type Web3ContextState = {
   network: Networkish;
   signer?: Signer;
   daoFactoryAddress?: string;
+  daoRegistryAddress?: string;
+  pluginRepoRegistryAddress?: string;
```
brickpop, line 37: Same

modules/client/src/client-common/interfaces/core.ts
```diff
@@ -21,6 +21,7 @@ export interface IClientWeb3Core {
     abi: ContractInterface,
   ) => Contract & T;
   getDaoFactoryAddress: () => string;
+  getPluginRepoRegistryAddress: () => string;
```
brickpop, line 24: Didn't we have a getter with a different name in the code from above?

brickpop, line 24: Didn't we have a getter with a different name in the code from above?

modules/client/src/client-common/modules/web3.ts
```diff
@@ -5,61 +5,52 @@ import { Contract, ContractInterface } from "@ethersproject/contracts";
...
   private static readonly PRECISION_FACTOR_BASE = 1000;
+  private providerIdx = -1;
+  private providers: JsonRpcProvider[] = [];
+  private signer?: Signer;
+  private daoFactoryAddress?: string;
+  private daoRegistryAddress?: string;
+  private pluginRepoRegistryAddress?: string;
+  private gasFeeEstimationFactor = 1;
 
   constructor(context: Context) {
```
brickpop: Why are we undoing the whole hardening PR here?

modules/client/src/erc20/internal/client/encoding.ts
```diff
@@ -45,11 +45,9 @@ export class ClientErc20Encoding extends ClientCore
       "initialize",
       args,
     );
-    // Strip 0x => encode in Uint8Array
-    const data = hexToBytes(strip0x(hexBytes));
     return {
       id: ERC20_PLUGIN_ID,
-      data,
+      data: hexBytes,
```
brickpop: I would suggest using the existing types (bigint, Uint8Array, etc) for standard external interoperability. People is free to reencode them however they want outside

brickpop (commented):
Round 2

.github/workflows/javascript_test.yml
mathewmeconry: we have empty integration tests and it does no harm. So lets leave them in there

modules/client/src/addressList/internal/client/decoding.ts
mathewmeconry: Yes, this is one thing that is confusing. Secondly, WhitelistVoting is the base of AllowlistVoting and ERC20Voting.
We cannot name it here AddressList because this is imported from the contracts

modules/client/src/client-common/context.ts
mathewmeconry: It is called `pluginRepoRegistry` in the contracts so I would stick with the same name in the SDK to prevent confusions. aka Allowlist and Addresslist

modules/client/src/client-common/context.ts
mathewmeconry: same as above

modules/client/src/client-common/context.ts
mathewmeconry: I guess it would be good to define the formatter somewhere.... 
In most projects, we use prettier.
And when I run `demo fmt` there are a lot of files not formatted properly
image
and the linter fails with
```
✖ 569 problems (569 errors, 0 warnings)
  569 errors and 0 warnings potentially fixable with the `--fix` option.
```

I suggest we create another task to check which formatter, formatter config, linter and linter config we wanna use.

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: Because it isn't correct. The data is abiEncoded and the default Abiencoder from `ethers` returns a `string` and not a `Uint8Array`

modules/client/src/client-common/interfaces/context.ts
mathewmeconry: same

modules/client/src/client-common/interfaces/context.ts
mathewmeconry: same

modules/client/src/client-common/interfaces/core.ts
mathewmeconry: because I followed the example of `getDaoFactoryAddress()` in `Web3Module`

mathewmeconry: because I followed the example of `getDaoFactoryAddress()` in `Web3Module`

modules/client/src/client-common/modules/web3.ts
mathewmeconry: because the hardening didn't do anything. everything is still exposed publicly and changeable. So the freeze didn't help.
Yes, you cannot change the functionality of a function, but you can get to the maps.

modules/client/src/erc20/internal/client/encoding.ts
mathewmeconry: that is not a typing but rather setting the property data on the return object to the variable hexBytes

modules/client/src/client-common/interfaces/context.ts
novaknole: trully same...

modules/client/src/client-common/modules/web3.ts
mathewmeconry: Additionally, the freeze happens during construction. This means 1. I can still modify the prototype as an attacker before the 1. instance is created. 
Secondly as an attacker I just need to make sure, that I get loaded before the sdk and do `Object.freeze = () => {}`. This prevents anything from getting frozen.

modules/client/src/client-common/context.ts
brickpop: No priority here, just trying to avoid the diff containing formatting changes only. For TS, deno was defined as the formatter in the coding guidelines. For the rest of files, either prettier or no formatting (markdown) is fine.

modules/client/src/client-common/interfaces/common.ts
brickpop: Right, but `ethers` should not be taken for granted. It's a helper that could potentially change (hope not). 
The SDK is ethers agnostic, hence `BigNumber`, `BigInteger`, `BigNumberish`, etc are library-specific.
Only idiomatic JS data types to be exported by an SDK.

modules/client/src/client-common/modules/web3.ts
brickpop: How can you access these maps if they are non-exported variables?

About `Object.freeze = () => {}`, that kind of language design...  :facepalm: 

- A browser extension could trivially do that, but not necessarily access arbitrary memory space of the JS heap to alter the maps.
- An NPM package alongside ours is the easier attack origin, but for this (1) the victim (or us) should import a malicious package and (2) do it before ours is initialized.

Not bulletprof, but still an improvement. The downside is developer complexity (for us), none for the rest of stakeholders IMO.

modules/client/src/erc20/internal/client/encoding.ts
brickpop: ... which has a different data type than the idiomatic one from JS regarding bytes. The SDK should not make sense to us internally, but rather make sense in terms of the person who is using it. 

They should know as little as possible about Ethers, if at all (granted, provider and signer are there, but tops).

Encoded bytes don't need to be human readable like an address, for example.

modules/client/src/client-common/interfaces/core.ts
brickpop: ~~Then I would suggest all following the same convention (getter everywhere when possible). Or in other words, having only one correct way of doing things~~

Improvement for another day. Let's move forward with this and improve it in a non blocking PR :+1:

brickpop: Sorry for the duplicate

modules/client/src/internal/client/methods.ts
```diff
@@ -96,23 +111,55 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+    // start tx
+    const receipt = await tx.wait();
+    // find dao address using the dao registry address
+    const newDaoAddress = receipt.events?.find(
+      e => e.address === daoRegistryAddress
+    )?.topics[1];
+
+    if (!newDaoAddress) {
+      // throw new CannotCreateDaoError()
+      throw new Error("oopsie");
```
brickpop: Oopsie :)

modules/client/src/internal/client/methods.ts
```diff
@@ -96,23 +111,55 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-        "0x1298376517236498176239851762938512359817623985761239486128937461",
+      txHash: tx.hash,
     };
+    // get dao registry address
+    const daoRegistryAddress = await daoFactoryInstance.daoRegistry();
+    // start tx
+    const receipt = await tx.wait();
+    // find dao address using the dao registry address
+    const newDaoAddress = receipt.events?.find(
+      e => e.address === daoRegistryAddress
```
brickpop: Isn't it possible to filter by the event name? this should do the trick too... unless there ever is a second event there

modules/client/src/internal/client/methods.ts
```diff
@@ -96,23 +111,55 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+
+    // await delay(3000);
+    // yield {
+    //   key: DaoCreationSteps.DONE,
+    //   address: "0x6592568247592378465987126349817263958713",
+    // };
+
     // TODO: Uncomment when the new DAO factory is available
 
     /**
```
brickpop: Shouldn't all these comments be cleaned up now?

modules/client/src/internal/utils.ts
```diff
@@ -37,43 +36,43 @@ import { Result } from "@ethersproject/abi";
...
-}
+//   return [
+//     {
+//       name: params.ensSubdomain,
+//       metadata: "ipfs://" + metadataCid,
+//     },
+//     {
+//       // TODO: Adapt the DAO creation parameters
+//       participationRequiredPct: BigInt(10), // BigInt(params.votingConfig.minParticipation),
+//       supportRequiredPct: BigInt(50), // BigInt(params.votingConfig.minSupport),
```
brickpop: Why are we commenting this out? We had this helper which made the logic handlers much simpler to read:

```ts
    const tx = await daoFactoryInstance.createDao(
      ...unwrapCreateDaoParams(params)
);
```

brickpop (commented):
Round 3

modules/client/src/internal/client/methods.ts
```diff
@@ -74,20 +76,33 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      const latestVersion = await PluginRepo__factory.connect(
+        plugin.id,
+        signer
+      ).getLatestVersion();
+      pluginInstallationData.push({
+        pluginSetup: latestVersion[1],
+        pluginSetupRepo: plugin.id,
+        data: plugin.data,
+      });
+    }
```
brickpop: We had `...unwrapCreateDaoParams(params)` doing this work in the code commented below. Why don't we use what we already had and keep things simpler here?

modules/client/src/internal/client/methods.ts
```diff
@@ -74,20 +76,33 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+
     // @ts-ignore
```
brickpop: If the variable is already used, we can remove this

modules/client/src/internal/client/methods.ts
```diff
@@ -300,7 +346,7 @@ export class ClientMethods extends ClientCore implements IClientMethods {
       // this is a temporal fix and should be changed by the line above
       // but the current daos in subgraph dont have a proper metadata
       const stringMetadata = await this.ipfs.fetchString(
-        "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr",
+        "QmebY8BGzWBUyVqZFMaFkFmz3JrfaDoFP5orDqzJ1uiEkr"
```
brickpop, line 333: Should we be using the real data from subgraph or not yet? We need a follow-up task if not.

modules/client/test/integration/client-erc20/encoding.test.ts
```diff
@@ -32,7 +32,7 @@ describe("Client ERC20", () => {
       );
 
       expect(typeof erc20InstallPluginItem).toBe("object");
-      expect(erc20InstallPluginItem.data).toBeInstanceOf(Uint8Array);
+      expect(typeof erc20InstallPluginItem.data).toBe("string");
     });
```
brickpop: Related to the comment above

modules/ipfs/src/internal/helpers.ts
```diff
@@ -164,11 +164,11 @@ export namespace Helpers {
   }
 
   function encodeParams<T>(
-    options: T,
+    options: ArrayLike<T> | {}
```
brickpop: Why is this type changing? The usages I see above use it as an object, not as an array

package.json
```diff
@@ -17,19 +17,22 @@
     "testEnvironment": "./test-environment.js"
   },
   "scripts": {
-    "clean": "rm -Rf node_modules ./modules/*/node_modules ./modules/*/dist",
+    "clean": "rm -Rf node_modules ./modules/*/node_modules ./modules/*/dist ./modules/*/.turbo",
     "lint": "tsdx lint modules/*/src",
     "build": "yarn turbo run build",
-    "test": "TESTING=true tsdx test --verbose --runInBand",
+    "test": "yarn workspaces run test",
```
brickpop, line 23: Does this preserve all the original parameters? Does this have the very same impact?

mathewmeconry: yes works the same

modules/client/src/internal/client/methods.ts
mathewmeconry: I don't think so, because the subgraph is still not deployed and tests are failing if I use it

modules/client/src/internal/client/methods.ts
```diff
@@ -1,17 +1,24 @@
...
   GraphQLError,
   InvalidAddressOrEnsError,
   NoProviderError,
+  NoSignerError,
 } from "@aragon/sdk-common";
 import { Signer } from "@ethersproject/abstract-signer";
 import { BigNumber } from "@ethersproject/bignumber";
 import { AddressZero } from "@ethersproject/constants";
-import { Contract, ContractTransaction } from "@ethersproject/contracts";
-import { ContractReceipt } from "ethers";
```
brickpop, line 14: Great that we removed this here. In general, we should always avoid importing from `ethers` and do it instead from the respective scoped packages.

Is there any other direct import like this? Only scoped ones should be working

modules/client/src/internal/client/methods.ts
```diff
@@ -74,70 +82,76 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-      e => e.address === registryAddress
-    )?.topics[1];
-    if (!newDaoAddress) {
-      return Promise.reject(new Error("Could not create DAO"));
+    const daoFactoryInterface = DAORegistry__factory.createInterface();
+    // find dao address using the dao registry address
+    const log = receipt.logs?.find(
+      e =>
+        e.topics[0] ===
+        id(daoFactoryInterface.getEvent("DAORegistered").format("sighash"))
```
brickpop, line 139: Nice one here. 

I wonder if we can unify things, because below we have this: `const value = cr.events?.find(e => e?.event === "Approval")?.args?.value;`

modules/client/src/internal/client/methods.ts
```diff
@@ -56,6 +62,8 @@ import {
   unwrapDepositParams,
 } from "../utils";
 import { isAddress } from "@ethersproject/address";
+import { toUtf8Bytes } from "@ethersproject/strings";
+import { id } from "ethers/lib/utils";
```
brickpop: Could we use the scoped package here instead of the whole `ethers` thing?

modules/client/src/internal/utils.ts
```diff
@@ -1,16 +1,15 @@
-import { DAOFactory } from "@aragon/core-contracts-ethers";
-import { strip0x } from "@aragon/sdk-common";
+// import { DAOFactory } from "@aragon/core-contracts-ethers";
+// import { strip0x } from "@aragon/sdk-common";
```
brickpop: Let's either remove this or use it

package.json
```diff
@@ -17,19 +17,22 @@
...
     "docs:client": "jsdoc2md --files ./modules/client/src/*.ts ./modules/client/src/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/client.md",
     "docs:addressList": "jsdoc2md --files ./modules/client/src/addressList/*.ts ./modules/client/src/addressList/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/addressList.md",
     "docs:erc20": "jsdoc2md --files ./modules/client/src/erc20/*.ts ./modules/client/src/erc20/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/erc20.md"
   },
   "devDependencies": {
     "@babel/preset-typescript": "^7.18.6",
+    "@ensdomains/ens-contracts": "^0.0.15",
+    "@types/node": "^18.11.9",
+    "ethers": "^5.7.2",
```
brickpop: I suggest we uninstall this and instead use the scoped packages only. Otherwise, we are bloating (yet more) the node_modules dependency hell

package.json
```diff
@@ -17,19 +17,22 @@
...
-    "test": "TESTING=true tsdx test --verbose --runInBand",
+    "test": "yarn workspaces run test",
     "docs:client": "jsdoc2md --files ./modules/client/src/*.ts ./modules/client/src/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/client.md",
     "docs:addressList": "jsdoc2md --files ./modules/client/src/addressList/*.ts ./modules/client/src/addressList/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/addressList.md",
     "docs:erc20": "jsdoc2md --files ./modules/client/src/erc20/*.ts ./modules/client/src/erc20/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/erc20.md"
   },
   "devDependencies": {
     "@babel/preset-typescript": "^7.18.6",
+    "@ensdomains/ens-contracts": "^0.0.15",
+    "@types/node": "^18.11.9",
```
brickpop: Where are these types used? I would be very careful here, because the SDK has to work in browsers, Node is just an internal development tool we use with non-standard and many incompatible API's.

If this dependency gives us false "OK's" in terms of typing/linting, that's a problem.

brickpop (commented):
Round 5

modules/client/src/internal/client/methods.ts
mathewmeconry: replaced now all of them except in the tests

modules/client/src/internal/client/methods.ts
```diff
@@ -105,7 +105,7 @@ export class ClientMethods extends ClientCore implements IClientMethods {
       pluginInstallationData.push({
         pluginSetup: latestVersion[1],
         pluginSetupRepo: plugin.id,
-        data: plugin.data,
+        data: toUtf8String(plugin.data),
```
brickpop, line 108: Why are we passing a binary string here?

modules/client/test/integration/client/methods.test.ts
```diff
@@ -75,9 +76,11 @@ describe("Client", () => {
           plugins: [
             {
               id: deployment.allowListPluginSetup.address,
-              data: ethers.utils.defaultAbiCoder.encode(
-                ["uint64", "uint64", "uint64", "address[]"],
-                [1, 1, 1, []]
+              data: toUtf8Bytes(
```
brickpop, line 78: Same

brickpop (commented):
Round 6

package.json
mathewmeconry: 1. this is in the root and not a module
2. we need ethers for the tests
3. these are dev dependencies and don't end up in the package itself

package.json
mathewmeconry: In the tests

modules/client/src/internal/client/methods.ts
mathewmeconry: because ethers doesn't convert the Uint8Array back to a string and ethers needs it as a string to work
That was why I initially changed it to be a string

modules/client/test/integration/client/methods.test.ts
mathewmeconry: same

package.json
brickpop: When you are developing (and even running the tests on the CI pipeline), the devDependencies are installed as well. You might get the wrong impression that some function will be fine, except that IRL it won't.

What is the NodeJS-only code that we need to run in the tests?

package.json
brickpop: Can't the tests also use the `@ethersproject/xxx` subpackages as well?

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1,7 +1,7 @@
 // @ts-ignore
 declare const describe, it, beforeAll, afterAll, expect, test, fail;
 
-import { JsonRpcProvider, Networkish } from "@ethersproject/providers";
+// import { JsonRpcProvider, Networkish } from "@ethersproject/providers";
```
brickpop:
```suggestion
```

modules/client/src/internal/client/methods.ts
brickpop: Oh, I wasn't aware of that. Ethers accepts many data types as binary data, doesn't it really work here?

Then I would suggest we lave your original solution, sorry for the confusion. Still, I assume the string should be a hex-encoded one, right? not a random binary string from the bytes

modules/client/test/integration/client/methods.test.ts
```diff
@@ -504,126 +529,126 @@ This code creates a dao with the legacy method
...
-    return [new JsonRpcProvider(url.href, network)];
-  } else {
-    return [endpoints];
-  }
-}
+// async function createLegacyDao(params: ContextParams) {
+//   if (!params.web3Providers) {
+//     throw new Error("A provider is needed");
+//   }
+//   if (!params.daoFactoryAddress) {
```
brickpop: Will we need this legacy function in the future? If not, I'd just delete the commented code

brickpop (commented):
Round 7

modules/client/src/internal/client/methods.ts
```diff
@@ -261,12 +260,12 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
       const log = cr.logs?.find(
-        e => e?.topics[0] === id("Approval(owner, spender, value)")
+        e => e?.topics[0] === id("Approval(address,address,uint256)")
       );
-      const value = defaultAbiCoder.decode(
-        ["address", "address", "uint"],
-        log?.data || ""
-      )[2];
+      if (!log) {
+        throw new Error("Failed to find log");
```
brickpop:
```suggestion
        throw new Error("Could not find the Approval event");
```

package.json
mathewmeconry: we need it because we run ganache and need to deploy locally all the contracts
that is nodejs stuff

package.json
mathewmeconry: we need it because we run ganache and need to deploy locally all the contracts
that is nodejs stuff

modules/client/src/internal/client/methods.ts
mathewmeconry: let's leave it as it is.
it is an abi encoded hex string with a 0x prefix needed.

modules/client/test/integration/client/methods.test.ts
mathewmeconry: deleted

modules/client/src/internal/client/methods.ts
brickpop: In my local environment, calling `toUtf8String(new Uint8Array([1,2,3]))` returns `'\x01\x02\x03'` instead of `0x010203`.

Shouldn't it be our helper `bytesToHex()` as you initially added in the encoder?

modules/client/src/internal/utils.ts
```diff
@@ -37,43 +34,43 @@ import { Result } from "@ethersproject/abi";
...
-      })
-      .join("");
+// export function unwrapCreateDaoParams(
+//   params: ICreateParams,
+//   metadataCid: string,
+// ): [DAOFactory.DAOConfigStruct, DAOFactory.VoteConfigStruct, string, string] {
+//   // TODO: Serialize plugin params into a buffer
+//   const pluginDataBytes = "0x" +
+//     params.plugins
+//       .map((entry) => {
```
brickpop: If this wrapper did the work for the legacy contracts, let's either remove or use+adapt this. But not leaving dead comments here, please

modules/client/test/integration/client/methods.test.ts
```diff
@@ -107,7 +114,15 @@ describe("Client", () => {
           },
           ensSubdomain: daoName.toLowerCase().replace(" ", "-"),
           plugins: [
-            { id: "0x1234", data: new Uint8Array([11, 11]) },
+            {
+              id: deployment.allowListRepo.address,
+              data: toUtf8Bytes(
+                ethers.utils.defaultAbiCoder.encode(
```
brickpop: Could we explicitly import the utils.defaultAbiCoder, so that we don't need `ethers` at all?

package.json
brickpop: Please. We are importing the whole `ethers` package because one test uses `ethers.utils.defaultAbiCoder`

package.json
brickpop: Then, can we move this to the `package.json` of the client module only?

modules/client/src/internal/client/methods.ts
mathewmeconry: then it is this way. it works

package.json
mathewmeconry: removed. Just that you are aware of it. `@aragon/core-contracts-ethers` also imports ethers 5.7.2 as a complete package

brickpop (approved):

PR #101, opened 2022-11-14
App 949 sdk use real contract calls to create proposals

mathewmeconry:
## Description

Implements support for AllowlistVoting and ERC20Voting plugins to the SDK. 
Removes the old ERC20Voting and AddresslistVoting implementation.
Fixes npm publish workflow

Tasks: APP-949, APP-1216, APP-1217, APP-1218

## Type of change

## Checklist:


.github/workflows/javascript_publish.yml
```diff
@@ -82,14 +82,3 @@ jobs:
...
-        run: |
-          git fetch
-          git pull
-          git config --global user.name "Arabot-1"
-          git config --global user.email "arabot-1@users.noreply.github.com"
-          sed -i "s/^\#\# \[UPCOMING\]/## [UPCOMING]\n\n\## ${{ needs.checkTag.outputs.version }}\nOn $(date +'%Y-%m-%d %H:%M:%S')/" "./modules/${{ needs.checkTag.outputs.package }}/CHANGELOG.md"
-          git add "./modules/${{ needs.checkTag.outputs.package }}/CHANGELOG.md"
-          git add "./modules/${{ needs.checkTag.outputs.package }}/package.json"
-          git commit -am "Update Changelog and package.json for ${{ needs.checkTag.outputs.package }}"
-          git push
```
brickpop, line 95: Why?

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
+import {
```
brickpop: Why renaming the entire folder and losing track of all history? It's impossible to spot what changed

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
...
+  AllowlistVoting__factory,
+} from "@aragon/core-contracts-ethers";
+import { ClientCore } from "../client-common";
+import { AllowlistVotingContextPlugin } from "./context";
+import { AllowlistVotingDecoding } from "./internal/decoding";
+import { AllowlistVotingEncoding } from "./internal/encoding";
+import { AllowlistVotingEstimation } from "./internal/estimation";
+import { AllowlistVotingMethods } from "./internal/methods";
+
+export class AllowlistVoting extends ClientCore {
```
brickpop: Why renaming the class? We have agreement in naming this plugin AddressList, even if contract+subgraph name changes have not happened yet.

modules/client/src/client-common/interfaces/common.ts
```diff
@@ -1,5 +1,7 @@
 // This file contains common types, interfaces, and enumerations
 
+import { BigNumberish } from "@ethersproject/bignumber";
```
brickpop, line 3: IMO the interfaces we export should be using JS's native data types (Uint8Array). Unless these are only internal for us, and we hand such values to ethers

modules/client/src/client-common/interfaces/common.ts
```diff
@@ -55,3 +48,16 @@ export interface IInterfaceParams {
...
   hash: string;
 }
+
+export interface IEncodingResult {
+  to: string;
+  value: BigNumberish;
+  data: Uint8Array;
+}
+
+export interface IDecodingTX {
```
brickpop: These look more like a type than like an interface to me. If there were methods that could match different implementations under the same umbrella, that would feel good here... but being a precise object I would use a type.

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -40,16 +40,6 @@ export interface IPluginSettings {
...
-export interface ICreateProposalParams {
-  pluginAddress: string;
-  metadata: ProposalMetadata;
-  actions?: DaoAction[];
-  startDate?: Date;
-  endDate?: Date;
-  executeOnPass?: boolean;
-  creatorVote?: VoteValues;
-}
-
```
brickpop, line 52: Why removing it from here?

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -89,22 +79,6 @@ export type ProposalMetadataSummary = {
   summary: string;
 };
 
-// Long version
-export type ProposalBase = {
```
brickpop, line 93: Same

modules/client/src/client.ts
```diff
@@ -3,43 +3,36 @@ import { ClientMethods } from "./internal/client/methods";
...
 /**
  * Provider a generic client with high level methods to manage and interact with DAO's
  */
-export class Client extends ClientCore implements IClient {
-  private privateMethods: IClientMethods;
-  private privateEncoding: IClientEncoding;
-  private privateDecoding: IClientDecoding;
-  private privateEstimation: IClientEstimation;
+export class Client extends ClientCore {
+  private privateMethods: ClientMethods;
```
brickpop, line 12: I'm surprised this works, because passing a class as a type, expects the value to be a class... and not an instance. 

I've had problems with this and my solution has been to use `typeof ClientMethods` to represent instances

modules/client/src/interfaces.ts
```diff
@@ -1,87 +1,12 @@
...
-import { IClientCore } from "./client-common/interfaces/core";
-/** Defines the shape of the general purpose Client class */
-export interface IClientMethods extends IClientCore {
-  create: (params: ICreateParams) => AsyncGenerator<DaoCreationStepValue>;
-  /** Retrieves the asset balances of the given DAO, by default, ETH, DAI, USDC and USDT on Mainnet*/
-  getBalances: (
-    daoAddressOrEns: string,
-    tokenAddresses: string[],
-  ) => Promise<AssetBalance[] | null>;
-  /** Retrieves the list of transfers from or to the given DAO, by default, ETH, DAI, USDC and USDT on Mainnet*/
```
brickpop, line 23: I guess we remove all of this because the client now relates to the class itself?

modules/client/test/integration/client/decoding.test.ts
```diff
@@ -221,24 +206,10 @@ describe("Client", () => {
...
-          {
-            url: "https://twitter.com/...",
-            name: "Twitter",
-          },
-        ],
-      };
+
       const updateMetadataAction = await client.encoding.updateMetadataAction(
         "0x1234567890123456789012345678901234567890",
-        params,
```
brickpop: Why do we remove this feature?  Now the SDK use will have to encode, upload, pin, etc on his own?

brickpop: The goal of the SDK is to act as a one stop shop, now to simply expose low level functions so that developers need to learn about many internal operations

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -87,21 +85,19 @@ describe("Client", () => {
           permission: Permissions.UPGRADE_PERMISSION,
         },
       ];
-      let actions: DaoAction[] = [];
+      let actions: IEncodingResult[] = [];
```
brickpop: Why demoving DAOAction? DAO actions are what developers submit on a proposal, encoded data is a different idea/concept.

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -122,45 +118,27 @@ describe("Client", () => {
...
-            name: "Discord",
-          },
-          {
-            url: "https://twitter.com/...",
-            name: "Twitter",
-          },
-        ],
-      };
-
       const installEntry = await client.encoding.updateMetadataAction(
```
brickpop, line 156: Same

brickpop (commented):
First round. Please, let's keep the scope of this PR to its original goal. Big refactors can be done elsewhere, but now we don't have the bandwidth

.github/workflows/javascript_publish.yml
mathewmeconry: this was the step that failed during the publishing of the previous version. I thought that this won't be needed because I took over the previous pull and there was the changelog already updated with the new version

modules/client/src/allowlistVoting/client.ts
mathewmeconry: because I did a complete rewrite of the code. also, the name of the folder was wrong. The plugin is called AllowlistVoting and not addresslist

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: the reason why I would prefer having bignumber is that it can hold bigger numbers than the native javascript types.
For sure we can return the hex string for a number, but that won't help users of the sdk do calculations on it.
The best example are balances of ERC20 tokens where we have 18 decimals for each coin

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: done

modules/client/src/client-common/interfaces/plugin.ts
mathewmeconry: because proposals are plugin specific and not something standardized in the framework of the core contracts

modules/client/src/client-common/interfaces/plugin.ts
mathewmeconry: same

modules/client/src/client.ts
mathewmeconry: work flawlessly. you only need to use `typeof` when you got out of the typing environment.
For example, if you wanna know if the const `client` is of type `Client`.
But given this are internal these checks aren't necessary and therefore we can use the class as type

modules/client/src/interfaces.ts
mathewmeconry: yes

modules/client/test/integration/client/decoding.test.ts
mathewmeconry: a mistake from where I started to refactor some stuff. 
It will be added back

modules/client/test/integration/client/encoding.test.ts
mathewmeconry: that is not correct. A plugin developer can choose what data it should need in a proposal.
The correct way is to return an "UnsignedTransaction" which could be either executed by the user in the code, by a gas station, can be given to a plugin to execute or used by the `execute()` function in the DAO.

modules/client/test/integration/client/encoding.test.ts
mathewmeconry: a mistake from where I started to refactor some stuff.
It will be added back

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.encoding = new AllowlistVotingEncoding(context.pluginAddress);
+    this.decoding = AllowlistVotingDecoding;
+  }
+
+  public getPluginInstanceWithSigner(): AllowlistVotingContract {
+    const signer = this.web3.getConnectedSigner();
+    if (!signer) {
+      throw new Error("A signer is needed");
+    } else if (!signer.provider) {
+      throw new Error("A web3 provider is needed");
```
josemarinas: Can we use `NoProviderError` from the `common-package`

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.methods = new AllowlistVotingMethods(this);
+    this.estimation = new AllowlistVotingEstimation(this);
+    this.encoding = new AllowlistVotingEncoding(context.pluginAddress);
+    this.decoding = AllowlistVotingDecoding;
+  }
+
+  public getPluginInstanceWithSigner(): AllowlistVotingContract {
+    const signer = this.web3.getConnectedSigner();
+    if (!signer) {
+      throw new Error("A signer is needed");
```
josemarinas: Same but with `NoSignerError`

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
...
+      throw new Error("A web3 provider is needed");
+    }
+
+    return this.pluginInstance.connect(signer);
+  }
+
+  public getPluginInstance(): AllowlistVotingContract {
+    const provider = this.web3.getProvider();
+    if (!provider) {
+      throw new Error("A web3 provider is needed");
```
josemarinas: Use `NoProviderError`

modules/client/src/allowlistVoting/client.ts
```diff
@@ -0,0 +1,49 @@
...
+  public methods: AllowlistVotingMethods;
+  public estimation: AllowlistVotingEstimation;
+  public encoding: AllowlistVotingEncoding;
+  public decoding: AllowlistVotingDecoding;
+
+  constructor(context: AllowlistVotingContextPlugin) {
+    super(context);
+    this.pluginInstance = AllowlistVoting__factory.connect(
+      context.pluginAddress,
+      this.web3.getConnectedSigner()
```
josemarinas: If we are going to use this `pluginInstance` across the whole plugin maybe it makes sense here to do a `this.web3.ensureOnline()`

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: found BitInt https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
Switching to this one

modules/client/src/allowlistVoting/client.ts
mathewmeconry: we ensure to have it online by using `getPluginInstanceWithSigner()` everywhere. the `getConnectedSigner()` is only here to get the attribute initialized

modules/client/src/allowlistVoting/client.ts
mathewmeconry: done

modules/client/src/allowlistVoting/client.ts
mathewmeconry: done

modules/client/src/allowlistVoting/client.ts
mathewmeconry: done

modules/client/src/allowlistVoting/context.ts
```diff
@@ -0,0 +1,20 @@
...
+> {
+  constructor(
+    params: Partial<ContextPluginParams> & IAllowlistVotingContextParams
+  ) {
+    super(params);
+    this.state.pluginAddress = params.pluginAddress;
+  }
+
+  get pluginAddress(): string {
+    return this.state.pluginAddress;
```
josemarinas: AFAIK we agreed to pass the plugin address as a parameter to the functions, following the pattern in the client

.github/workflows/javascript_publish.yml
brickpop: Does this mean that we'll need to edit the version number + date manually in the future?  (not against it, BTW)

modules/client/src/client-common/interfaces/common.ts
brickpop: Right, the `bigint` type is the native JavaScript one. Browser support is basically everywhere at this point. You can multiply a 18 digit number by another 18 digit number in the browser's console and it handles it fine:

```
> 1000000000000000000n * 1000000000000000000n
1000000000000000000000000000000000000n 
```

modules/client/src/client-common/interfaces/plugin.ts
brickpop: Technically it makes sense, but at the same time, our own governance plugins are going to have identical traits, also on the UI side. Third party plugin developers are free to not use them, but if we want their plugins to be compatible our App, we do need to export the definition of how a proposal is represented on the UI side.

The same applies with future Finance interfaces, Membership interfaces, etc. Even if they have no used as of today.

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -0,0 +1,470 @@
...
+    this.allowlistVoting = allowlistVoting;
+  }
+
+  /**
+   * Returns the hash for the MODIFY_ALLOWLIST_PERMISSION permission
+   *
+   * @return {*}  {Promise<string>}
+   * @memberof AllowlistVotingMethods
+   */
+  public MODIFY_ALLOWLIST_PERMISSION_ID(): Promise<string> {
```
josemarinas: Atm, we have a constant with the permissions so we dont have to await to get a permission, we can move them to this way, but if we do it I'd also do it in the client

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -0,0 +1,470 @@
...
+    return (
+      await this.allowlistVoting.getPluginInstanceWithSigner().PCT_BASE()
+    ).toNumber();
+  }
+
+  /**
+   * Returns the hash for the SET_CONFIGURATION_PERMISSION permission
+   *
+   * @return {*}  {Promise<string>}
+   * @memberof AllowlistVotingMethods
```
josemarinas: Sae with the rest of the permissions, we must choose either one or the other, we must also take into account that if we switch we will break the UI

modules/client/src/client.ts
brickpop: Right, one thing is `typeof` used in run time to get a value, but it has a completely different use on the TypeScript land. 

Anyway if these types are used only internally, then there's no problem here :+1:

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -0,0 +1,470 @@
...
+      key: Steps.DONE,
+    };
+  }
+
+  /**
+   * Returns the amount of allowed users at a given block
+   *
+   * @param {BigInt} blockNumber
+   * @return {*}  {Promise<BigInt | number>}
+   * @memberof AllowlistVotingMethods
```
josemarinas: We've been mostly using the JS native `BigInt`, I dont have a problem with switching but we must be consistent with it

modules/client/test/integration/client/encoding.test.ts
brickpop: Right, but UnsignedTransactions have many more parameters than the ones we need, isn't it? gasPrice, gasLimit, nonce, chainId, .... We only care about the `{to, value, data}`, and this is what the smart contracts receive and send to the `execute` function. 

The `DaoAction` feels like a global type that everybody will use, whereas `IEncodingResult` is a generic concept that could apply to anything, even a base64 representation of a protobuf payload

modules/client/src/allowlistVoting/interfaces.ts
```diff
@@ -10,23 +8,23 @@ export interface IAllowlistVotingContextPluginState extends IAllowlistVoting {}
...
-  abstain: BigNumberish;
+  startDate: number;
+  endDate: number;
+  snapshotBlock: BigInt;
+  supportRequired: BigInt;
+  participationRequired: BigInt;
+  votingPower: BigInt;
+  yes: BigInt;
+  no: BigInt;
+  abstain: BigInt;
```
brickpop: It might be the exact same, but technically the TypeScript type name for them is `bigint` lowercase. `BigInt(...)`  is the constructor, although it may pass

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -0,0 +1,470 @@
...
+      await this.allowlistVoting
+        .getPluginInstanceWithSigner()
+        .allowedUserCount(BigNumber.from(blockNumber))
+    ).toBigInt();
+  }
+
+  /**
+   * Checks if a proposal can be executed
+   *
+   * @param {BigInt | number} _proposalId
```
josemarinas: Why is `proposalId` a `Bignumber`? 

I guess it is the proposal number for this particular `pluginAddress`, right now we've been calling `proposalId` to the id that subgraph returned `{pluginAddress}_0x{proposal_number}`

modules/client/src/allowlistVoting/internal/decoding.ts
```diff
@@ -105,29 +108,29 @@ export class AllowlistVotingDecoding {
...
    */
   public static execute(
-    txOrData: IDecodingTX | BytesLike
-  ): { _proposalId: string } {
+    txOrData: DecodingTXType | BytesLike
+  ): { _proposalId: number } {
     const data = AllowlistVotingDecoding.getDecodedData("execute", txOrData);
     return {
-      _proposalId: data._voteId,
+      _proposalId: parseInt(data._voteId.toString()),
```
brickpop: should the proposalId be a number? Addresses and ID's are generally represented by a hex string that can be copied and pasted

modules/client/src/allowlistVoting/internal/decoding.ts
```diff
@@ -139,30 +142,40 @@ export class AllowlistVotingDecoding {
...
+    const data = AllowlistVotingDecoding.getDecodedData(
+      "setConfiguration",
+      txOrData
+    );
+    return {
+      _minDuration: parseInt(data._minDuration.toString()),
+      _participationRequiredPct: parseInt(
+        data._participationRequiredPct.toString()
+      ),
+      _supportRequiredPct: parseInt(data._supportRequiredPct.toString()),
```
brickpop: Why do all fields have an underscore at the beginning?

brickpop: @josemarinas should there be a precision decoding here?

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -0,0 +1,470 @@
...
+    return this.allowlistVoting
+      .getPluginInstanceWithSigner()
+      .canExecute(BigNumber.from(_proposalId));
+  }
+
+  /**
+   * Checks if a voter can vote on a proposal
+   *
+   * @param {BigInt | number} _proposalId
+   * @param {string} _voter
```
josemarinas: Same as before for the `proposalId`, also we plan on start using `yup` to validate the inputs coming from the user so I believe this needs to become a params type or Interface, otherwise when we introduce `yup` this will mean grouping this into a yup schema and posibly breaking the ui

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -37,11 +37,13 @@ export class AllowlistVotingMethods {
...
-   * @return {*}  {Promise<BigNumber>}
+   * @return {*}  {Promise<number>}
    * @memberof AllowlistVotingMethods
    */
-  public PCT_BASE(): Promise<BigNumber> {
-    return this.allowlistVoting.getPluginInstanceWithSigner().PCT_BASE();
+  public async PCT_BASE(): Promise<number> {
+    return (
+      await this.allowlistVoting.getPluginInstanceWithSigner().PCT_BASE()
+    ).toNumber();
```
brickpop: Can we use a slightly clearer name? :) 
We are calling the concept "precision" and dealing with "digits"

modules/client/src/allowlistVoting/internal/methods.ts
```diff
@@ -219,87 +231,96 @@ export class AllowlistVotingMethods {
...
    *
-   * @return {*}  {Promise<BigNumber>}
+   * @return {*}  {Promise<number>}
    * @memberof AllowlistVotingMethods
    */
-  public participationRequiredPct(): Promise<BigNumber> {
-    return this.allowlistVoting
-      .getPluginInstanceWithSigner()
-      .participationRequiredPct();
+  public async participationRequiredPct(): Promise<number> {
```
brickpop: Could we name it in terms of the SDK? The contract params are going to be renamed for sure. I suggest `minParticipation` and it should return a number between `0` and `1` as a ratio

brickpop: See also above on `minDuration` and `supportRequiredPct`.... unless there is a layer on top that exports the 0-1 number

modules/client/src/erc20Voting/interfaces.ts
```diff
@@ -10,23 +8,23 @@ export interface IERC20VotingContextPluginState extends IERC20Voting {}
...
-  participationRequired: BigNumberish;
-  votingPower: BigNumberish;
-  yes: BigNumberish;
-  no: BigNumberish;
-  abstain: BigNumberish;
+  startDate: number;
+  endDate: number;
+  snapshotBlock: BigInt;
+  supportRequired: BigInt;
+  participationRequired: BigInt;
```
brickpop: FYI same as before

modules/client/src/erc20Voting/internal/decoding.ts
```diff
@@ -88,47 +91,57 @@ export class ERC20VotingDecoding {
...
+    const data = ERC20VotingDecoding.getDecodedData(
+      "setConfiguration",
+      txOrData
+    );
+    return {
+      _minDuration: parseInt(data._minDuration.toString()),
+      _participationRequiredPct: parseInt(
+        data._participationRequiredPct.toString()
+      ),
+      _supportRequiredPct: parseInt(data._supportRequiredPct.toString()),
```
brickpop, line 127: Related to the comments above

modules/client/src/internal/client/encoding.ts
```diff
@@ -11,13 +12,20 @@ import {
...
 
 /**
  * Encoding module the SDK Generic Client
  */
-export class ClientEncoding {
+export class ClientEncoding extends ClientCore {
+  constructor(context: Context) {
+    super(context);
+    Object.freeze(ClientEncoding.prototype);
+    Object.freeze(this);
```
brickpop: Did we keep this in the end? :)

.github/workflows/javascript_publish.yml
mathewmeconry: yes

modules/client/src/allowlistVoting/client.ts
mathewmeconry: renamed back to `addresslist`

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: yep as long as you use `n` which informs JS to use BigInt. That is why I refactored it to use BigInt

modules/client/src/client-common/interfaces/plugin.ts
mathewmeconry: here we go again into the same discussions if our plugins should live in the SDK. Please tell me yes or no. 
If no I will add all this stuff back and we will have a blast if we ever wanna split it

modules/client/test/integration/client/encoding.test.ts
mathewmeconry: that is why IEncodingResult has the typing of `data`, `value` and `to`. `DaoAction` is the wrong naming because an encoding function doesn't have to return necessarily return a `DaoAction` but more of an unsigned transaction as argued before.  
Tell me what you prefer and I'll do so

modules/client/src/internal/client/encoding.ts
mathewmeconry: as discussed here https://github.com/aragon/sdk/pull/97#discussion_r1018044873 and in Discord we keep the "hardening" in place, that we don't change to much

modules/client/src/allowlistVoting/context.ts
mathewmeconry: done

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: moved them out

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: done

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: you mean only `BigInt` and no `number`?

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: switched to `number` only

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: same

modules/client/src/allowlistVoting/interfaces.ts
mathewmeconry: switched

modules/client/src/allowlistVoting/internal/decoding.ts
mathewmeconry: `proposalId` is a `uint256`. Tell me what you like to have

modules/client/src/allowlistVoting/internal/decoding.ts
mathewmeconry: > Why do all fields have an underscore at the beginning?

was also the case for some parameters in the core client. so I guessed that params pushed as function parameters and not as interface are with an `_` to mark them as contract parameters

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: done

modules/client/src/allowlistVoting/internal/methods.ts
mathewmeconry: why a number between 0 and 1 if I may ask?
the contracts cannot return floats so we would need to do extra "math" on the contracts and/or the sdk just to go into this scale. why is 0-100 bad?

modules/client/src/erc20Voting/interfaces.ts
mathewmeconry: done

modules/client/src/internal/client/encoding.ts
mathewmeconry: same response as before

PR #102, opened 2022-11-15
Feature: Implement `hasPermission`

josemarinas:
## Description

Implement hasPermission interaction with real contracts

Task: [APP-1033](https://aragonassociation.atlassian.net/browse/APP-1033)

## Type of change

## Checklist:


modules/client/src/interfaces.ts
```diff
@@ -129,6 +123,14 @@ export interface IFreezePermissionDecodedParams
   permissionId: string;
 }
 
+export interface IHasPermissionParams {
+  daoAddressOrEns: string;
+  where: string;
+  who: string;
+  permission: string;
```
brickpop: is this the permission name? or its hex hash?
Adding `/** doc comments */` will help here

modules/client/test/integration/client/methods.test.ts
```diff
@@ -266,19 +271,50 @@ describe("Client", () => {
...
+          who: deployment.daoRegistry.address,
+          where: deployment.daoFactory.address,
+          permission: Permissions.REGISTER_DAO_PERMISSION,
+        };
+        const hasPermission = await client.methods.hasPermission(params);
+        // TODO
+        // this permission is granted when the contracts are deployed so it must be true
+        // it should be true?
+        expect(hasPermission).toBe(false);
+      });
```
brickpop: We should add tests for things that are actually true. For example, we should check that in a DAO with an AddressList plugin, `hasPermission()` is true when the plugin wants to execute something on the DAO

josemarinas: Yes, this is what I tried, but I was unable to find a permission that returns true, I tried with what you suggest and it is still returning false
```ts
        const params: IHasPermissionParams = {
          daoAddressOrEns: daoAddress,
          who: deployment.allowListPluginRepo.address,
          where: daoAddress,
          permission: Permissions.EXECUTE_PERMISSION,
        };

```

josemarinas: Never mind, the who and where were reversed

modules/client/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ TEMPLATE:
 - Update Allowance estimation
 - Create dao using real contracts
 - Added `daoRegistryAddress` and `pluginRepoRegistryAddress` to web3 module
+- HasPermission using real contracts
```
mathewmeconry: use `hasPermission`

mathewmeconry (approved):
LGTM

brickpop (approved):

modules/client/CHANGELOG.md
```suggestion
- `hasPermission()` using real contract calls
```

PR #107, opened 2022-11-16
Feat: canVote integration

mathewmeconry:
## Description

canVote uses now real contract call

Task: APP-1218

Depends on: #106 

## Type of change

## Checklist:

### DIFF
--- a/modules/client/package.json
+++ b/modules/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.12.0-alpha",
+  "version": "0.13.0-alpha",
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/sdk-client.esm.js",
--- a/modules/client/package.json
+++ b/modules/client/package.json
@@ -60,7 +60,7 @@
   },
   "dependencies": {
     "@aragon/core-contracts-ethers": "^0.4.1-alpha",
-    "@aragon/sdk-common": "^0.6.0-alpha",
+    "@aragon/sdk-common": "^0.7.0-alpha",
     "@aragon/sdk-ipfs": "^0.2.0-alpha",
     "@ethersproject/abstract-signer": "^5.5.0",
     "@ethersproject/bignumber": "^5.6.0",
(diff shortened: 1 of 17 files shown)


modules/common/CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ TEMPLATE:
 - Fix 1, 2, 3
 -->
 
-## [Upcomfing]
+## [0.6.1-alpha]
```
brickpop: Extremely low importance, but adding a feature should be a minor release, not a patch

modules/common/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-common",
   "author": "Aragon Association",
-  "version": "0.6.0-alpha",
+  "version": "0.6.1-alpha",
```
brickpop:
```suggestion
  "version": "0.7.0-alpha",
```

modules/common/CHANGELOG.md
brickpop: Suggesting `0.7.0-alpha`

modules/client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.12.0-alpha",
+  "version": "0.13.0-alpha",
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/sdk-client.esm.js",
```
brickpop, line 7: We should also be bumping the `client` dependency on `common`'s version

modules/client/test/integration/client-addressList/methods.test.ts
```diff
@@ -184,7 +184,7 @@ describe("Client Address List", () => {
 
       const params: ICanVoteParams = {
         address: "0x1234567890123456789012345678901234567890",
-        proposalId: "0x1234567890123456789012345678901234567890_0x1",
+        proposalId: "0x1234567890123456789012345678901234567890",
```
josemarinas, line 187: I believe this should be a `BigInt` if we are not using the subgraph definition for proposalID

brickpop: mmm here I disagree
Identifiers that can be copied+pasted by the user should be hex strings. Low level binary values should be exchanged in a language-native way, but I don't think this is the case

brickpop (approved):

PR #108, opened 2022-11-18
Fix: Subgraph read calls

josemarinas:
## Description

Fix subgraph read calls with the new structure

Task: [APP-1294](https://aragonassociation.atlassian.net/browse/APP-1294)

## Type of change

## Checklist:


modules/client/src/addressList/interfaces.ts
```diff
@@ -103,10 +103,10 @@ export type SubgraphAddressListProposalListItem = SubgraphProposalBase;
...
   createdAt: string;
   actions: SubgraphAction[];
-  supportRequiredPct: string;
-  participationRequired: string;
+  totalSupportThresholdPct: string;
+  relativeSupportThresholdPct: string;
   voters: SubgraphAddressListVoterListItem[];
-  votingPower: string;
+  census: string;
 };
```
brickpop, line 110: These are Subgraph's types, right? (they are going to change soon)

modules/client/src/addressList/internal/graphql-queries/proposal.ts
```diff
@@ -2,7 +2,7 @@ import { gql } from "graphql-request";
 
 export const QueryAddressListProposal = gql`
 query addressListProposal($proposalId: ID!) {
-  allowlistProposal(id: $proposalId){
+  addresslistProposal(id: $proposalId){
```
brickpop, line 5: Does Subgraph already provide a query with the new name?

modules/client/src/erc20/internal/graphql-queries/proposal.ts
```diff
@@ -32,15 +32,15 @@ query erc20Proposal($proposalId: ID!) {
       vote
       weight
     }
-    pkg{
+    plugin{
 			token{
         symbol
```
brickpop: There seems to be some space/tab formatting issue, but not a big deal right now

modules/client/src/erc20/internal/graphql-queries/proposal.ts
```diff
@@ -61,7 +61,7 @@ query erc20Proposals($where: ERC20VotingProposal_filter!, $limit:Int!, $skip: In
     endDate
     executed
     executable
-    pkg{
+    plugin{
 			token{
         symbol
```
brickpop: Same

modules/client/test/integration/constants.ts
```diff
@@ -46,7 +46,7 @@ const grapqhlEndpoints = {
   working: [
     {
       url:
-        "https://api.thegraph.com/subgraphs/name/aragon/aragon-zaragoza-goerli",
+        "https://api.thegraph.com/subgraphs/name/josemarinas/josemarinas-core-goerli",
```
brickpop: Let's either `env` parameterize this or keep the standard URL

modules/client/src/addressList/interfaces.ts
josemarinas: yes, when they change i'll change it again

modules/client/src/addressList/internal/graphql-queries/proposal.ts
josemarinas: yes :)

modules/client/src/erc20/internal/utils.ts
```diff
@@ -78,31 +78,31 @@ export function toErc20Proposal(
...
       // ),
       // minTurnout: decodeRatio(
-      //   BigInt(proposal.participationRequiredPct),
+      //   BigInt(proposal.relativeSupportThresholdPct),
       //   2,
       // ),
       // TODO DELETE ME
       minSupport: parseFloat(
-        formatEther(proposal.supportRequiredPct),
+        formatEther(proposal.totalSupportThresholdPct),
```
mathewmeconry, line 90: is this `formatEther()` necessary? to imply that a value labeled with `...Pct` isn't actual a percentage but a number of coins with 18 decimals.

modules/client/test/integration/constants.ts
mathewmeconry: remove testing url

josemarinas: Yes, once the subgraph is deployed I'll make a commit updating that

mathewmeconry (changes requested):

modules/client/src/erc20/internal/utils.ts
brickpop: I assume that the current plugin implementation expects the percentage to be expressed with the 18 decimals... of token balances? :-/    `57.00000000000000000000`

Otherwise, it makes no sense IMO

josemarinas: As far as I know subgraph should store a number with 18 decimals, so to transform it into a decimal between 0 and 1, yes, it's necessary

josemarinas: > I assume that the current plugin implementation expects the percentage to be expressed with the 18 decimals... of token balances? :-/ `57.00000000000000000000`
> 
> Otherwise, it makes no sense IMO

Yes when i talked about this with sarkawt

50% =>  5*10e17 (in subgraph) => 0.5 in the UI

brickpop (approved):

PR #109, opened 2022-11-21
feat: changes stored metadata string to ipfs://cid

mathewmeconry:
## Description

Stores CIDs with `ipfs://` prefix in contracts

Task: APP-940

## Type of change

## Checklist:


modules/client/src/internal/client/decoding.ts
```diff
@@ -125,7 +129,7 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
     }
     const result = daoInterface.decodeFunctionData("setMetadata", data);
     const bytes = hexToBytes(result[0]);
-    const cid = new TextDecoder().decode(bytes);
+    const cid = new TextDecoder().decode(bytes).replace("ipfs://", "");
```
brickpop: At least temporarily, but I'd make sure that the value contains `ipfs://<cid>`. Otherwise I would throw an error.

We may add further URI types support, but until then I would validate at least this.

modules/client/src/internal/client/encoding.ts
```diff
@@ -199,7 +199,7 @@ export class ClientEncoding extends ClientCore implements IClientEncoding {
       throw new Error("Could not pin the metadata on IPFS");
     }
     const daoInterface = DAO__factory.createInterface();
-    const args = new TextEncoder().encode(cid);
```
brickpop, line 202: Should we add/modify the existing tests for this? We are going to split the encoding and the pinning, but at least the returned URI format should be checked for regressions

mathewmeconry: the tests succeed. what would you want to modify?

modules/client/src/internal/client/decoding.ts
```diff
@@ -126,12 +130,17 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
     const result = daoInterface.decodeFunctionData("setMetadata", data);
     const bytes = hexToBytes(result[0]);
     const cid = new TextDecoder().decode(bytes);
+
+    if (cid.indexOf("ipfs://") !== 0) {
+      throw new URIError("only ipfs:// is supported");
+    }
+
     const ipfsRegex =
```
brickpop, line 142: Might not be needed if using the URL parser, but otherwise, we don't need to reinitialize this every time. Could be a constant at the top of the file

modules/client/src/internal/client/decoding.ts
```diff
@@ -126,12 +130,17 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
     const result = daoInterface.decodeFunctionData("setMetadata", data);
     const bytes = hexToBytes(result[0]);
     const cid = new TextDecoder().decode(bytes);
```
brickpop: Now this is a URI, not a cid :)  =>    `contenrUri`?

modules/client/src/internal/client/decoding.ts
```diff
@@ -126,12 +130,17 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
...
+
+    if (cid.indexOf("ipfs://") !== 0) {
+      throw new URIError("only ipfs:// is supported");
+    }
```
brickpop: FYI: Might be simpler to check the protocol and get the cid by parsing the value into a `URL` object:

```ts
const url = new URL("ipfs://1234")
// url.protocol
// url.host
```

modules/client/src/internal/client/encoding.ts
brickpop: If the metadata values have changed and tests keep passing without any modification on them, this means that we are not covering this part

modules/client/src/internal/client/decoding.ts
```diff
@@ -152,7 +165,7 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
     }
     const result = daoInterface.decodeFunctionData("setMetadata", data);
     const bytes = hexToBytes(result[0]);
-    const cid = new TextDecoder().decode(bytes);
+    const cid = new TextDecoder().decode(bytes).replace("ipfs://", "");
```
brickpop: Could we do the same with URL here?

modules/client/src/internal/client/decoding.ts
brickpop: mmm ok, so it's still needed. If we keep the regex, let's move it to the top of the file to only initialize it once and not every time.  Or maybe to a constants file on `common`?

modules/client/src/internal/client/decoding.ts
```diff
@@ -152,7 +165,7 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
...
     const ipfsRegex =
```
brickpop, line 174: Same

brickpop (approved):

PR #110, opened 2022-11-22
Feature: Split `createDao` and `createProposal` from `pinMetadata`

josemarinas:
## Description

- Split `createDao()` and `pinMetadata()` 
- Split `createProposal()` and `pinMetadata()` 

Task: [APP-1286](https://aragonassociation.atlassian.net/browse/APP-1286)

## Type of change

## Checklist:

### DIFF
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -80,9 +80,9 @@ const metadata: IMetadata = {
     url: "https://...",
   }],
 };
-const ipfsHash = await client.methods.pinMetadata(metadata);
+const ipfsUri = await client.methods.pinMetadata(metadata);
 const createParams: ICreateParams = {
-  metadataUri: `ipfs://${ipfsHash}`,
+  metadataUri: ipfsUri,
   ensSubdomain: "my-org", // my-org.dao.eth,
   plugins: [],
 };
@@ -630,11 +630,11 @@ const metadata:ProposalMetadata= {
   },
 }
 
-const cid = await client.methods.pinMetadata(metadata)
+const ipfsUri = await client.methods.pinMetadata(metadata);
 
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadataUri: `ipfs://${cid}`,
+  metadataUri: ipfsUri,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -1249,11 +1249,11 @@ const metadata: ProposalMetadata = {
   },
 };
 
-const cid = await client.methods.pinMetadata(metadata);
+const ipfsUri = await client.methods.pinMetadata(metadata);
 
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadataUri: `ipfs://${cid}`,
+  metadataUri: ipfsUri,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -1818,11 +1818,11 @@ const metadataParams: IMetadata = {
 };
 const daoAddressOrEns = "0x12345";
 
-const cid = await client.methods.pinMetadata(metadataParams);
+const ipfsUri = await client.methods.pinMetadata(metadataParams);
 
 const updateMetadataAction = await client.encoding.updateMetadataAction(
   daoAddressOrEns,
-  cid,
+  ipfsUri,
 );
 console.log(updateMetadataAction);
 ```
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -82,7 +82,7 @@ const metadata: IMetadata = {
 };
 const ipfsHash = await client.methods.pinMetadata(metadata);
 const createParams: ICreateParams = {
-  metadata: ipfsHash,
+  metadataUri: ipfsHash,
   ensSubdomain: "my-org", // my-org.dao.eth,
   plugins: [],
 };
@@ -634,7 +634,7 @@ const cid = await client.methods.pinMetadata(metadata)
 
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadata:cid,
+  metadataUri:cid,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -1253,7 +1253,7 @@ const cid = await client.methods.pinMetadata(metadata);
 
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadata: cid,
+  metadataUri: cid,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -2099,9 +2099,7 @@ const context: Context = new Context(contextParams);
 const client: Client = new Client(context);
 const data: Uint8Array = new Uint8Array([12, 56]);
 
-const cid: string = await client.decoding.updateMetadataAction(data);
-
-const params = await client.methods.fetchMetadata(cid);
+const params: string = await client.decoding.updateMetadataAction(data);
 
 console.log(params);
 /*
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -66,20 +66,23 @@ import {
   GasFeeEstimation,
   ICreateParams,
 } from "@aragon/sdk-client";
+import { IMetadata } from "../../src";
 import { contextParams } from "./00-context";
 
 const context: Context = new Context(contextParams);
 const client: Client = new Client(context);
+const metadata: IMetadata = {
+  name: "My DAO",
+  description: "This is a description",
+  avatar: "",
+  links: [{
+    name: "Web site",
+    url: "https://...",
+  }],
+};
+const ipfsHash = await client.methods.pinMetadata(metadata);
 const createParams: ICreateParams = {
-  metadata: {
-    name: "My DAO",
-    description: "This is a description",
-    avatar: "",
-    links: [{
-      name: "Web site",
-      url: "https://...",
-    }],
-  },
+  metadata: ipfsHash,
   ensSubdomain: "my-org", // my-org.dao.eth,
   plugins: [],
 };
@@ -595,6 +598,7 @@ import {
   ContextPlugin,
   ICreateProposalParams,
   ProposalCreationSteps,
+  ProposalMetadata,
   VoteValues,
 } from "@aragon/sdk-client";
 import { contextParams } from "../00-client/00-context";
@@ -606,27 +610,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
 // Create an ERC20 client
 const client = new ClientErc20(contextPlugin);
 
+const metadata:ProposalMetadata= {
+  title: "Test Proposal",
+  summary: "This is a short description",
+  description: "This is a long description",
+  resources: [
+    {
+      name: "Discord",
+      url: "https://discord.com/...",
+    },
+    {
+      name: "Website",
+      url: "https://website...",
+    },
+  ],
+  media: {
+    logo: "https://...",
+    header: "https://...",
+  },
+}
+
+const cid = await client.methods.pinMetadata(metadata)
+
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadata: {
-    title: "Test Proposal",
-    summary: "This is a short description",
-    description: "This is a long description",
-    resources: [
-      {
-        name: "Discord",
-        url: "https://discord.com/...",
-      },
-      {
-        name: "Website",
-        url: "https://website...",
-      },
-    ],
-    media: {
-      logo: "https://...",
-      header: "https://...",
-    },
-  },
+  metadata:cid,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -1209,6 +1217,7 @@ import {
   ContextPlugin,
   ICreateProposalParams,
   ProposalCreationSteps,
+  ProposalMetadata,
   VoteValues,
 } from "@aragon/sdk-client";
 import { contextParams } from "../00-client/00-context";
@@ -1220,27 +1229,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
 // Create an address list client
 const client = new ClientAddressList(contextPlugin);
 
+const metadata: ProposalMetadata = {
+  title: "Test Proposal",
+  summary: "This is a short description",
+  description: "This is a long description",
+  resources: [
+    {
+      name: "Discord",
+      url: "https://discord.com/...",
+    },
+    {
+      name: "Website",
+      url: "https://website...",
+    },
+  ],
+  media: {
+    logo: "https://...",
+    header: "https://...",
+  },
+};
+
+const cid = await client.methods.pinMetadata(metadata);
+
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadata: {
-    title: "Test Proposal",
-    summary: "This is a short description",
-    description: "This is a long descrioption",
-    resources: [
-      {
-        name: "Discord",
-        url: "https://discord.com/...",
-      },
-      {
-        name: "Website",
-        url: "https://website...",
-      },
-    ],
-    media: {
-      logo: "https://...",
-      header: "https://...",
-    },
-  },
+  metadata: cid,
   actions: [],
   startDate: new Date(),
   endDate: new Date(),
@@ -1805,9 +1818,11 @@ const metadataParams: IMetadata = {
 };
 const daoAddressOrEns = "0x12345";
 
+const cid = await client.methods.pinMetadata(metadataParams);
+
 const updateMetadataAction = await client.encoding.updateMetadataAction(
   daoAddressOrEns,
-  metadataParams,
+  cid,
 );
 console.log(updateMetadataAction);
 ```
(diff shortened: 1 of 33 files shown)


modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -42,7 +42,7 @@ export interface IPluginSettings {
 
 export interface ICreateProposalParams {
   pluginAddress: string;
-  metadata: ProposalMetadata;
+  metadata: string;
```
mathewmeconry: metadata doesn't have to be a CID. It could also be an arbitrary bytes string. So using `Uint8Array` would make more sense to follow the general pattern of the SDK

modules/client/src/client-common/utils.ts
```diff
@@ -76,3 +75,10 @@ export function isProposalId(propoosalId: string): boolean {
   const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
   return regex.test(propoosalId);
 }
+
+export function isIpfsCid(hash: string): boolean {
```
mathewmeconry: there are 2 different versions of CID. 
https://docs.ipfs.tech/concepts/content-addressing/#cid-versions

why not using a library that supports everything https://github.com/multiformats/js-multiformats

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -130,6 +122,22 @@ export class ClientAddressListMethods extends ClientCore
...
   }
+
+  /**
+   * Pins a metadata object into IPFS and retruns the generated hash
+   *
+   * @param {ProposalMetadata} params
+   * @return {*}  {Promise<string>}
+   * @memberof ClientMethods
+   */
+  public async pinMetadata(params: ProposalMetadata): Promise<string> {
```
mathewmeconry, line 139: this function doesn't pin the data on the cluster. but rather just adds it. The ipfs cluster will delete this files with the next GC run. 
So I suggest we either actually pin the data in this function or rename it to `addMetadata` for example

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -132,6 +124,22 @@ export class ClientErc20Methods extends ClientCore
...
   }
+
+  /**
+   * Pins a metadata object into IPFS and retruns the generated hash
+   *
+   * @param {ProposalMetadata} params
+   * @return {*}  {Promise<string>}
+   * @memberof ClientMethods
+   */
+  public async pinMetadata(params: ProposalMetadata): Promise<string> {
```
mathewmeconry, line 141: same as above

modules/client/src/addressList/internal/client/methods.ts
mathewmeconry: maybe it is also worth putting it into one of the common folders/packages to remove code duplication. It is the same function everywhere

modules/client/src/internal/client/methods.ts
```diff
@@ -153,6 +151,41 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
     };
   }
+  /**
+   * Pins a metadata object into IPFS and retruns the generated hash
+   *
+   * @param {IMetadata} params
+   * @return {*}  {Promise<string>}
+   * @memberof ClientMethods
+   */
+  public async pinMetadata(params: IMetadata): Promise<string> {
```
mathewmeconry, line 203: same

mathewmeconry (changes requested):

modules/client/examples.md
```diff
@@ -1229,27 +1238,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
...
+    logo: "https://...",
+    header: "https://...",
   },
+};
+
+const cid = await client.methods.pinMetadata(metadata);
+
+const proposalParams: ICreateProposalParams = {
+  pluginAddress: "0x1234567890123456789012345678901234567890",
+  metadata: cid,
```
brickpop: Should `metadata` now be called something like `metadataUri` or `metadataOrigin`? Same elsewhere

modules/client/src/addressList/internal/client/methods.ts
brickpop: Well, then I would add a call to `this.ipfs.pin()` instead of renaming this to `addMetadata()` and letting GC dispose it?

modules/client/src/client-common/interfaces/plugin.ts
brickpop: this relates to the other task for converting this to a general purpose URI. I would rename this into `metadataUri` but then we're probably mixing 2 PR's in one :thinking:

modules/client/src/client-common/utils.ts
brickpop: Preventing bloatware is an asset, if this library is lightweight then +1 here

BTW, this will need to change as soon as the metadata is a URI

modules/client/src/internal/client/decoding.ts
```diff
@@ -152,18 +151,7 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
...
-    if (!ipfsRegex.test(cid)) {
-      throw new Error("The metadata URL defined on the DAO is invalid");
-    }
-    try {
-      const stringMetadata = await this.ipfs.fetchString(cid);
-      return JSON.parse(stringMetadata);
-    } catch {
-      throw new Error("Error reading data from IPFS");
-    }
+    return new TextDecoder().decode(bytes);
```
brickpop: Having some CiD format validation may still be good

modules/client/src/client-common/utils.ts
mathewmeconry: ok then please implement both versions so that the test is at least correct.

josemarinas: It aready suports both cids, there is an or in the regex
 ` /^Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/,`

v0:
` /^Qm([1-9A-HJ-NP-Za-km-z]{44,}$/`

v1:
`/b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/`

modules/client/src/addressList/internal/client/methods.ts
josemarinas: I duplicated it mainly for the return of the function, being `IMetadata` or `ProposalMetadata`. Between the two voting plugin the function is exacly the same so I'll move it to the common. 

About the pin thin I tried to edit the `add` function in `core` to add and then pin the result and rename the function to `addAndPin`, but it kept throwing errors when trying to `pin`, I'm not entirely sure why

modules/client/examples.md
josemarinas: `metadataUri` makes sense to me

modules/client/src/internal/client/decoding.ts
josemarinas: I don't think so, one of the reasons for splitting this is to allow users to use another protocols outside `ipfs` to sore their metadata, if we add a `cid ` validation here we will not allow the user to store the metadata outside `ipfs`

modules/client/src/addressList/internal/client/methods.ts
mathewmeconry: then I suggest you send me the error messages you get so I can fix it for you. Otherwise the naming is wrong

modules/client/src/client-common/utils.ts
mathewmeconry: then we are good. maybe a comment would help, that we support v0 and v1 in case v2 comes out

josemarinas: Sure, I'll add a comment :+1:

modules/client/src/addressList/internal/client/methods.ts
brickpop: I strongly recommend having the `add` and `pin` functions as their own thing... and not `addAndPin`

modules/client/src/client-common/utils.ts
brickpop: given the other PR in progress, I would repurpose this function into something like `isIpfsUri()`, more generic. This will allow detecting potential different formats if we need in the future

modules/client/src/internal/client/decoding.ts
brickpop: See above, by this I mean having some function to detect whether it is an IPFS link that we can `cat()`. In the future, the same building block will remain relevant alongside other potential alternatives

mathewmeconry (approved):

modules/client/examples/00-client/02-create-dao.ts
```diff
@@ -9,20 +9,23 @@ import {
...
-  metadata: {
-    name: "My DAO",
-    description: "This is a description",
-    avatar: "",
-    links: [{
-      name: "Web site",
-      url: "https://...",
-    }],
-  },
+  metadataUri: ipfsHash,
```
brickpop: is this an IPFS hash? or a URI?

modules/client/examples/01-client-erc20/03-create-proposal.ts
```diff
@@ -18,27 +19,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
...
+    logo: "https://...",
+    header: "https://...",
   },
+}
+
+const cid = await client.methods.pinMetadata(metadata)
+
+const proposalParams: ICreateProposalParams = {
+  pluginAddress: "0x1234567890123456789012345678901234567890",
+  metadataUri:cid,
```
brickpop: Same, URI or CiD?

modules/client/src/erc20/internal/client/estimation.ts
```diff
@@ -42,19 +41,11 @@ export class ClientErc20Estimation extends ClientCore
...
-    } catch {
-      throw new IpfsPinError();
-    }
-
     const startTimestamp = params.startDate?.getTime() || 0;
     const endTimestamp = params.endDate?.getTime() || 0;
 
     const estimatedGasFee = await erc20Contract.estimateGas.createVote(
-      toUtf8Bytes(`ipfs://${cid}`),
+      toUtf8Bytes(`ipfs://${params.metadataUri}`),
```
brickpop: Aren't we repeating things here? `metadataUri` should already contain the protocol...?

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -88,19 +88,11 @@ export class ClientErc20Methods extends ClientCore
...
-    } catch {
-      throw new IpfsPinError();
-    }
-
     const startTimestamp = params.startDate?.getTime() || 0;
     const endTimestamp = params.endDate?.getTime() || 0;
 
     const tx = await erc20Contract.createVote(
-      toUtf8Bytes(`ipfs://${cid}`),
+      toUtf8Bytes(`ipfs://${params.metadataUri}`),
```
brickpop: Same, double protocol?

modules/client/src/interfaces.ts
```diff
@@ -14,6 +14,8 @@ import { IClientCore } from "./client-common/interfaces/core";
 /** Defines the shape of the general purpose Client class */
 export interface IClientMethods extends IClientCore {
   create: (params: ICreateParams) => AsyncGenerator<DaoCreationStepValue>;
+  pinMetadata: (params: IMetadata) => Promise<string>;
+  fetchMetadata: (cid: string) => Promise<IMetadata>;
```
brickpop: Isn't `fetchMetadata()` redundant, given that we have `getDao()`?

modules/client/src/interfaces.ts
```diff
@@ -49,7 +51,7 @@ export interface IClientEncoding extends IClientCore {
   ) => Promise<DaoAction>;
   updateMetadataAction: (
     daoAddressOrEns: string,
-    params: IMetadata,
+    params: string,
```
brickpop: Does the string contain the new URI? If so, we should name it accordingly.

Do we have the equivalent on DAO creation?

modules/client/src/interfaces.ts
```diff
@@ -59,7 +61,7 @@ export interface IClientDecoding {
   freezeAction: (data: Uint8Array) => IFreezePermissionDecodedParams;
   withdrawAction: (data: Uint8Array) => IWithdrawParams;
   updateMetadataRawAction: (data: Uint8Array) => string;
-  updateMetadataAction: (data: Uint8Array) => Promise<IMetadata>;
+  updateMetadataAction: (data: Uint8Array) => Promise<string>;
```
brickpop: What does the resulting string contain?

modules/client/src/internal/client/decoding.ts
```diff
@@ -165,23 +165,8 @@ export class ClientDecoding extends ClientCore implements IClientDecoding {
...
-      throw new Error("The metadata URL defined on the DAO is invalid");
-    }
-    try {
-      const stringMetadata = await this.ipfs.fetchString(cid);
-      return JSON.parse(stringMetadata);
-    } catch {
-      throw new Error("Error reading data from IPFS");
-    }
+    const cid = new TextDecoder().decode(bytes);
+    return resolveIpfsCid(cid);
```
brickpop: We were returning the new metadata, whereas we are now returning the CiD, why is this?

modules/client/examples/00-client/02-create-dao.ts
josemarinas: It does not matter, I was planning on supporting both options, if it commes with `ipfs://` we are good, if it comes without it and is a valid cid, we add it, however this is easier to do with `yup` because I can add a validation an transformation to a custom schema called `ipfsCidOrUri`

modules/client/src/internal/client/decoding.ts
josemarinas: My bad I forgot about the `updateMetadataRawAction`

modules/client/examples/00-client/02-create-dao.ts
brickpop: well, the result of  `pinMetadata` should be a URI IMO

modules/client/examples.md
```diff
@@ -1220,27 +1229,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
...
+    logo: "https://...",
+    header: "https://...",
   },
+};
+
+const cid = await client.methods.pinMetadata(metadata);
+
+const proposalParams: ICreateProposalParams = {
+  pluginAddress: "0x1234567890123456789012345678901234567890",
+  metadataUri: cid,
```
brickpop: `cid` vs `metadataUri`?

modules/client/examples/02-client-addressList/03-create-proposal.ts
```diff
@@ -18,27 +19,31 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
...
+    logo: "https://...",
+    header: "https://...",
   },
+};
+
+const cid = await client.methods.pinMetadata(metadata);
+
+const proposalParams: ICreateProposalParams = {
+  pluginAddress: "0x1234567890123456789012345678901234567890",
+  metadataUri: cid,
```
brickpop: cid?

modules/client/examples/03-encoders/05-update-metadata.ts
```diff
@@ -24,8 +24,10 @@ const metadataParams: IMetadata = {
 };
 const daoAddressOrEns = "0x12345";
 
+const cid = await client.methods.pinMetadata(metadataParams);
+
 const updateMetadataAction = await client.encoding.updateMetadataAction(
   daoAddressOrEns,
-  metadataParams,
+  cid,
```
brickpop: cid => metadataUri

modules/client/src/addressList/internal/client/estimation.ts
```diff
@@ -44,18 +43,11 @@ export class ClientAddressListEstimation extends ClientCore
...
-      cid = await this.ipfs.add(JSON.stringify(params.metadata));
-    } catch {
-      throw new IpfsPinError();
-    }
     const startTimestamp = params.startDate?.getTime() || 0;
     const endTimestamp = params.endDate?.getTime() || 0;
 
     const estimatedGasFee = await addresslistContract.estimateGas.createVote(
-      toUtf8Bytes(`ipfs://${cid}`),
+      toUtf8Bytes(`ipfs://${params.metadataUri}`),
```
brickpop: Double protocol?

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -136,6 +128,23 @@ export class ClientAddressListMethods extends ClientCore
...
+    try {
+      const cid = await this.ipfs.add(JSON.stringify(params));
+      await this.ipfs.pin(cid);
+      return cid;
```
brickpop: IMO we should return a URI, not the CiD.... or otherwise refactor the rest of the places where this is used, because there are still mismatches

modules/client/src/erc20/internal/client/estimation.ts
brickpop: ping

modules/client/examples/00-client/02-create-dao.ts
josemarinas: ok then I'll remove all the `ipfs://`

modules/client/src/internal/client/decoding.ts
```diff
@@ -1,8 +1,4 @@
-import {
-  bytesToHex,
-  hexToBytes,
-  UnsupportedProtocolError,
-} from "@aragon/sdk-common";
+import * as sdkCommon from "@aragon/sdk-common";
```
brickpop: Why don't we use selective imports? This allows for dead code elimination

brickpop (approved):

PR #111, opened 2022-11-22
feat: mocks ipfs client in testing

mathewmeconry:
## Description

Mocks IPFS client during testing

Task: APP-1290

## Type of change

## Checklist:


modules/client/test/integration/client-erc20/methods.test.ts
```diff
@@ -349,6 +377,8 @@ describe("Client ERC20", () => {
           expect(typeof proposal.token.address).toBe("string");
           expect(proposal.token.address).toMatch(/^0x[A-Fa-f0-9]{40}$/i);
         }
+
+        mockedIPFSClient.cat.mockImplementation(defaultCatImplementation);
```
josemarinas, line 381: What is this for?

modules/client/test/integration/client-erc20/estimation.test.ts
```diff
@@ -1,6 +1,9 @@
 // @ts-ignore
 declare const describe, it, expect, beforeAll, afterAll;
 
+// mocks need to be at the top of the imports
+import "../../mocks/aragon-sdk-ipfs";
```
josemarinas, line 5: Why is this imported and not used in this file?

modules/client/test/integration/client-addressList/estimation.test.ts
```diff
@@ -1,6 +1,9 @@
 // @ts-ignore
 declare const describe, it, expect, beforeAll, afterAll;
 
+// mocks need to be at the top of the imports
+import "../../mocks/aragon-sdk-ipfs";
```
josemarinas, line 5: same

modules/client/test/integration/client/estimation.test.ts
```diff
@@ -1,6 +1,9 @@
 // @ts-ignore
 declare const describe, it, expect, beforeAll, afterAll;
 
+// mocks need to be at the top of the imports
```
josemarinas, line 4: same

modules/client/test/integration/client-erc20/methods.test.ts
mathewmeconry: to revert back to the overwritten implementation on line 343.
This is necessary because we don't mock the graphql calls and therefore don't know beforehand how many calls to `cat` we are going to make

modules/client/test/integration/client-erc20/estimation.test.ts
mathewmeconry: to trigger the mock. The code gets executed as soon as you import it.
an alternative would be to rewrite it so it exports an init function and call it directly after the import.
Similar to what you need to do with the `dotenv` package

modules/client/test/integration/client-addressList/estimation.test.ts
mathewmeconry: same

modules/client/test/integration/client/estimation.test.ts
mathewmeconry: same

modules/client/test/integration/client-erc20/methods.test.ts
brickpop: Would `mockResolvedValueOnce` do the trick instead?

modules/client/test/mocks/aragon-sdk-ipfs.ts
```diff
@@ -0,0 +1,19 @@
+export const mockedIPFSClient = {
+  nodeInfo: jest.fn().mockResolvedValue(true),
+  add: jest.fn().mockResolvedValue({
+    hash: "QmXhJawTJ3PkoKMyF3a4D89zybAHjpcGivkb7F1NkHAjpo",
+  }),
+  cat: jest.fn().mockImplementation(async () => Buffer.from("{}")),
+  __proto__: {},
```
brickpop, line 7: We seem to be mocking `add` only once, it feels like there should be many more uses in the testing?  Should `pin` be mocked as well?

modules/client/test/integration/client-erc20/methods.test.ts
mathewmeconry: nop as the name suggests it is `once` and graphql can return more than one result

modules/client/test/mocks/aragon-sdk-ipfs.ts
mathewmeconry: what do you mean with once? this mock is the general mock. the other instances is overriding the default mock. Especially needed for `cat`
Yep `pin` would make sense, once we call it somewhere

brickpop: Oh, so that single mocking of `cat` affects everywhere? I thought the mock somehow was file-scoped :thinking: 

That may bring unexpected behavior in unrelated files where this line of code is not visible. Do we have a way to "enable" and "disable" these mocks, independently on each file?

I would expect having that at the `before()` and `after()` blocks of each test suite... otherwise, the tests where the `cat` mocking is done are polluting the rest.

Same if this applies to `cat`

brickpop (approved):

josemarinas (approved):

PR #116, opened 2022-11-29
Feature: Add schema validation for user inputs

josemarinas:
## Description

Add schema validation with `yup` for user inputs

Task: [APP-946](https://aragonassociation.atlassian.net/browse/APP-946)

## Type of change

## Checklist:


modules/client/src/addressList/internal/client/decoding.ts
```diff
@@ -30,6 +31,7 @@ export class ClientAddressListDecoding extends ClientCore
    * @memberof ClientAddressListDecoding
    */
   public updatePluginSettingsAction(data: Uint8Array): IPluginSettings {
+    uint8ArraySchema.validateSync(data);
```
brickpop: Does this throw an error? Should we control it ourselves or is it better explained by yup?

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -64,6 +64,8 @@ export interface IExecuteProposalParams {
 export interface ICanVoteParams {
   pluginAddress: string;
   proposalId: string;
+  // TODO
+  // addressOrEns ??
```
brickpop: Why TO DO? why this?

modules/client/src/client-common/schemas/common.ts
```diff
@@ -0,0 +1,79 @@
...
+  new InvalidIpfsUriError().message,
+  (value) => value ? isIpfsUri(value) : true,
+);
+
+// address list or ens
+export const addressOrEnsSchema = string().test(
+  "isAddressOrEns",
+  new InvalidAddressError().message,
+  // check ens
+  (value) => value ? isAddress(value) : true,
```
brickpop: Does this also check the ENS? I think it doesn't

modules/client/src/client-common/schemas/common.ts
```diff
@@ -0,0 +1,79 @@
...
+  "isUint8Array",
+  new InvalidDataError().message,
+  (value) => value ? value instanceof Uint8Array : true,
+);
+
+// uint8array
+export const bigNumberishSchema = mixed().test(
+  "isBigNumberish",
+  new InvalidBigNumberishError().message,
+  (value) => value ? isBigNumberish(value) : true,
```
brickpop: I don't understand the testing logic here

modules/client/src/client-common/schemas/common.ts
```diff
@@ -0,0 +1,79 @@
...
+);
+
+// uint8array
+export const uint8ArraySchema = mixed().test(
+  "isUint8Array",
+  new InvalidDataError().message,
+  (value) => value ? value instanceof Uint8Array : true,
+);
+
+// uint8array
```
brickpop: // number, big number, bigint, etc.

brickpop: BTW, I think we should be always accepting `bigint` values, never BigNumber's

brickpop (commented):
Round 1

modules/client/src/addressList/internal/client/decoding.ts
josemarinas: This already throws an error depending on what fails in the validation, on the custom validations we specify the errors when we define the custom schema

modules/client/src/client-common/interfaces/plugin.ts
josemarinas: Yes, I left that there so I don't forget about it, this address can be an ens, from what I understand the contracts resolve ens addresses so all addresses should be `addressOrEns`, is this correct?

If it is correct should we validate this the ens address by checking if it has the correct format `something.whatever.eth` or should we check that the address resolves on the validation?

modules/client/src/client-common/schemas/common.ts
josemarinas: Check other comment

modules/client/src/client-common/schemas/common.ts
josemarinas: I added this as a temporary measure until we decide the new `proposalId` format

the logic with the `value? something : true` is that we don't always know if the value is defined, so if it is defined we do the validations, if it is not, we don't do them because if we do it we can't have optional values.

If a required field receives an `undefined`, will pass this, but it won't pass the `required()` flag

modules/client/src/client-common/schemas/common.ts
josemarinas: Yes I briefly commented on this on my other comment, until we don't have a final `proposalId` format I used `BigNumber` because is what's on the contracts, so you can receive `5` `"5"` or `"0x5"` to refer to a `proposalId`. Once we decide and deploy the new format, this will be changed accordingly

modules/client/src/addressList/schemas.ts
```diff
@@ -0,0 +1,9 @@
+import { array, object } from "yup";
+import { addressOrEnsSchema, pluginSettingsSchema } from "../client-common";
+
+export const membersSchema = array(addressOrEnsSchema);
+
+export const addressListPluginInstallSchema = object({
```
brickpop: Don't we need to implement any other addressList specific schema here?

modules/client/src/client-common/interfaces/plugin.ts
brickpop: 1) If this is something we should do now, let's do it. If it is a task for later, let's have the task created. 

2) Yes, if we accept ENS domains, we definitely need to validate them, otherwise checks will fail

modules/client/src/client-common/schemas/common.ts
brickpop: - We should definitely support/test ENS domains
- if `value` is falsy, then why do we return `true`?

modules/client/src/client-common/schemas/common.ts
brickpop: In general, the SDK should only have one correct way of receiving and returning values, and for big numbers this should be `bigint` and not `BigNumber`.

If we want to accept hex strings, then let's go all in and validate it

modules/client/src/client-common/schemas/common.ts
brickpop: I see
If the value is empty, then let's name the schema something like `optionalXXX` instead of leaving it the way it is

modules/client/src/client-common/schemas/common.ts
```diff
@@ -0,0 +1,79 @@
...
+  "isBigNumberish",
+  new InvalidBigNumberishError().message,
+  (value) => value ? isBigNumberish(value) : true,
+);
+
+// date
+export const dateSchema = mixed().test(
+  "isDate",
+  new InvalidDateError().message,
+  (value) => value ? value instanceof Date : true,
```
brickpop: Same as before... if we are expecting a date but we get  `undefined` why are we accepting the intput?

modules/client/src/client-common/schemas/plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+export const createProposalParamsSchema = object({
+  pluginAddress: addressOrEnsSchema.required(),
+  metadataUri: ipfsUriSchema.required(),
+  actions: array(daoActionSchema).default([]),
+  startDate: dateSchema,
+  endDate: dateSchema,
+  executeOnPass: boolean(),
+  creatorVote: voteValuesSchema,
```
brickpop, line 57: @Michael-A-Heuer is this field still relevant? I think it directly collides with the fact that votes with a start date in the future cannot have any vote from the creator. 
At the same time, the creator could have no tokens and still vote

modules/client/src/client-common/schemas/plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+export const voteProposalParamsSchema = object({
+  pluginAddress: addressOrEnsSchema.required(),
+  vote: voteValuesSchema.required(),
+  proposalId: bigNumberishSchema.required(),
```
brickpop: Why don't se simply use a `hexStringSchema` here?

modules/client/src/client-common/schemas/plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+  // TODO
+  // addressOrEns ??
+  address: addressSchema.required(),
+});
+
+export const pluginSettingsSchema = object({
+  minSupport: number().lessThan(1).positive().required(),
+  minTurnout: number().lessThan(1).positive().required(),
+  minDuration: number().positive().required(),
+});
```
brickpop: This is not common to all plugins. This applies to AddressList as of today, but definitely not to ERC20 (now TokenVoting).

modules/client/src/client-common/utils.ts
```diff
@@ -71,7 +71,13 @@ export function computeProposalStatusFilter(
...
 }
 
-export function isProposalId(propoosalId: string): boolean {
+export function isProposalId(proposalId: string): boolean {
   const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
-  return regex.test(propoosalId);
+  return regex.test(proposalId);
+}
+export function isIpfsUri(ipfsUri: string): boolean {
+  const regex = new RegExp(
```
brickpop, line 87: IPFS URI's can theoretically have a path after the CID. We are not supporting it at this point, but `isIpfsUri` sounds bad if it doesn't support paths

modules/client/src/erc20/schemas.ts
```diff
@@ -0,0 +1,26 @@
...
+  })).required(),
+});
+export const existingTokenParamsSchema = object({
+  address: addressOrEnsSchema.required(),
+});
+
+export const erc20PluginInstallSchema = object({
+  settings: pluginSettingsSchema.required(),
+  newToken: newTokenParamsSchema.default(undefined),
+  useToken: existingTokenParamsSchema.default(undefined),
```
brickpop: Is `settings` the common settings and `newtoken` and `useToken` the custom bit?

I suggest we have a 100% custom settings object. SDK users should be building a single settings object, regardless of the code that we can reuse internally. 

This may also include the `earlyExecution` flag and the `voteReplacement` one as well

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+  addressSchema,
+  bigintSchema,
+  ipfsUriSchema,
+  permissionSchema,
+  pluginInstallItemSchema,
+  uint8ArraySchema,
+} from "./client-common";
+
+// DAO details
+const daoResourceLinkSchema = object({
```
brickpop, line 13: What's the criteria to put some schemas here and some on the common folder?

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+  name: string().required(),
+  url: string().required(),
+});
+
+export const metadataSchema = object({
```
brickpop: which metadata? DAO or proposal?

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+  name: string().required(),
+  description: string().required(),
+  avatar: string().optional(),
+  links: array(daoResourceLinkSchema).required(),
+});
+// Create proposal
+export const createParamsSchema = object({
```
brickpop, line 25: create what? a DAO or a proposal?

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+  tokenAddress: addressSchema,
+  reference: string(),
+});
+
+// has permission
+export const hasPermissionParamsSchema = object({
+  daoAddressOrEns: addressOrEnsSchema.required(),
+  where: addressOrEnsSchema.required(),
+  who: addressOrEnsSchema.required(),
+  permission: permissionSchema.required(),
```
brickpop, line 53: I'm not sure if we call it `permissions ` everywhere here. In general we speak of permission ID's when computed in Solidity, but the variable names are just permission. 
Feel free to resolve

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+  data: uint8ArraySchema,
+});
+
+export const permissionParamsSchema = object({
+  where: addressOrEnsSchema.required(),
+  who: addressOrEnsSchema.required(),
+  permission: permissionSchema.required(),
+});
```
brickpop, line 61: Permissions can also have an ~~oracle~~ (condition, filter)
Coming up, not yet

modules/client/src/schemas.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const freezePermissionParamsSchema = object({
+  where: addressOrEnsSchema.required(),
+  permission: permissionSchema.required(),
```
brickpop, line 65: Doesn't freezing permissions also need the `who`?

modules/client/test/integration/client-addressList/estimation.test.ts
```diff
@@ -62,7 +62,7 @@ describe("Client Address List", () => {
 
       const voteParams: IVoteProposalParams = {
         pluginAddress: "0x1234567890123456789012345678901234567890",
-        proposalId: "0x1234567890123456789012345678901234567890",
+        proposalId: "0x0",
```
brickpop: Why 0x0?

modules/client/test/integration/client-addressList/estimation.test.ts
```diff
@@ -82,7 +82,7 @@ describe("Client Address List", () => {
 
       const executeParams: IExecuteProposalParams = {
         pluginAddress: "0x1234567890123456789012345678901234567890",
-        proposalId: "0x1234567890123456789012345678901234567890",
+        proposalId: "0x0",
```
brickpop: Why 0x0?

modules/common/src/errors.ts
```diff
@@ -95,3 +95,43 @@ export class IpfsFetchError extends Error {
...
+}
+
+export class InvalidDateError extends Error {
+  constructor() {
+    super("Invalid type, Date was expected");
+  }
+}
+export class InvalidBigNumberishError extends Error {
+  constructor() {
+    super("Invalid type, BigNumberish was expected");
```
brickpop: `BigNumberish` the ethers type?
I don't think we should become ethers-dependent like this

modules/client/src/client-common/schemas/common.ts
josemarinas: I'm not sure about changing the name, optionality is assumed in all the `yup` schemas unless you specify otherwise.

modules/client/src/addressList/schemas.ts
josemarinas: Nope, most of them are common with the erc20 package, only the installation ones are unique

modules/client/src/client-common/schemas/plugin.ts
josemarinas: Why not?
`minSupport` and `minTurnout` are both an integer between 0 and 1 in both cases

modules/client/src/schemas.ts
josemarinas: Nope, the freeze function is one in the smart contracts
```sol
function freeze(address _where, bytes32 _permissionId)
        external
        auth(_where, ROOT_PERMISSION_ID)
    {
        _freeze(_where, _permissionId);
    }

```

modules/client/test/integration/client-addressList/estimation.test.ts
josemarinas: this should be a `BigNumberish`, we can lock it to a `bigint` and solve the other issues with the `BigNumber` validator

modules/client/src/client-common/utils.ts
josemarinas: Ok, i'll edit the regex to accept paths :+1:

modules/client/src/erc20/schemas.ts
josemarinas: Yes, this is exactly how we have the types and interfaces, what do you suggest for changing it?

modules/client/src/schemas.ts
josemarinas: Here used only by this plugin 

`client-common/schemas/common` => used by all clients
`client-common/schemas/plugin` => used only by plugin clients

It follows the same rules as the interfaces folder.

modules/client/src/schemas.ts
josemarinas: Dao, it follows the types and interfaces naming, for some reason the DAO metadata is `Metadata` but the proposal metadata is `ProposalMetadata`

modules/client/src/schemas.ts
josemarinas: Dao, same as before, in the dao methods, interfaces, etch we dident add the `Dao` at the end, in the Proposal ones we did. For example:

Creating a dao
`client.methods.create`

Creating a proposal
`client.methods.createProposal`

We can fix these inconsistencies across the code, but I think it's too big to be addressed in this PR

modules/common/src/errors.ts
josemarinas: We can change this to a big int and close some conversations regarding this

modules/client/src/schemas.ts
brickpop: Can we replace `Metadata` with `DaoMetadata`?

modules/client/src/schemas.ts
brickpop: Validate string regex `A-Z_`  

Arbitrary permissions `MINT_PERMISSION`

modules/client/examples.md
```diff
@@ -382,7 +382,7 @@ console.log(transfers);
     reference: "withdrawing from dao to:0xc8541aAE19C5069482239735AD64FAC3dCc52Ca2",
     transactionId: "0xdb0f9422b5c3199021481c98a655741ca16119ff8a59571854a94a6f31dad7ba",
     to: "0xc8541aae19c5069482239735ad64fac3dcc52ca2",
-    proposalId: "0x1234567890123456789012345678901234567890_0x0"
+    proposalId: 0n
```
brickpop, line 385: This will be breaking the app team if we publish it now. We also need internal alignment on how to represent the proposal ID's :warning:

modules/client/src/client-common/schemas/plugin.ts
brickpop: But these settings are not applicable to the admin plugin, multisig, etc. The same problem with the naming exists in the contracts. These are voting settings, not general plugin settings

modules/client/src/erc20/schemas.ts
brickpop: As I said before (and also applies to the contracts) we should have the (1) voting settings that are common in some plugins and (2) the specific settings object for each plugin. 

(2) will always vary depending on the plugin. The Admin plugin will never share common settings with the TokenVoting one

modules/client/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ TEMPLATE:
 -->
 
 ## [UPCOMING]
+### Added
+- Input vaidation for onchain methods
```
brickpop: I think we should also add the change to `createDao()` and related

modules/client/examples.md
brickpop: I think this is blocked until we get internal alignment

modules/client/src/addressList/schemas.ts
```diff
@@ -0,0 +1,9 @@
...
+  settings: pluginSettingsSchema.required(),
```
brickpop: Same as above... this schema should be scoped to voting, not to "pluginSettings"

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -48,23 +47,25 @@ export interface ICreateProposalParams {
...
-  proposalId: string;
+  proposalId: bigint;
 }
 
 export interface ICanVoteParams {
   pluginAddress: string;
-  proposalId: string;
-  address: string;
+  proposalId: bigint;
+  addressOrEns: string;
```
brickpop: As discussed before, this should be a single hex string value, everywhere

modules/client/src/client-common/utils.ts
```diff
@@ -71,7 +71,26 @@ export function computeProposalStatusFilter(
   return where;
 }
 
-export function isProposalId(propoosalId: string): boolean {
+export function isProposalId(proposalId: string): boolean {
   const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
```
brickpop, line 83: According to what we discussed, this should be 64 hex chars long (plus the 0x I assume)

brickpop (commented):
Next round

modules/client/src/client-common/interfaces/plugin.ts
josemarinas: Yes, the problem is that if we specify the new format `0x...{64}` we won’t be passing tests because neither the contracts nor the subgraph is updated

modules/client/src/client-common/utils.ts
josemarinas: See above, I woud't change this for now

modules/client/src/client-common/schemas/plugin.ts
josemarinas: Changed to `IVotingSettings`

modules/client/src/schemas.ts
josemarinas: Added validation function `isPermission` that solves this

modules/client/src/client-common/interfaces/plugin.ts
brickpop: Well, then we're going to be stuck forever... just becacuse one old subgraph version had a different type :shrug: 

We could have a loose regex validation, but the type should definitely be a string with hex characters

Rekard0:
Not relevant anymore

PR #118, opened 2022-12-06
Fixes metadata URI parsing

brickpop:
## Description

Task:APP-1351

## Type of change

FIX

## Checklist:

### DIFF
--- a/modules/client/src/addressList/internal/client/methods.ts
+++ b/modules/client/src/addressList/internal/client/methods.ts
@@ -276,7 +276,9 @@ export class ClientAddressListMethods extends ClientCore
         return null;
       }
       // format in the metadata field
-      const metadataString = await this.ipfs.fetchString(addressListProposal.metadata);
+      const metadataUri = new URL(addressListProposal.metadata);
+      const metadataCid = metadataUri.host;
+      const metadataString = await this.ipfs.fetchString(metadataCid);
       // TODO: Parse and validate schema
       const metadata = JSON.parse(metadataString) as ProposalMetadata;
       return toAddressListProposal(addressListProposal, metadata);
@@ -348,8 +350,10 @@ export class ClientAddressListMethods extends ClientCore
             proposal: SubgraphAddressListProposalListItem,
           ): Promise<AddressListProposalListItem> => {
             // format in the metadata field
+            const metadataUri = new URL(proposal.metadata);
+            const metadataCid = metadataUri.host;
             return this.ipfs
-              .fetchString(proposal.metadata)
+              .fetchString(metadataCid)
               .then((stringMetadata: string) => {
                 // TODO: Parse and validate schema
                 const metadata = JSON.parse(stringMetadata) as ProposalMetadata;
--- a/modules/client/src/addressList/internal/client/methods.ts
+++ b/modules/client/src/addressList/internal/client/methods.ts
@@ -356,17 +356,17 @@ export class ClientAddressListMethods extends ClientCore
         addressListProposals.map(
           (
             proposal: SubgraphAddressListProposalListItem,
-          ):
-            | Promise<AddressListProposalListItem>
-            | AddressListProposalListItem => {
+          ): Promise<AddressListProposalListItem> => {
             let metadataCid = "";
             try {
               // format in the metadata field
               metadataCid = resolveIpfsCid(proposal.metadata);
             } catch (err) {
-              return toAddressListProposalListItem(
-                proposal,
-                UNSUPPORTED_PROPOSAL_METADATA_LINK,
+              return Promise.resolve(
+                toAddressListProposalListItem(
+                  proposal,
+                  UNSUPPORTED_PROPOSAL_METADATA_LINK,
+                ),
               );
             }
             return this.ipfs
--- a/modules/client/src/addressList/internal/client/methods.ts
+++ b/modules/client/src/addressList/internal/client/methods.ts
@@ -49,7 +49,7 @@ import { id } from "@ethersproject/hash";
 import { hexZeroPad } from "@ethersproject/bytes";
 import { toUtf8Bytes } from "@ethersproject/strings";
 import {
-  UNAVAILABLE_PROPOSAL_METADATA_LINK,
+  UNAVAILABLE_PROPOSAL_METADATA,
   UNSUPPORTED_PROPOSAL_METADATA_LINK,
 } from "../../../client-common/constants";
 
@@ -296,7 +296,7 @@ export class ClientAddressListMethods extends ClientCore
         }
         return toAddressListProposal(
           addressListProposal,
-          UNAVAILABLE_PROPOSAL_METADATA_LINK,
+          UNAVAILABLE_PROPOSAL_METADATA,
         );
       }
     } catch (err) {
@@ -374,17 +374,15 @@ export class ClientAddressListMethods extends ClientCore
               return toAddressListProposalListItem(proposal, metadata);
             } catch (err) {
               if (err instanceof InvalidCidError) {
-                return Promise.resolve(
-                  toAddressListProposalListItem(
-                    proposal,
-                    UNSUPPORTED_PROPOSAL_METADATA_LINK,
-                  ),
+                return toAddressListProposalListItem(
+                  proposal,
+                  UNSUPPORTED_PROPOSAL_METADATA_LINK,
                 );
               }
-              return Promise.resolve(toAddressListProposalListItem(
+              return toAddressListProposalListItem(
                 proposal,
-                UNAVAILABLE_PROPOSAL_METADATA_LINK,
-              ));
+                UNAVAILABLE_PROPOSAL_METADATA,
+              );
             }
           },
         ),
--- a/modules/client/src/addressList/internal/client/methods.ts
+++ b/modules/client/src/addressList/internal/client/methods.ts
@@ -47,7 +47,7 @@ import { AllowlistVoting__factory } from "@aragon/core-contracts-ethers";
 import { id } from "@ethersproject/hash";
 import { hexZeroPad } from "@ethersproject/bytes";
 import { toUtf8Bytes } from "@ethersproject/strings";
-import { UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA } from "../../../client-common/constants";
+import { UNSUPPORTED_PROPOSAL_METADATA_LINK } from "../../../client-common/constants";
 
 /**
  * Methods module the SDK Address List Client
@@ -277,16 +277,18 @@ export class ClientAddressListMethods extends ClientCore
       if (!addressListProposal) {
         return null;
       }
-      // format in the metadata field
-      let metadata: ProposalMetadata = UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA;
+      let metadataCid = "";
       try {
-        const metadataCid = resolveIpfsCid(addressListProposal.metadata);
-        const metadataString = await this.ipfs.fetchString(metadataCid);
+        metadataCid = resolveIpfsCid(addressListProposal.metadata);
         // TODO: Parse and validate schema
-        metadata = JSON.parse(metadataString) as ProposalMetadata;
       } catch (err) {
-        console.warn(err);
+        return toAddressListProposal(
+          addressListProposal,
+          UNSUPPORTED_PROPOSAL_METADATA_LINK,
+        );
       }
+      const metadataString = await this.ipfs.fetchString(metadataCid);
+      const metadata = JSON.parse(metadataString) as ProposalMetadata;
       return toAddressListProposal(addressListProposal, metadata);
     } catch (err) {
       throw new GraphQLError("AddressList proposal");
@@ -357,25 +359,25 @@ export class ClientAddressListMethods extends ClientCore
           ):
             | Promise<AddressListProposalListItem>
             | AddressListProposalListItem => {
+            let metadataCid = "";
             try {
               // format in the metadata field
-              const metadataCid = resolveIpfsCid(proposal.metadata);
-              return this.ipfs
-                .fetchString(metadataCid)
-                .then((stringMetadata: string) => {
-                  // TODO: Parse and validate schema
-                  const metadata = JSON.parse(
-                    stringMetadata,
-                  ) as ProposalMetadata;
-                  return toAddressListProposalListItem(proposal, metadata);
-                });
+              metadataCid = resolveIpfsCid(proposal.metadata);
             } catch (err) {
-              console.warn(err);
               return toAddressListProposalListItem(
                 proposal,
-                UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA,
+                UNSUPPORTED_PROPOSAL_METADATA_LINK,
               );
             }
+            return this.ipfs
+              .fetchString(metadataCid)
+              .then((stringMetadata: string) => {
+                // TODO: Parse and validate schema
+                const metadata = JSON.parse(
+                  stringMetadata,
+                ) as ProposalMetadata;
+                return toAddressListProposalListItem(proposal, metadata);
+              });
           },
         ),
       );
--- a/modules/client/src/addressList/internal/client/methods.ts
+++ b/modules/client/src/addressList/internal/client/methods.ts
@@ -7,6 +7,7 @@ import {
   NoProviderError,
   ProposalCreationError,
   Random,
+  resolveIpfsCid,
 } from "@aragon/sdk-common";
 import { isAddress } from "@ethersproject/address";
 import {
@@ -276,8 +277,7 @@ export class ClientAddressListMethods extends ClientCore
         return null;
       }
       // format in the metadata field
-      const metadataUri = new URL(addressListProposal.metadata);
-      const metadataCid = metadataUri.host;
+      const metadataCid = resolveIpfsCid(addressListProposal.metadata);
       const metadataString = await this.ipfs.fetchString(metadataCid);
       // TODO: Parse and validate schema
       const metadata = JSON.parse(metadataString) as ProposalMetadata;
@@ -350,8 +350,7 @@ export class ClientAddressListMethods extends ClientCore
             proposal: SubgraphAddressListProposalListItem,
           ): Promise<AddressListProposalListItem> => {
             // format in the metadata field
-            const metadataUri = new URL(proposal.metadata);
-            const metadataCid = metadataUri.host;
+            const metadataCid = resolveIpfsCid(proposal.metadata);
             return this.ipfs
               .fetchString(metadataCid)
               .then((stringMetadata: string) => {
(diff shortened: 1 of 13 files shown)


modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -276,7 +276,9 @@ export class ClientAddressListMethods extends ClientCore
         return null;
       }
       // format in the metadata field
-      const metadataString = await this.ipfs.fetchString(addressListProposal.metadata);
+      const metadataUri = new URL(addressListProposal.metadata);
```
josemarinas: The problem with this is that it's going to crash when the metadata does not have the correct format and throw a `GraphQLError`. For example, the metadata with the cid without the `ipfs://` will throw an error. I believe we should make a helper to get the metadata that checks this stuff and returns the metadata depending on the protocol, and if the protocol is not supported by the sdk return the same content that we have on the subgraph or null.

This is a first idea, let me know what do you think

mathewmeconry: also instead of using the URL which assumes that a protocol is present you can use the IPFS cid regex and check if there is a CID and use the match to get the data from ipfs

```typescript
      const ipfsRegex =
        /(Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,}))/;
      const matches = addressListProposal.metadata.match(ipfsRegex);

      if (matches && matches.length > 0) {
        const metadataString = await this.ipfs.fetchString(
          matches[0],
        );
        // TODO: Parse and validate schema
        const metadata = JSON.parse(metadataString) as ProposalMetadata;
        return toAddressListProposal(addressListProposal, metadata);
      }
      return toAddressListProposal(addressListProposal, {
        description: "",
        resources: [],
        summary: "",
        title: "",
      });
```

brickpop: Added it

modules/common/src/uri.ts
```diff
@@ -0,0 +1,20 @@
+const IPFS_CID_REGEX =
+  /^Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/;
+
+/** Attempts to parse the given string as a URL and returns the IPFS CiD contained in it.
+ * Alternatively it tries to use the raw value after validating it.
+ */
+export function resolveIpfsCid(data: string): string {
+  if (!data.includes(":")) {
```
mathewmeconry: check at least `://` or even better check with `^\w+://`
a `:` can happen anywhere. One example would be basic authentication in the URL

mathewmeconry (changes requested):

modules/common/src/uri.ts
```diff
@@ -0,0 +1,21 @@
...
+  /^Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/;
+
+const PROTOCOL_REGEX = /^\w+:\/\//;
+/** Attempts to parse the given string as a URL and returns the IPFS CiD contained in it.
+ * Alternatively it tries to use the raw value after validating it.
+ */
+export function resolveIpfsCid(data: string): string {
+  if (!PROTOCOL_REGEX.test(data)) {
+    if (!IPFS_CID_REGEX.test(data)) {
+      throw new Error("The value does not contain a valid CiD");
```
mathewmeconry: please move the error into the common lib where we have all the other errors

modules/common/src/uri.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+    return data;
+  }
+
+  const metadataUri = new URL(data);
+  if (!IPFS_CID_REGEX.test(metadataUri.host)) {
+    throw new Error("The value does not contain a valid CiD");
```
mathewmeconry: please move the error into the common lib where we have all the other errors

mathewmeconry (changes requested):

mathewmeconry (approved):

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -276,9 +278,15 @@ export class ClientAddressListMethods extends ClientCore
...
-      // TODO: Parse and validate schema
-      const metadata = JSON.parse(metadataString) as ProposalMetadata;
+      let metadata: ProposalMetadata = UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA;
+      try {
+        const metadataCid = resolveIpfsCid(addressListProposal.metadata);
+        const metadataString = await this.ipfs.fetchString(metadataCid);
+        // TODO: Parse and validate schema
+        metadata = JSON.parse(metadataString) as ProposalMetadata;
+      } catch (err) {
+        console.warn(err);
```
brickpop: Why printing things?

modules/client/src/client-common/constants.ts
```diff
@@ -0,0 +1,8 @@
+import { ProposalMetadata } from "./interfaces/plugin";
+
+export const UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA: ProposalMetadata = {
+  title: "Unsupported protocol",
+  summary: "Unsupported protocol",
+  description: "Unsupported protocol",
```
brickpop:
```suggestion
  title: "Unsupported metadata link",
  summary: "Unsupported metadata link",
  description: "Unsupported metadata link",
```

modules/client/src/client-common/constants.ts
```diff
@@ -0,0 +1,8 @@
+import { ProposalMetadata } from "./interfaces/plugin";
+
+export const UNSUPPORTED_PROTOCOL_PROPOSAL_METADATA: ProposalMetadata = {
```
brickpop:
```suggestion
export const UNSUPPORTED_METADATA_LINK: ProposalMetadata = {
```

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -339,15 +347,26 @@ export class ClientErc20Methods extends ClientCore
...
+                .fetchString(metadataCid)
+                .then((stringMetadata: string) => {
+                  // TODO: Parse and validate schema¡
+                  const metadata = JSON.parse(
+                    stringMetadata,
+                  ) as ProposalMetadata;
+                  return toErc20ProposalListItem(proposal, metadata);
+                });
+            } catch (err) {
+              console.warn(err)
```
brickpop: Same

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -339,15 +347,26 @@ export class ClientErc20Methods extends ClientCore
...
+              return this.ipfs
+                .fetchString(metadataCid)
+                .then((stringMetadata: string) => {
+                  // TODO: Parse and validate schema¡
+                  const metadata = JSON.parse(
+                    stringMetadata,
+                  ) as ProposalMetadata;
+                  return toErc20ProposalListItem(proposal, metadata);
+                });
+            } catch (err) {
```
brickpop, line 369: Shouldn't this try/catch be inside of the `.then()` clause? 

This try/catch block will only trigger in case the sync code throws an Error, but internal promise-related stuff would be out of scope here

modules/client/src/internal/client/methods.ts
```diff
@@ -336,10 +335,18 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      // const stringMetadata = await this.ipfs.fetchString(metadataCid);
+      // const metadata = JSON.parse(stringMetadata);
+      let metadata: IMetadata = UNSUPPORTED_PROTOCOL_DAO_METADATA;
+      try {
+        const metadataCid = resolveIpfsCid(dao.metadata);
+        const metadataString = await this.ipfs.fetchString(metadataCid);
+        // TODO: Parse and validate schema
+        metadata = JSON.parse(metadataString) as IMetadata;
+      } catch (err) {
+        console.warn(err);
```
brickpop: Same

modules/client/src/internal/client/methods.ts
```diff
@@ -378,11 +385,19 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+            try {
+              const metadataCid = resolveIpfsCid(dao.metadata);
+              return this.ipfs.fetchString(metadataCid).then(
+                (stringMetadata) => {
+                  const metadata = JSON.parse(stringMetadata);
+                  return toDaoListItem(dao, metadata);
+                },
+              );
+            } catch (err) {
+              console.warn(err);
```
brickpop:
```suggestion
```

modules/client/src/internal/client/methods.ts
```diff
@@ -378,11 +385,19 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-            });
+          (dao: SubgraphDaoListItem): Promise<DaoListItem> | DaoListItem => {
+            try {
+              const metadataCid = resolveIpfsCid(dao.metadata);
+              return this.ipfs.fetchString(metadataCid).then(
+                (stringMetadata) => {
+                  const metadata = JSON.parse(stringMetadata);
+                  return toDaoListItem(dao, metadata);
+                },
+              );
```
brickpop: IMO the catch should happen here. We don't need the try/catch.

```
).catch(err => {
   // ... here
});
```

modules/client/src/erc20/internal/client/methods.ts
brickpop: Also, IPFS fetching could also fail because the data isn't available. We should make a difference here. Identify if the thrown error is because of the invalid IPFS Link.... otherwise, I would return a metadata object with `title: "(metadata not available)"`

modules/client/src/addressList/internal/client/methods.ts
josemarinas: Left it here for the end user to know that there is an error but its not fatal, I'll delete it if you want

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -346,13 +356,26 @@ export class ClientAddressListMethods extends ClientCore
         addressListProposals.map(
           (
             proposal: SubgraphAddressListProposalListItem,
-          ): Promise<AddressListProposalListItem> => {
-            // format in the metadata field
+          ):
+            | Promise<AddressListProposalListItem>
```
brickpop: Is this valid TS syntax?

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -346,13 +356,26 @@ export class ClientAddressListMethods extends ClientCore
...
+            | AddressListProposalListItem => {
```
brickpop: We should always return a promise, not both

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -275,9 +277,17 @@ export class ClientAddressListMethods extends ClientCore
...
         return null;
       }
-      // format in the metadata field
-      const metadataString = await this.ipfs.fetchString(addressListProposal.metadata);
-      // TODO: Parse and validate schema
+      let metadataCid = "";
+      try {
+        metadataCid = resolveIpfsCid(addressListProposal.metadata);
+        // TODO: Parse and validate schema
+      } catch (err) {
```
brickpop, line 290: We have another try/catch block outside, why not reusing it and detecting the error type?

modules/client/src/client-common/constants.ts
```diff
@@ -0,0 +1,8 @@
+import { ProposalMetadata } from "./interfaces/plugin";
+
+export const UNSUPPORTED_PROPOSAL_METADATA_LINK: ProposalMetadata = {
```
brickpop, line 3: I strongly suggest we have another template for cases where metadata cannot be fetched despite the link being valid. Otherwise one proposal with unpinned metadata would make everything fail

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -277,8 +279,18 @@ export class ClientErc20Methods extends ClientCore
         return null;
       }
       // format in the metadata field
-      const metadataString = await this.ipfs.fetchString(erc20VotingProposal.metadata);
-      // TODO: Parse and validate schema
+      let metadataCid = "";
+      try {
+        metadataCid = resolveIpfsCid(erc20VotingProposal.metadata);
+        // TODO: Parse and validate schema
+      } catch (err) {
```
brickpop, line 292: Same as above, double try/catch

modules/client/src/internal/client/methods.ts
```diff
@@ -336,11 +335,17 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
       }
-      const stringMetadata = await this.ipfs.fetchString(
-        dao.metadata,
-      );
-      const metadata = JSON.parse(stringMetadata);
+      let metadataCid = "";
+      try {
+        metadataCid = resolveIpfsCid(dao.metadata);
+        // TODO: Parse and validate schema
+      } catch {
```
brickpop: Same as above

modules/client/src/internal/constants.ts
```diff
@@ -7,3 +8,9 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
   DAO__factory.createInterface().getFunction("grant").format("minimal"),
   DAO__factory.createInterface().getFunction("freeze").format("minimal"),
 ];
+
+export const UNSUPPORTED_DAO_METADATA_LINK: IMetadata = {
```
brickpop, line 12: Same suggestion, metadata fetching could fail because something is unpinned for example

modules/client/src/addressList/internal/client/methods.ts
josemarinas: Because we need to do the return and we dont have `dao`, `proposalAddressList` and `proposalErc20` inside the the outer catch

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -46,6 +48,10 @@ import { AllowlistVoting__factory } from "@aragon/core-contracts-ethers";
 import { id } from "@ethersproject/hash";
 import { hexZeroPad } from "@ethersproject/bytes";
 import { toUtf8Bytes } from "@ethersproject/strings";
+import {
+  UNAVAILABLE_PROPOSAL_METADATA_LINK,
```
brickpop: The thing that is unavailable is the metadata, not the link IMO
```suggestion
  UNAVAILABLE_PROPOSAL_METADATA,
```

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -344,17 +363,29 @@ export class ClientAddressListMethods extends ClientCore
...
-                return toAddressListProposalListItem(proposal, metadata);
-              });
+            try {
+              const metadataCid = resolveIpfsCid(proposal.metadata);
+              const stringMetadata = await this.ipfs.fetchString(metadataCid);
+              const metadata = JSON.parse(stringMetadata) as ProposalMetadata;
+              return toAddressListProposalListItem(proposal, metadata);
+            } catch (err) {
+              if (err instanceof InvalidCidError) {
+                return Promise.resolve(
```
brickpop: If the inner function is `async` then we don't need to wrap everything with `Promise.resolve()`

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -344,17 +363,29 @@ export class ClientAddressListMethods extends ClientCore
...
+                  toAddressListProposalListItem(
+                    proposal,
+                    UNSUPPORTED_PROPOSAL_METADATA_LINK,
+                  ),
+                );
+              }
+              return Promise.resolve(toAddressListProposalListItem(
```
brickpop: Same here

modules/client/src/client-common/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  title: "Unsupported metadata link",
+  summary: "Unsupported metadata link",
+  description: "Unsupported metadata link",
+  resources: [],
+};
+export const UNAVAILABLE_PROPOSAL_METADATA_LINK: ProposalMetadata = {
+  title: "Unavailable metadata link",
+  summary: "Unavailable metadata link",
+  description: "Unavailable metadata link",
```
brickpop:
```suggestion
export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
  title: "(unavailable metadata)",
  summary: "(the proposal metadata is not available)",
  description: "(the proposal metadata is not available)",
```

modules/client/src/client-common/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { ProposalMetadata } from "./interfaces/plugin";
+
+export const UNSUPPORTED_PROPOSAL_METADATA_LINK: ProposalMetadata = {
+  title: "Unsupported metadata link",
+  summary: "Unsupported metadata link",
+  description: "Unsupported metadata link",
```
brickpop:
```suggestion
  title: "(unsupported metadata link)",
  summary: "(the link to the metadata is not supported)",
  description: "(the link to the metadata is not supported)",
```

modules/client/src/erc20/internal/client/methods.ts
```diff
@@ -337,17 +357,27 @@ export class ClientErc20Methods extends ClientCore
...
+              if (err instanceof InvalidCidError) {
+                return Promise.resolve(toErc20ProposalListItem(
+                  proposal,
+                  UNSUPPORTED_PROPOSAL_METADATA_LINK,
+                ));
+              }
+              return Promise.resolve(toErc20ProposalListItem(
+                proposal,
+                UNAVAILABLE_PROPOSAL_METADATA_LINK,
+              ));
```
brickpop:
```suggestion
                return toErc20ProposalListItem(
                  proposal,
                  UNSUPPORTED_PROPOSAL_METADATA_LINK,
                );
              }
              return toErc20ProposalListItem(
                proposal,
                UNAVAILABLE_PROPOSAL_METADATA_LINK,
              );
```

modules/client/src/internal/constants.ts
```diff
@@ -7,3 +8,15 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
...
+  name: "Unsupported metadata link",
+  description: "Unsupported metadata link",
```
brickpop:
```suggestion
  name: "(unsupported metadata link)",
  description: "(the metadata link is not supported)",
```

modules/client/src/internal/constants.ts
```diff
@@ -7,3 +8,15 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
...
+  links: [],
+};
+
+export const UNAVAILABLE_DAO_METADATA_LINK: IMetadata = {
+  name: "Unavailable metadata link",
+  description: "Unavailable metadata link",
```
brickpop:
```suggestion
  name: "(unavailable metadata)",
  description: "(the DAO metadata is not available)",
```

PR #124, opened 2022-12-16
Feature: Add admin governance plugin client

josemarinas:
## Description
Create a client for the admin governance plugin

Task: [APP-1315](https://aragonassociation.atlassian.net/browse/APP-1315)

## Type of change

## Checklist:

[APP-1315]: https://aragonassociation.atlassian.net/browse/APP-1315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/examples.md
```diff
@@ -2350,3 +2350,297 @@ console.log(members);
...
+### Create an Admin client
+
+```ts
+import { ClientAdmin, Context, ContextPlugin } from "@aragon/sdk-client";
+import { contextParams } from "../00-client/00-context";
+
+const context = new Context(contextParams);
+const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
+
+const client = new ClientAdmin(contextPlugin);
```
brickpop: ClientAdmin or AdminClient? Probably the current is consistent with the rest but I see the upcoming MultisigClient, OvoteClient, TokenVotingClient... making more sense :-/

modules/client/examples.md
```diff
@@ -2350,3 +2350,297 @@ console.log(members);
...
+
+console.log(client);
+```
+
+### Creating a admin proposal
```
brickpop: Creating and executing a proposal right away

modules/client/examples.md
```diff
@@ -2350,3 +2350,297 @@ console.log(members);
...
+    header: "https://...",
+  },
+};
+
+const ipfsUri = await client.methods.pinMetadata(metadata);
+
+const proposalParams: ExecuteProposalParams = {
+  pluginAddress: "0x1234567890123456789012345678901234567890",
+  metadataUri: ipfsUri,
+  actions: [],
```
brickpop: No actions?

modules/client/examples.md
```diff
@@ -2350,3 +2350,297 @@ console.log(members);
...
+// Create a simple context
+const context: Context = new Context(contextParams);
+// Create a plugin context from the simple context
+const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
+// Create an address list client
+const client = new ClientAdmin(contextPlugin);
+
+const queryParams: IProposalQueryParams = {
+  skip: 0, // optional
+  limit: 10, // optional,
```
brickpop: Didn't we say that subgraph doesn't allow pagination?

modules/client/examples.md
```diff
@@ -2350,3 +2350,297 @@ console.log(members);
...
+  status: ProposalStatus.ACTIVE, // optional
+};
+
+const proposals: AdminProposalListItem[] = await client.methods
+  .getProposals(queryParams);
+console.log(proposals);
+/*
+[
+  {
+    id: "0x12345...",
```
brickpop: This is inconsistent with the proposal ID's used elsewhere, now bigints

modules/client/src/admin/client.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ClientAdminEncoding } from "./internal/client/encoding";
+import { ClientAdminEstimation } from "./internal/client/estimation";
+import { ClientAdminMethods } from "./internal/client/methods";
+
+/**
+ * Provides a generic client with high level methods to manage and interact an Admin plugin installed in a DAO
+ */
+export class ClientAdmin extends ClientCore implements IClientAdmin {
+  public methods: IClientAdminMethods;
+  public encoding: IClientAdminEncoding;
```
brickpop: No decoding here?

modules/client/src/admin/interfaces.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export interface IClientAdminEncoding extends IClientCore {}
+
+export interface IClientAdminEstimation extends IClientCore {
+  executeProposal: (parms: ExecuteProposalParams) => Promise<GasFeeEstimation>;
+}
+
+export interface IClientAdmin {
+  methods: IClientAdminMethods;
+  encoding: IClientAdminEncoding;
```
brickpop: Encoding without decoding?

modules/client/src/admin/interfaces.ts
```diff
@@ -0,0 +1,92 @@
...
+  pinMetadata: (
+    params: ProposalMetadata,
+  ) => Promise<string>;
+  getProposal: (proposalId: string) => Promise<AdminProposal | null>;
+  getProposals: (
+    params: IAdminProposalQueryParams,
+  ) => Promise<AdminProposalListItem[]>;
+}
+
+export interface IClientAdminEncoding extends IClientCore {}
```
brickpop: Nothing to encode?

modules/client/src/admin/interfaces.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export type ExecuteProposalParams = {
+  pluginAddress: string;
+  metadataUri: string;
+  actions: DaoAction[];
+};
+
+type ProposalBase = {
+  id: string;
```
brickpop, line 45: id? or proposal id below?

modules/client/src/admin/interfaces.ts
```diff
@@ -0,0 +1,92 @@
...
+  actions: DaoAction[];
+  pluginAddress: string;
+  proposalId: bigint;
+};
+export type AdminProposalListItem = ProposalBase;
+
+export interface IAdminProposalQueryParams extends IPagination {
+  sortBy?: ProposalSortBy;
+  status?: ProposalStatus;
+  administratorAddressOrEns?: string;
```
brickpop:
```suggestion
  adminAddressOrEns?: string;
```

modules/client/src/admin/interfaces.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+export type SubgraphAdminProposalListItem = SubgraphAdminProposalBase;
+
+export type SubgraphAdminProposal = SubgraphAdminProposalBase & {
+  actions: SubgraphAction[];
+  plugin: {
+    address: string;
+  };
+  proposalId: string;
```
brickpop, line 93: Same

modules/client/src/admin/internal/client/methods.ts
```diff
@@ -0,0 +1,229 @@
...
+      if (!adminProposal) {
+        return null;
+      }
+
+      try {
+        const metadataCid = resolveIpfsCid(adminProposal.metadata);
+        const metadataString = await this.ipfs.fetchString(metadataCid);
+        const metadata = JSON.parse(metadataString) as ProposalMetadata;
+        return toAdminProposal(adminProposal, metadata);
+        // TODO: Parse and validate schema
```
brickpop, line 141: Do we need a follow up task?

modules/client/test/integration/admin/methods.test.ts
```diff
@@ -0,0 +1,260 @@
...
+            }],
+          }),
+        ),
+      );
+
+      const proposal = await client.methods.getProposal(proposalId);
+
+      expect(typeof proposal).toBe("object");
+      expect(proposal === null).toBe(false);
+      if (proposal) {
```
brickpop: Isn't it simpler to do `if(!proposal) throw new Error()`? No need for the reverse if

modules/client/examples.md
josemarinas: Okay, we can change it here and make a task to change it everywhere

modules/client/examples.md
josemarinas: Well this is not a pagination, we cannot specify a page because we can't have the total amount of entries of a particular object, for example proposals. We can add a variable to count the number of proposals, the problem that we cannot have the total amount of proposals in a specific state, because the states mainly depend on the current time so from subgraph we cannot detect when a proposal passes from a `pending` state to a `active` state or when it passes from an `active` state to `succeded` or `defeated`, because this changes depend on the `startDate` and `endDate` of the proposal, and there is no way of tracking that in subgraph. 
This was what the UI requested and what the task that we closed was about

modules/client/examples.md
josemarinas: I know, but the subgraph is using `pluginAdddress_proposalId` as `id` because it has to be unique, the bigint proposal ID is in the field `proposalId`

modules/client/src/admin/client.ts
josemarinas: We only have one encoding function, that is the initialize function, so there is nothing to decode

modules/client/src/admin/interfaces.ts
josemarinas: id = `pluginAddress_proposalId`, has to be unique on subgraph

modules/client/src/admin/internal/client/methods.ts
josemarinas: If we merge this one first I'll add the validation on the validation branch, if we merge the validation branch first i'll add it on this branch

PR #129, opened 2023-01-05
Feature: Update `ethers-contracts` version

josemarinas:
## Description

Update ethers contracts and adapts events, responses...

Task: [APP-1387](https://aragonassociation.atlassian.net/browse/APP-1387)

## Type of change

## Checklist:

[APP-1387]: https://aragonassociation.atlassian.net/browse/APP-1387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -1119,10 +1119,11 @@ import {
   Context,
   DaoCreationSteps,
   GasFeeEstimation,
-  IAddressListPluginInstall,
   ICreateParams,
+  IMetadata,
+  VotingMode,
+  VotingSettings,
 } from "@aragon/sdk-client";
-import { IMetadata } from "../../src";
 import { contextParams } from "../00-client/00-context";
 
 const context: Context = new Context(contextParams);
@@ -1130,14 +1131,13 @@ const client: Client = new Client(context);
 
 // Define the plugins to install and their params
 
-const pluginInitParams: IAddressListPluginInstall = {
+const pluginInitParams: VotingSettings = {
   votingSettings: {
     minDuration: 60 * 60 * 24 * 2, // seconds
     minParticipation: 0.25, // 25%
     supportThreshold: 0.5, // 50%
     minProposerVotingPower: BigInt("5000"), // default 0
-    earlyExecution: true, // default false
-    voteReplacement: false, // default false,
+    votingMode: VotingMode.VOTE_REPLACEMENT,
   },
   addresses: [
     "0x1234567890123456789012345678901234567890",
@@ -1862,7 +1862,6 @@ import {
   VotingMode,
   VotingSettings,
 } from "@aragon/sdk-client";
-import {} from "../../src";
 import { contextParams } from "../00-client/00-context";
 
 const context: Context = new Context(contextParams);
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -466,6 +466,7 @@ import {
   ICreateParams,
   ITokenVotingPluginInstall,
 } from "@aragon/sdk-client";
+import { IMetadata } from "../../src";
 import { contextParams } from "../00-client/00-context";
 
 const context: Context = new Context(contextParams);
@@ -475,10 +476,13 @@ const client: Client = new Client(context);
 
 // Use an already existing ERC20 token
 const pluginInitParams1: ITokenVotingPluginInstall = {
-  settings: {
+  votingSettings: {
     minDuration: 60 * 60 * 24 * 2, // seconds
-    minTurnout: 0.25, // 25%
-    minSupport: 0.5, // 50%
+    minParticipation: 0.25, // 25%
+    supportThreshold: 0.5, // 50%
+    minProposerVotingPower: BigInt("5000"),
+    earlyExecution: true,
+    voteReplacement: false
   },
   useToken: {
     address: "0x...",
@@ -486,10 +490,13 @@ const pluginInitParams1: ITokenVotingPluginInstall = {
 };
 // Mint a new token
 const pluginInitParams2: ITokenVotingPluginInstall = {
-  settings: {
-    minDuration: 60 * 60 * 24, // seconds
-    minTurnout: 0.4, // 40%
-    minSupport: 0.55, // 55%
+  votingSettings: {
+    minDuration: 60 * 60 * 24 * 2, // seconds
+    minParticipation: 0.25, // 25%
+    supportThreshold: 0.5, // 50%
+    minProposerVotingPower: BigInt("5000"),
+    earlyExecution: true,
+    voteReplacement: false
   },
   newToken: {
     name: "Token",
@@ -519,18 +526,22 @@ const tokenVotingInstallPluginItem2 = TokenVotingClient.encoding.getPluginInstal
   pluginInitParams2,
 );
 
+const daoMetadata: IMetadata = {
+  name: "My DAO",
+  description: "This is a description",
+  avatar: "",
+  links: [{
+    name: "Web site",
+    url: "https://...",
+  }],
+};
+
+const metadataUri = await client.methods.pinMetadata(daoMetadata);
+
 const createParams: ICreateParams = {
-  metadata: {
-    name: "My DAO",
-    description: "This is a description",
-    avatar: "https://...",
-    links: [{
-      name: "Web site",
-      url: "https://...",
-    }],
-  },
+  metadataUri,
   ensSubdomain: "my-org", // my-org.dao.eth
-  plugins: [tokenInstallPluginItem1, tokenInstallPluginItem2],
+  plugins: [tokenVotingInstallPluginItem1, tokenVotingInstallPluginItem2],
 };
 
 // gas estimation
@@ -575,7 +586,7 @@ contextPlugin.setFull(contextParams);
 console.log(contextPlugin)
 ```
 
-### Create a TokenVoting client
+### Create an TokenVoting client
 
 ```ts
 import { TokenVotingClient, Context, ContextPlugin } from "@aragon/sdk-client";
@@ -589,7 +600,7 @@ const client = new TokenVotingClient(contextPlugin);
 console.log(client);
 ```
 
-### Creating a TokenVoting proposal
+### Creating an TokenVoting proposal
 
 ```ts
 import {
@@ -607,7 +618,7 @@ import { contextParams } from "../00-client/00-context";
 const context: Context = new Context(contextParams);
 // Create a plugin context from the simple context
 const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
-// Create a TokenVoting client
+// Create an TokenVoting client
 const client = new TokenVotingClient(contextPlugin);
 
 const metadata:ProposalMetadata= {
@@ -659,7 +670,7 @@ for await (const step of steps) {
 }
 ```
 
-### Creating a TokenVoting proposal with an action
+### Creating an TokenVoting proposal with an action
 
 ```ts
 import {
@@ -667,9 +678,9 @@ import {
   Context,
   ContextPlugin,
   ICreateProposalParams,
-  IPluginSettings,
   ProposalCreationSteps,
   VoteValues,
+  VotingSettings,
 } from "@aragon/sdk-client";
 import { contextParams } from "../00-client/00-context";
 
@@ -677,14 +688,17 @@ import { contextParams } from "../00-client/00-context";
 const context: Context = new Context(contextParams);
 // Create a plugin context from the simple context
 const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
-// Create a TokenVoting client
+// Create an TokenVoting client
 const client = new TokenVotingClient(contextPlugin);
 
 // create config action
-const configActionPrarms: IPluginSettings = {
-  minDuration: 60 * 60 * 24,
-  minSupport: 0.3, // 30%
-  minTurnout: 0.5, // 50%
+const configActionPrarms: VotingSettings = {
+  minDuration: 60 * 60 * 24 * 2, // seconds
+  minParticipation: 0.25, // 25%
+  supportThreshold: 0.5, // 50%
+  minProposerVotingPower: BigInt("5000"),
+  earlyExecution: true,
+  voteReplacement: false
 };
 
 const pluginAddress = "0x1234567890123456789012345678901234567890";
@@ -696,25 +710,7 @@ const configAction = client.encoding.updatePluginSettingsAction(
 
 const proposalParams: ICreateProposalParams = {
   pluginAddress: "0x1234567890123456789012345678901234567890",
-  metadata: {
-    title: "Test Proposal",
-    summary: "This is a short description",
-    description: "This is a long descrioption",
-    resources: [
-      {
-        name: "Discord",
-        url: "https://discord.com/...",
-      },
-      {
-        name: "Website",
-        url: "https://website...",
-      },
-    ],
-    media: {
-      logo: "https://...",
-      header: "https://...",
-    },
-  },
+  metadataUri: "ipfs://123456",
   actions: [configAction],
   startDate: new Date(),
   endDate: new Date(),
@@ -739,7 +735,7 @@ for await (const step of steps) {
 }
 ```
 
-### Voting on a TokenVoting proposal
+### Voting on an TokenVoting proposal
 
 ```ts
 import {
@@ -756,7 +752,7 @@ import { contextParams } from "../00-client/00-context";
 const context: Context = new Context(contextParams);
 // Create a plugin context from the simple context
 const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
-// Create a TokenVoting client
+// Create an TokenVoting client
 const client = new TokenVotingClient(contextPlugin);
 
 const voteParams: IVoteProposalParams = {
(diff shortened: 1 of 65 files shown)


modules/client/examples.md
```diff
@@ -475,21 +476,27 @@ const client: Client = new Client(context);
...
 // Use an already existing ERC20 token
 const pluginInitParams1: ITokenVotingPluginInstall = {
-  settings: {
+  votingSettings: {
     minDuration: 60 * 60 * 24 * 2, // seconds
-    minTurnout: 0.25, // 25%
-    minSupport: 0.5, // 50%
+    minParticipation: 0.25, // 25%
+    supportThreshold: 0.5, // 50%
+    minProposerVotingPower: BigInt("5000"),
```
brickpop: Is this parameter optional? If by default it is 0, I would add a comment to the right (`Default: 0`)

modules/client/examples.md
```diff
@@ -575,7 +586,7 @@ contextPlugin.setFull(contextParams);
 console.log(contextPlugin)
 ```
 
-### Create a TokenVoting client
+### Create an TokenVoting client
```
brickpop: This is introducing a regression. Mathias fixed this here, but probably it needs fixing on the source files as well?

modules/client/examples.md
```diff
@@ -815,7 +810,7 @@ true
 
 ### Loading the list of members (TokenVoting)
 
-Retrieving all the members of a TokenVoting DAO.
+Retrieving all the members of an Token DAO.
```
brickpop: Do we want this change?

modules/client/examples.md
```diff
@@ -920,7 +915,6 @@ console.log(proposal);
     address: "0x1234567890123456789012345678901234567890,
     name: "The Token",
     symbol: "TOK",
-    decimals: 18
```
brickpop, line 923: Just to confirm, aren't we returning the decimals now?

modules/client/examples.md
```diff
@@ -1392,7 +1411,6 @@ for await (const step of steps) {
         console.log(step.txHash);
         break;
       case VoteProposalStep.DONE:
-        console.log(step.voteId);
```
brickpop, line 1395: Removing this line?

modules/client/src/addressList/interfaces.ts
```diff
@@ -43,13 +44,13 @@ export interface IClientAddressListMethods extends IClientCore {
   getProposals: (
     params: IProposalQueryParams,
   ) => Promise<AddressListProposalListItem[]>;
-  getSettings: (pluginAddress: string) => Promise<IPluginSettings | null>;
```
brickpop, line 46: If we now have "voting" settings, as opposed to just settings, is there any other type of settings that might need a separate `getXXXSettings`?  If so, we might want to rename `getSettings` to `getVotingSettings`

modules/client/src/addressList/internal/graphql-queries/proposal.ts
```diff
@@ -1,8 +1,8 @@
 import { gql } from "graphql-request";
 
-export const QueryAddressListProposal = gql`
-query addressListProposal($proposalId: ID!) {
-  addresslistProposal(id: $proposalId){
+export const QueryAddressListVotingProposal = gql`
+query addressListVotingProposal($proposalId: ID!) {
```
brickpop, line 4: This query has a lowercase first character on the name, whereas above, `AddressListVotingPluginMembers` has an uppercase one.

Does this come from Subgraph?

modules/client/src/client-common/encoding.ts
```diff
@@ -7,55 +10,75 @@ import {
...
+// export function votingSettingsToContract(params:VotingSettings):ContractVotingSettings {
+//   return   [
+//     BigNumber.from(getVotingMode(params)),
+//     BigNumber.from(encodeRatio(params.supportThreshold, 2)),
+//     BigNumber.from(encodeRatio(params.minParticipation, 2)),
+//     BigNumber.from(params.minDuration),
+//     BigNumber.from(params.minProposerVotingPower || 1)
+//   ]
+// }
+
```
brickpop: Is this code needed?

brickpop (commented):
Round 1

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -31,14 +30,25 @@ export interface IProposalSettings {
...
-  /** Float: 0 to 1 */
-  minSupport: number;
-  /** Float: 0 to 1 */
-  minTurnout: number;
-  /** In seconds */
+
+export type VotingSettings = {
+  earlyExecution?: boolean;
+  voteReplacement?: boolean;
+  // Float between 0 and 1
```
brickpop:
```suggestion
  /** Float between 0 and 1 */
```

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -31,14 +30,25 @@ export interface IProposalSettings {
...
+  supportThreshold: number;
+  // Float between 0 and 1
```
brickpop:
```suggestion
  /** Float between 0 and 1 */
```

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -203,7 +213,7 @@ export enum VoteProposalStep {
 
 export type VoteProposalStepValue =
   | { key: VoteProposalStep.VOTING; txHash: string }
-  | { key: VoteProposalStep.DONE; voteId: string };
+  | { key: VoteProposalStep.DONE };
```
brickpop, line 222: Don't we have the ID of the vote anymore?

modules/client/src/client-common/utils.ts
```diff
@@ -75,3 +80,28 @@ export function isProposalId(propoosalId: string): boolean {
   const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
   return regex.test(propoosalId);
 }
+
+export function findLog(
+  receipt: ContractReceipt,
+  iface: Interface,
+  event: string,
```
brickpop: Would eventName be more explicit?

modules/client/examples.md
josemarinas: Yes, this returned the `proposalId` for some reason and didn't make much sense. In the contracst there is no `voteId` so I removed it

modules/client/src/client-common/utils.ts
```diff
@@ -75,3 +80,28 @@ export function isProposalId(propoosalId: string): boolean {
...
+        id(
+          iface.getEvent(event).format(
+            "sighash",
+          ),
+        ),
+  );
+}
+
+export function getVotingMode(votingSettings: VotingSettings): number {
+  if (votingSettings.earlyExecution) {
```
brickpop: If this function is to be used from subgraph data, I guess it's fine. But if it handles data provided by the user, we should throw an error if both booleans are set to `true`

I would rather encapsulate this logic here, instead of having this check duplicated in other places

modules/client/src/internal/utils.ts
```diff
@@ -152,7 +152,9 @@ export function toTransfer(transfer: SubgraphTransferListItem): Transfer {
       creationDate,
       reference: transfer.reference,
       transactionId: transfer.transaction,
-      proposalId: transfer.proposal.id || "",
+      proposalId: transfer.proposal && transfer.proposal.id
+        ? transfer.proposal.id
+        : "",
```
brickpop:
```suggestion
      proposalId: transfer.proposal?.id || "",
```

modules/client/src/tokenVoting/interfaces.ts
```diff
@@ -43,22 +45,24 @@ export interface ITokenVotingClientMethods extends IClientCore {
   getProposals: (
     params: IProposalQueryParams,
   ) => Promise<TokenVotingProposalListItem[]>;
-  getSettings: (pluginAddress: string) => Promise<IPluginSettings | null>;
-  getToken: (pluginAddress: string) => Promise<Erc20TokenDetails | null>;
+  getSettings: (pluginAddress: string) => Promise<VotingSettings | null>;
```
brickpop: Same as before, I would consider getVotingSettings if it makes sense

modules/client/src/addressList/internal/graphql-queries/proposal.ts
josemarinas: The thing is that in subgraph and contracts we have `addresslistVoting`, and in the sdk we have a mix of `addressList` and `addressListVoting` I would do de change on the sdk to unify the naming but I'd do it in a separate PR that also addreses another naming changes such as `ClientAddressList` => `AddresslistVotingClient`

brickpop (commented):
Round 2

modules/client/examples.md
josemarinas: Nope, but it was changed on the examples.md but wasnt on the example itself so when i regenereated the example it changed

modules/client/src/addressList/interfaces.ts
josemarinas: Nope, what I was thinking was to go back on the voting mode, right now it is handled with the booleans but maybe it makes mor sense to use an enum because its easier to handle and return

modules/client/examples.md
josemarinas: no we are not, i deleted it but i fixed it, idk why this appears on the diff on local does not appear like this

modules/client/src/client-common/utils.ts
josemarinas: Yes, It makes sense

modules/client/src/client-common/interfaces/plugin.ts
josemarinas: Same as before I couldnt find an event thrown in the contracts

modules/client/src/client-common/utils.ts
josemarinas: I was planning on making this check with yup thats why I avoided it here

modules/client/examples.md
```diff
@@ -2176,16 +2205,16 @@ import {
...
 import { contextParams } from "../00-client/00-context";
 const context: Context = new Context(contextParams);
 // Create a plugin context from the simple context
 const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
-const clientAddressList = new TokenVotingClient(contextPlugin);
+const clientToken = new TokenVotingClient(contextPlugin);
 const data: Uint8Array = new Uint8Array([12, 56]);
 
-const params: IPluginSettings = clientAddressList.decoding
+const params: VotingSettings = clientToken.decoding
```
brickpop: Maybe `clientTokenVoting`? on the source file
It's ok like this too :+1:

modules/client/src/addressList/interfaces.ts
brickpop: My point was about the function name (getSettings). Wether to keep it with a generic name (when it only has part of the settings) or not.

Regarding the booleans/enum's, whatever is more comfortable for a developer. I think the App team will have two separate checkboxes but this may change at any time.

modules/client/src/addressList/internal/graphql-queries/proposal.ts
brickpop: Added to APP-1514 👍

modules/client/src/client-common/utils.ts
brickpop: Makes sense. But in my opinion, a function like this should not be allowing an undefined state to return a valid value.

Returning `1` when both are true is an error in my view. Yup may filter one source of errors but we could still use this function from elsewhere and miss a bug

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -427,17 +429,14 @@ export class TokenVotingClientMethods extends ClientCore
       }
       return {
         minDuration: parseInt(tokenVotingPlugin.minDuration),
-        minSupport: decodeRatio(
-          parseFloat(
-            tokenVotingPlugin.totalSupportThresholdPct,
-          ),
-          2,
+        supportThreshold: parseFloat(
+          formatEther(tokenVotingPlugin.supportThreshold),
```
brickpop, line 433: Why not using our own decoding? We have a separate task to handle this, and I think there's another code section where such change is not made in this PR

brickpop (commented):
Final round

modules/client/src/addressList/interfaces.ts
josemarinas: Yeah the other settings are handled with `getMembers` for the addresslist and `getToken`for the tokenvoting, I dont plan on having a generic `getSettings`that resturns all settings, but maybe it makes changing it to `getVotingSettings` sense for the sake of clarity and consistency.

modules/client/src/tokenVoting/internal/client/methods.ts
josemarinas: Our encoder/decoder stopeed making sense when we need to transform 0.5 => 5*10^17 because the precission is not good enough we have like 15 digits so I decide to use `formatEther`and `parseEther` that have the exact behaivour that we want. This was just a test of both this functions to see if I coulñd use them in the other PR, it worked and left it there

modules/client/src/addressList/interfaces.ts
josemarinas: About the enum think I believe I'm going to switch to enum, for me as a developer seems more straigthforward using and enum for things that cannot happen at the same time and booleans for the things that can happen, for example if `earlyExecution` and `voteReplacement` could be true at the same time this would make sense to be a boolean since this is not going to happen, imo it does not make sense

modules/client/examples.md
```diff
@@ -481,8 +482,7 @@ const pluginInitParams1: ITokenVotingPluginInstall = {
     minParticipation: 0.25, // 25%
     supportThreshold: 0.5, // 50%
     minProposerVotingPower: BigInt("5000"), // default 0
-    earlyExecution: true, // default false
-    voteReplacement: false, // default false,
+    votingMode: VotingMode.STANDARD, // default standard
```
brickpop, line 485: I'm not sure "standard" is easy enough to understand, as I don't think there is a "standard" way of voting. If the enum 1:1 matches the smart contract one, then we might rather address this at the contract level and not here.

- Would "DEFAULT" be any more logical for an SDK consumer?
- Do we have any example for vote replacement?
- Otherwise, I would maybe leave the two booleans with a default to false, and throwing an error when both are improperly combined

modules/client/examples.md
```diff
@@ -1052,15 +1051,17 @@ const client = new ClientAddressList(contextPlugin);
 
 const pluginAddress: string = "0x1234567890123456789012345678901234567890";
 
-const settings: VotingSettings | null = await client.methods.getSettings(
+const settings: VotingSettings | null = await client.methods.getVotingSettings(
```
brickpop, line 1054: Could we add `getVotingSettings` to the Changelog?

modules/client/examples/03-encoders/06-update-plugin-settings-address-list.ts
```diff
@@ -5,8 +5,10 @@ import {
   ClientAddressList,
   Context,
   ContextPlugin,
+  VotingMode,
   VotingSettings,
 } from "@aragon/sdk-client";
+import {} from "../../src";
```
brickpop:
```suggestion
```

modules/client/src/addressList/internal/client/encoding.ts
```diff
@@ -79,7 +79,7 @@ export class ClientAddressListEncoding extends ClientCore
     return {
       to: pluginAddress,
       value: BigInt(0),
-      data: encodeUpdatePluginSettingsAction(params),
+      data: encodeUpdateVotingSettingsAction(params),
```
brickpop, line 82: Could we add this to the changelog?

modules/client/src/tokenVoting/internal/client/encoding.ts
```diff
@@ -79,7 +79,7 @@ export class TokenVotingClientEncoding extends ClientCore
     return {
       to: pluginAddress,
       value: BigInt(0),
-      data: encodeUpdatePluginSettingsAction(params),
+      data: encodeUpdateVotingSettingsAction(params),
```
brickpop, line 82: Changelog

modules/common/test/unit/encoding.test.ts
```diff
@@ -325,6 +325,7 @@ describe("Value encoding", () => {
       { bigint: BigInt("367483947"), digits: 9, out: 0.367483947 },
       { bigint: 1, digits: 9, out: 0.000000001 },
       { bigint: 367483947, digits: 9, out: 0.367483947 },
+      { bigint: 367483947, digits: 9, out: 0.367483947 },
```
brickpop: Why a duplicate?

modules/client/examples.md
josemarinas: Yes this matches 1:1 the contracts, https://github.com/aragon/core/blob/420c6346cb750d58b434d3d98a6178c24c4729fd/packages/contracts/test/test-utils/voting.ts#L12
 No we dont have an example for vote replacement

brickpop: Because it will be implemented later?

brickpop (approved):

modules/client/examples.md
josemarinas: I added an example with `VotingMode.VOTE_REPLACEMENT`, it is exactly the same 👍

PR #132, opened 2023-01-10
Fix: Precission in VotingSettings

josemarinas:
## Description

Fix precission in Voting Setings

0.5 in sdk => 5x(10^17) in contracts

Task: [APP-1142](https://aragonassociation.atlassian.net/browse/APP-1142)

## Type of change

## Checklist:

[APP-1142]: https://aragonassociation.atlassian.net/browse/APP-1142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/tokenVoting/internal/utils.ts
```diff
@@ -79,21 +79,6 @@ export function toTokenVotingProposal(
...
-      // this should be decoded using the number of decimals that we want
-      // right now the encoders/recoders use 2 digit precission but the actual
-      // subgraph values are 18 digits precision. Uncomment below for 2 digits
-      // precision
-
-      // minSupport: decodeRatio(
-      //   BigInt(proposal.totalSupportThresholdPct),
-      //   2,
-      // ),
-      // minTurnout: decodeRatio(
```
brickpop: If the solution comes later, removing this will make things a bit slower

brickpop (approved):

modules/client/src/tokenVoting/internal/utils.ts
josemarinas: My understanding is that this is not going to change, so I dont see a problem. 

The subgraph is settled in using 18 digits and we are ettled in using a number between 0 and 1

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -436,10 +436,14 @@ export class ClientAddressListMethods extends ClientCore
       return {
         minDuration: parseInt(addresslistVotingPlugin.minDuration),
         supportThreshold: parseFloat(
-          formatEther(addresslistVotingPlugin.supportThreshold),
+          parseFloat(
+            formatEther(addresslistVotingPlugin.supportThreshold),
+          ).toFixed(2),
```
brickpop: Why don't we encapsulate this into a helper? This is verbose and spotting two parseFloat() together is weird

modules/client/src/tokenVoting/internal/utils.ts
brickpop: There is a task on the contract side to standardize the precision numbers (APP-1264).

Having 18 digits may make sense for token balances, but not to indicate a ratio

josemarinas: Hmm, ok, nice to know then i'll leave this comment

modules/client/src/addressList/internal/client/methods.ts
josemarinas: Sure, makes sense

modules/client/src/addressList/internal/client/methods.ts
```diff
@@ -22,6 +22,7 @@ import {
   ClientCore,
   computeProposalStatusFilter,
   ContextPlugin,
+  etherToUnitInterval,
```
brickpop: interval? Something like `ratio` feels more meaningful. Maybe `parseEtherRatio()` or related?

modules/client/src/client-common/utils.ts
```diff
@@ -110,3 +111,11 @@ export function votingModeToContracts(votingMode: VotingMode): number {
       throw new InvalidVotingModeError();
   }
 }
+
+export function etherToUnitInterval(ether: string): number {
```
brickpop: how about parameterizing the amount of digits? 2 by default

modules/client/src/addressList/internal/client/methods.ts
josemarinas: Yes i used unit interval bcause is the set of numbres between 0 and 1, but I wasnt super convinced about the name

brickpop (approved):

PR #134, opened 2023-01-10
Feature: Export ensureAllowance

josemarinas:
## Description

Make ensure allowance a public callable method

Task: [APP-1510](https://aragonassociation.atlassian.net/browse/APP-1510)

## Type of change

## Checklist:

[APP-1510]: https://aragonassociation.atlassian.net/browse/APP-1510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -446,7 +446,7 @@ console.log(balances);
 */
 ```
 
-### Ensure an allowance is enouch
+### Ensure an a minimum token allowance
 Check if the allowance is enough and updates it if it is not
 
 ```ts
--- a/modules/client/examples.md
+++ b/modules/client/examples.md
@@ -446,6 +446,45 @@ console.log(balances);
 */
 ```
 
+### Ensure an allowance is enouch
+Check if the allowance is enough and updates it if it is not
+
+```ts
+import {
+  Client,
+  Context,
+  DaoDepositSteps,
+  EnsureAllowanceParams,
+} from "@aragon/sdk-client";
+import { contextParams } from "./00-context";
+
+const context = new Context(contextParams);
+const client = new Client(context);
+const ensureAllowanceParams: EnsureAllowanceParams = {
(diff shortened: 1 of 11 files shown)


modules/client/examples.md
```diff
@@ -446,6 +446,45 @@ console.log(balances);
 */
 ```
 
+### Ensure an allowance is enouch
```
brickpop:
```suggestion
### Ensure a minimum token allowance
```

brickpop: Follow-up question. If the actual contract method is `approve()`, shouldn't the whole thing be called like `ensureMinApproval()`?

modules/common/src/errors.ts
```diff
@@ -100,3 +100,8 @@ export class InvalidVotingModeError extends Error {
     super("Invalid voting mode");
   }
 }
+export class EnsureAllowanceError extends Error {
+  constructor() {
+    super("Failed to ensure allowance");
```
brickpop:
```suggestion
    super("Could not define a minimum allowance");
```

modules/client/examples.md
josemarinas: Mmm, this can lead into confusion imo, we have som parameter on the proposal that is called `minApproval`, also when you check the actual value in the contract you use `allowance` method so ensuring an allowance kinda makes sense, if i had to change it I would change it to `updateAllowance`or something similar

brickpop: mmm multisig-wise that's a very good point, but approving a transfer is ERC20-related, nothing to do with a multisig. If reading is done through `allowance()` then `ensureAllowance()` feels good, because "update" would set it no matter what, whereas "ensure" denoted a "lazy" operation

brickpop (approved):

PR #138, opened 2023-01-17
Fix: Support for ERC721 on TokenVotingClient function getProposals

josemarinas:
## Description

The graphql query didnt suport ERC721 which caused the request to return null on the token field

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/modules/client/package.json
+++ b/modules/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.17.0-alpha",
+  "version": "0.17.1-alpha",
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/sdk-client.esm.js",
--- a/modules/client/test/integration/multisig-client/methods.test.ts
+++ b/modules/client/test/integration/multisig-client/methods.test.ts
@@ -20,7 +20,6 @@ import {
   ProposalCreationSteps,
   ProposalMetadata,
   ProposalSortBy,
-  ProposalStatus,
   SortDirection,
 } from "../../../src";
 import { GraphQLError, InvalidAddressOrEnsError } from "@aragon/sdk-common";
@@ -311,12 +310,10 @@ describe("Client Multisig", () => {
       const ctxPlugin = ContextPlugin.fromContext(ctx);
       const client = new MultisigClient(ctxPlugin);
       const limit = 5;
-      const status = ProposalStatus.EXECUTED;
       const params: IProposalQueryParams = {
         limit,
         sortBy: ProposalSortBy.CREATED_AT,
         direction: SortDirection.ASC,
-        status,
       };
       const proposals = await client.methods.getProposals(params);
 
@@ -332,7 +329,6 @@ describe("Client Multisig", () => {
         expect(proposal.creatorAddress).toMatch(/^0x[A-Fa-f0-9]{40}$/i);
         expect(typeof proposal.metadata.title).toBe("string");
         expect(typeof proposal.metadata.summary).toBe("string");
-        expect(proposal.status).toBe(status);
       }
     });
     it("Should get a list of proposals from a specific dao", async () => {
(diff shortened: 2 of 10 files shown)


Fabricevladimir (approved):

modules/client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-client",
   "author": "Aragon Association",
-  "version": "0.17.0-alpha",
+  "version": "0.17.1-alpha",
```
brickpop, line 4: it's ok now, but the very next version adding features I would already label it as `beta`. Feel free to resolve

modules/client/test/integration/multisig-client/methods.test.ts
```diff
@@ -311,12 +310,10 @@ describe("Client Multisig", () => {
       const ctxPlugin = ContextPlugin.fromContext(ctx);
       const client = new MultisigClient(ctxPlugin);
       const limit = 5;
-      const status = ProposalStatus.EXECUTED;
       const params: IProposalQueryParams = {
         limit,
         sortBy: ProposalSortBy.CREATED_AT,
         direction: SortDirection.ASC,
-        status,
       };
```
brickpop, line 317: Why removing this?

modules/client/test/integration/tokenVoting-client/methods.test.ts
```diff
@@ -348,7 +348,7 @@ describe("Token Voting Client", () => {
         const ctxPlugin = ContextPlugin.fromContext(ctx);
         const client = new TokenVotingClient(ctxPlugin);
         const limit = 5;
-        const status = ProposalStatus.EXECUTED;
+        const status = ProposalStatus.DEFEATED;
```
brickpop, line 351: Curious to understand why this test is changed

brickpop (approved):

modules/client/test/integration/multisig-client/methods.test.ts
josemarinas: Currenly on subgraph we dont have any proposal thats executed so the test will fail 😓

modules/client/test/integration/tokenVoting-client/methods.test.ts
josemarinas: There are no executed proposals in subgraph

modules/client/test/integration/multisig-client/methods.test.ts
brickpop: Why don't we change the filter before/after and check that the results change from "something" to "nothing" ?

modules/client/test/integration/tokenVoting-client/methods.test.ts
brickpop: Is there some automated data population that we can run to avoid this? Or anything else @mathewmeconry

josemarinas: There are somee scripts, but I belive they are not automated, but yes it will be super cool if we can populate with test data each time we launch the contracts

modules/client/test/integration/multisig-client/methods.test.ts
josemarinas: I would't make sense, because tomorrow we could have an executed proposal and break the tests again :(

Either we automate the creation of daos and proposals each time we deploy the contracts or we start mocking the subgraph if we want to have a consistent response here

modules/client/test/integration/tokenVoting-client/methods.test.ts
mathewmeconry: I would like to mention this again. I suggest we start mocking this and don't rely on external services to have the correct data so that tests succeed.
As mentioned in multiple dev guides: unit tests shouldn't have external dependencies. External dependencies should be either mocked or stubbed to get reproducible test results

I don't see the point of wasting our goerliETH to generate dummy data so that the SDK has tests passing

mathewmeconry (approved):

modules/client/test/integration/tokenVoting-client/methods.test.ts
brickpop: @mathewmeconry let's get it done on next sprint (APP-1643) :+1:

brickpop (approved):

PR #155, opened 2023-02-15
Feature: Add voteReplacemeent flag and fix transfers and balances queries

josemarinas:
## Description
Add voteReplacemeent flag and fix transfers and balances queries

Task: [APP-1760](https://aragonassociation.atlassian.net/browse/APP-1760)

## Type of change

## Checklist:

[APP-1760]: https://aragonassociation.atlassian.net/browse/APP-1760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/examples.md
```diff
@@ -421,13 +421,13 @@ console.log(transfers);
 Handles retrieving DAO asset balances using the DAO address or its ENS domain.
 
 ```ts
-import { AssetBalance, Client, Context } from "@aragon/sdk-client";
+import { TokenBalance, Client, Context } from "@aragon/sdk-client";
```
brickpop: Is there any reason to change this type? I'm fine if the UI is not already using/importing it

modules/client/src/interfaces.ts
```diff
@@ -259,20 +262,34 @@ export enum TransferType {
...
   creationDate: Date;
   transactionId: string;
+  to: string;
+  from: string;
 };
 
 type NativeTokenTransfer = BaseTokenTransfer & {
   tokenType: TokenType.NATIVE;
+  amount: bigint;
+  reference: string;
```
brickpop: reference has been removed. At the very, very least it should be optional... but it will be completely removed because no standard support it

modules/client/src/interfaces.ts
```diff
@@ -259,20 +262,34 @@ export enum TransferType {
...
+  token: {
+    address: string;
+    name: string;
+    symbol: string;
+  };
 };
 
 type Erc20TokenTransfer = BaseTokenTransfer & {
   tokenType: TokenType.ERC20;
+  amount: bigint;
```
brickpop, line 290: ERC721 transfers have more fields than represented here. At the very least there is the ID of the token that you are transferring the ownership to someone else

I would personally leave a TO DO here, because this is not the focus now

modules/client/src/interfaces.ts
```diff
@@ -281,16 +298,18 @@ type Erc20TokenTransfer = BaseTokenTransfer & {
   };
 };
 
-export type Deposit = (NativeTokenTransfer | Erc20TokenTransfer) & {
-  from: string;
-  type: TransferType.DEPOSIT;
-};
+export type Deposit =
+  & (NativeTokenTransfer | Erc20TokenTransfer | Erc721TokenTransfer)
```
brickpop, line 300: Is it OK to start with `&` after no particle?

modules/client/src/internal/graphql-queries/transfer.ts
```diff
@@ -1,24 +1,38 @@
...
+      token{
+      	id
+        name
+        symbol
+      }
+      
+    }
+    ... on NativeTransfer {
+      amount
+      reference
```
brickpop: Reference should be removed here

modules/client/examples.md
josemarinas: Just adapting the naming to the subgraph, maybe not the best moment for this change , you are right :)

modules/client/src/interfaces.ts
josemarinas: I was there on the subgraph for the native token transfers

modules/client/src/interfaces.ts
josemarinas: Yes forgot to add the address, my bad

EDIT: Reviewed the code and all the token addresses are there

modules/client/src/interfaces.ts
josemarinas: This is how the formatter left it, looked weird to me too

modules/client/src/interfaces.ts
```diff
@@ -18,7 +19,7 @@ export interface IClientMethods extends IClientCore {
   pinMetadata: (params: DaoMetadata) => Promise<string>;
   /** Retrieves the asset balances of the given DAO, by default, ETH, DAI, USDC and USDT on Mainnet*/
   getDaoBalances: (
-    daoAddressOrEns: string,
+    params: AssetBalanceQueryParams,
```
brickpop: `DaoBalancesQueryParams` maybe?  for future easy discoverability

modules/client/src/interfaces.ts
brickpop: What I mean is that when you transfer an NFT... (i.e. the crypto punk number 23) you are transferring a specific thing, not a simple amount.

You are transferring the ownership, which means that `amount` is not going to work in general

modules/client/src/internal/client/methods.ts
```diff
@@ -21,13 +21,15 @@ import { AddressZero } from "@ethersproject/constants";
 import { Contract, ContractTransaction } from "@ethersproject/contracts";
 import { erc20ContractAbi } from "../abi/erc20";
 import {
+  QueryTokenBalances,
```
brickpop: Aren't we calling things like `AssetBalances` before?

brickpop: If there's a difference between subgraph and the SDK, I would suggest the SDK always talks to the exterior world in terms of "Token"

modules/client/test/integration/client/methods.test.ts
```diff
@@ -469,22 +466,26 @@ describe("Client", () => {
         const ctx = new Context(contextParamsMainnet);
         const client = new Client(ctx);
         const daoAddress = TEST_DAO_ADDRESS;
-        const balances = await client.methods.getDaoBalances(daoAddress);
+        const balances = await client.methods.getDaoBalances({
+          daoAddressOrEns: daoAddress,
```
brickpop, line 470: Do we have all of these changes on the changelog?

brickpop (commented):
Please, let's have a complete list of the breaking changes on the Changelog

modules/client/src/internal/client/methods.ts
josemarinas: Yes, internally is TokenBalances, to the exterioir is AssetBalances

modules/client/src/interfaces.ts
josemarinas: This is ERC20 typing only, ERC721 does not have amount

modules/client/src/interfaces.ts
josemarinas: makes sense

modules/client/src/internal/client/methods.ts
brickpop: Then.... why don't we unify this? What's the benefit of having both?

docs/client.md
```diff
@@ -235,7 +235,7 @@ This does not estimate the gas cost of updating the allowance of an ERC20 token<
     * [.hasPermission(params)](#ClientMethods+hasPermission) ⇒ <code>\*</code>
     * [.getDao(daoAddressOrEns)](#ClientMethods+getDao) ⇒ <code>\*</code>
     * [.getDaos({)](#ClientMethods+getDaos) ⇒ <code>\*</code>
-    * [.getDaoBalances(daoAddressorEns)](#ClientMethods+getDaoBalances) ⇒ <code>\*</code>
+    * [.getDaoBalances({)](#ClientMethods+getDaoBalances) ⇒ <code>\*</code>
```
brickpop, line 238: Don't we have this parameter now??

docs/client.md
```diff
@@ -324,15 +324,15 @@ This does not estimate the gas cost of updating the allowance of an ERC20 token<
 
 <a name="ClientMethods+getDaoBalances"></a>
 
-### clientMethods.getDaoBalances(daoAddressorEns) ⇒ <code>\*</code>
+### clientMethods.getDaoBalances({) ⇒ <code>\*</code>
```
brickpop, line 327: Same

modules/client/examples.md
```diff
@@ -2973,11 +2870,14 @@ import {
...
   ContextPlugin,
+  CreateMultisigProposalParams,
   MultisigClient,
   ProposalCreationSteps,
   ProposalMetadata,
+  TokenType,
+  WithdrawParams,
 } from "@aragon/sdk-client";
-import { CreateMultisigProposalParams, WithdrawParams } from "../../src";
+import {} from "../../src/interfaces";
```
brickpop: Needed?

modules/client/src/internal/client/methods.ts
```diff
@@ -21,16 +21,16 @@ import { AddressZero } from "@ethersproject/constants";
...
   QueryDaos,
+  QueryTokenBalances,
   QueryTokenTransfers,
 } from "../graphql-queries";
 import {
   AssetBalance,
-  AssetBalanceQueryParams,
   AssetBalanceSortBy,
   CreateDaoParams,
+  DaoBalancesQueryParams,
```
brickpop: I see, these 3 make sense together as "asset". But assets doesn't relate to the internal usage of `Token` as you said 🤔 

Is there any quick fix to make everything consistent/coherent here?

docs/client.md
josemarinas: Yes bbut it seems that this is a problem with the jsdoc2md because I noticed it happens in other places too

modules/client/src/internal/client/methods.ts
josemarinas: Without breaking what they have in the ui rn, no. If we dont wanna break the UI atm we can leave it like this for now and change to `tokenBalance`once we are in a more stable position and we refactor all the types and interfaces

docs/client.md
brickpop: but is this parameter actually gone now?  I think it isn't

If this is a problem with the tool, we close this comment and look into it later

modules/client/src/internal/client/methods.ts
brickpop: Ok, then let's keep things stable for the UI where possible :+1:

docs/client.md
josemarinas: Yes instead of `{` It should appear `DaoBalancesQueryParams`. I think It has to do with show we document the default values

brickpop (approved):

PR #157, opened 2023-02-18
F/proposalid encoding

mathewmeconry:
## Description

Task: [ID]()

## Type of change

## Checklist:


modules/client/package.json
```diff
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@aragon/core-contracts-ethers": "^0.7.1-alpha",
-    "@aragon/sdk-common": "^0.10.0-beta",
+    "@aragon/sdk-common": "^0.11.0-beta",
     "@aragon/sdk-ipfs": "^0.2.0-alpha",
```
brickpop: Could we also force a bump on sdk-ipfs to `0.3.0-beta` as well?

modules/client/examples/06-multisig-client/09-get-proposal.ts
```diff
@@ -16,15 +16,16 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
...
+  "0x1234567890123456789012345678901234567890_0x0000000000000000000000000000000000000000000000000000000000000000";
 
 const proposal: MultisigProposal | null = await client.methods.getProposal(
   proposalId,
 );
 console.log(proposal);
 /*
 {
-  id: "0x12345...",
+  id: "0x1234567890123456789012345678901234567890_0x0000000000000000000000000000000000000000000000000000000000000000",
```
brickpop: Let's follow the Subgraph format if possible, and save nonce chars if not strictly necessary. The only requirement is that nonce values can be parsed back non-ambiguously (related to potential odd-length hex values)

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -266,7 +282,7 @@ export class AddresslistVotingClientMethods extends ClientCore
    *
    * @param {string} addressOrEns
    * @return {*}  {Promise<boolean>}
-   * @memberof MultisigClientMethods
+   * @memberof AddresslistVotingClientMethods
    */
   public async canExecute(
     params: CanExecuteParams,
```
brickpop: If `CanExecuteParams` only contains `proposalId` now, we might consider removing the type and leaving just `proposalId: string` here.  What are your thoughts @mathewmeconry?

modules/client/src/addresslistVoting/internal/utils.ts
```diff
@@ -99,8 +105,9 @@ export function toAddresslistVotingProposalListItem(
     parseInt(proposal.startDate) * 1000,
   );
   const endDate = new Date(parseInt(proposal.endDate) * 1000);
+  const decodedProposalId = decodeProposalIdSubgraph(proposal.id);
   return {
-    id: proposal.id,
+    id: encodeProposalId(decodedProposalId.pluginAddress, decodedProposalId.id),
```
brickpop: Why not simply passing `id: proposal.id`? Instead of decoding and reencoding

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -81,26 +80,22 @@ export type CreateProposalBaseParams = {
...
 export interface ICanVoteParams {
-  pluginAddress: string;
-  proposalId: number;
+  proposalId: string;
   address: string;
-  vote: VoteValues
+  vote: VoteValues;
 }
 
 export type CanExecuteParams = {
```
brickpop: As said above, we can keep this type or make it redundant.
Is it likely that new params would be needed in the future?

modules/client/src/client-common/utils.ts
```diff
@@ -78,7 +78,7 @@ export function computeProposalStatusFilter(
 }
 
 export function isProposalId(propoosalId: string): boolean {
-  const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,}$/i);
+  const regex = new RegExp(/^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{64}$/i);
```
brickpop: If we keep a dynamic nonce hex size, this could remain

brickpop (commented):
Round 1

modules/client/package.json
mathewmeconry: we can if you want to

modules/client/examples/06-multisig-client/09-get-proposal.ts
mathewmeconry: changed

modules/client/src/addresslistVoting/internal/utils.ts
mathewmeconry: was necessary because of the different formats. this is now removed because of https://github.com/aragon/sdk/pull/157#discussion_r1111935932

modules/client/src/client-common/utils.ts
mathewmeconry: Improved the regex to be of max 32 bytes length

modules/client/package.json
brickpop: Yes please, let's use this to have the 3 packages in `beta` stage. The IPFS wrapper won't change at this point, most likely

modules/client/src/addresslistVoting/internal/client/methods.ts
mathewmeconry: removed

modules/client/src/client-common/interfaces/plugin.ts
mathewmeconry: removed

modules/client/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ TEMPLATE:
 - Change 1, 2, 3
 
 -->
+## UPCOMING
+### Changed
+- `proposalId` is now a `string` everywhere and follows the new general format
```
brickpop, line 18: Could we also add the mention to `canExecute()` and `execute()` now receiving the `proposalId` directly?

modules/client/examples.md
```diff
@@ -884,9 +887,8 @@ const client = new TokenVotingClient(contextPlugin);
 
 const voteParams: ICanVoteParams = {
   address: "0x1234567890123456789012345678901234567890",
```
brickpop, line 889: Is `address` the address of the person to check whether he/she can vote?

I know it's not part of the PR per se, but this feels very ambiguous here. Could we update the type in all plugins, so that it's clear what it does?

I suggest `voterAddressOrEns` here and `approverAddressOrEns` on multisig (unless there's a better name)

Types, doc, examples and tests

modules/client/examples.md
```diff
@@ -1579,9 +1587,8 @@ const client = new AddresslistVotingClient(contextPlugin);
...
```
brickpop, line 1589: Same as above :+1:

modules/client/examples.md
```diff
@@ -3056,9 +3089,8 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
 // Create an multisig client
 const client = new MultisigClient(contextPlugin);
 const canApproveParams: CanApproveParams = {
-  pluginAddress: "0x1234567890123456789012345678901234567890",
   addressOrEns: "0x1234567890123456789012345678901234567890",
```
brickpop, line 3092: Same as above, could we refactor `addressOrEns` into `approverAddressOrEns`?

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -372,6 +375,9 @@ export class MultisigClientMethods extends ClientCore
     if (!proposalId) {
       throw new InvalidProposalIdError();
     }
+    if (!/^0x[A-Fa-f0-9]{40}_(0x[A-Fa-f0-9]{1,64})$/.test(proposalId)) {
```
josemarinas: Use the common function to validate

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -328,6 +344,9 @@ export class TokenVotingClientMethods extends ClientCore
     if (!isProposalId(proposalId)) {
       throw new InvalidProposalIdError();
     }
+    if (!/^0x[A-Fa-f0-9]{40}_(0x[A-Fa-f0-9]{1,64})$/.test(proposalId)) {
```
josemarinas: Same here, use the common `isProposalId`

modules/client/examples/01-tokenVoting-client/06-can-vote.ts
```diff
@@ -18,10 +18,9 @@ const contextPlugin: ContextPlugin = ContextPlugin.fromContext(context);
 const client = new TokenVotingClient(contextPlugin);
 
 const voteParams: ICanVoteParams = {
-  address: "0x1234567890123456789012345678901234567890",
-  proposalId: "0x1234567890123456789012345678901234567890_0x1",
+  proposalId: "0x1234567890123456789012345678901234567890_0x0",
   pluginAddress: "0x1234567890123456789012345678901234567890",
```
brickpop: `pluginAddress` might be redundant after the change

modules/client/src/addresslistVoting/internal/client/estimation.ts
```diff
@@ -103,25 +116,33 @@ export class AddresslistVotingClientEstimation extends ClientCore
...
     if (!signer) {
       throw new NoSignerError();
     } else if (!signer.provider) {
       throw new NoProviderError();
     }
+
+    if (!isProposalId(proposalId)) {
+      throw new InvalidProposalIdError();
+    }
+    const { pluginAddress, id } = decodeProposalId(
```
brickpop: Could we move the `isProposalId()` checks inside of `decodeProposalId()`?

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -136,15 +136,14 @@ export class AddresslistVotingClientMethods extends ClientCore
...
+    const proposalId = parsedLog.args["proposalId"];
     if (!proposalId) {
       throw new ProposalCreationError();
     }
 
     yield {
       key: ProposalCreationSteps.DONE,
-      // TODO remove this when new proposal format
-      proposalId: proposalId.toNumber(),
+      proposalId: encodeProposalId(params.pluginAddress, Number(proposalId)),
```
brickpop, line 146: will `proposalId` be a base 10 number encoded in a string? I would make the numeric base explicit

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -180,14 +179,18 @@ export class AddresslistVotingClientMethods extends ClientCore
     } else if (!signer.provider) {
       throw new NoProviderError();
     }
+    if (!isProposalId(params.proposalId)) {
```
brickpop: Same

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -221,11 +224,16 @@ export class AddresslistVotingClientMethods extends ClientCore
       throw new NoProviderError();
     }
 
+    if (!isProposalId(proposalId)) {
```
brickpop: (same, etc)

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -328,6 +341,9 @@ export class AddresslistVotingClientMethods extends ClientCore
     if (!proposalId) {
       throw new InvalidProposalIdError();
     }
+    if (!/^0x[A-Fa-f0-9]{40}_(0x[A-Fa-f0-9]{1,64})$/.test(proposalId)) {
+      throw new InvalidProposalIdError();
+    }
```
brickpop: Using `isProposalId()` instead of these 2 checks may be better

modules/client/src/client-common/interfaces/plugin.ts
```diff
@@ -81,28 +80,16 @@ export type CreateProposalBaseParams = {
...
-  pluginAddress: string;
-  proposalId: number;
+  proposalId: string;
 }
 
 export interface ICanVoteParams {
-  pluginAddress: string;
-  proposalId: number;
+  proposalId: string;
   address: string;
```
brickpop, line 89: (see the above comment)

brickpop (commented):
Round 2

modules/client/package.json
mathewmeconry: done

modules/client/src/multisig/internal/client/methods.ts
mathewmeconry: the common function doesn't validate but also returns stuff. Idk if this is the expected way how to test it. @brickpop what shall I do?

modules/client/CHANGELOG.md
mathewmeconry: done

modules/client/examples.md
mathewmeconry: I have no clue. @josemarinas you changed these lines the last time so please answer the question

modules/client/examples.md
mathewmeconry: same @josemarinas

modules/client/examples.md
mathewmeconry: same @josemarinas

modules/client/examples/01-tokenVoting-client/06-can-vote.ts
mathewmeconry: this was odd. changed it to match the interface

modules/client/src/addresslistVoting/internal/client/estimation.ts
mathewmeconry: decodeProposalId() checks for validity. no need to check it before decoding

modules/client/src/client-common/interfaces/plugin.ts
mathewmeconry: same @josemarinas

modules/client/src/addresslistVoting/internal/client/estimation.ts
mathewmeconry: `decodeProposalId()` checks for validity. no need to check it before decoding

modules/client/src/multisig/internal/client/methods.ts
brickpop: @mathewmeconry `isProposalId()` should be returning `true/false`, so encapsulating the test there sounds simpler than repeating the code in some places.

`decodeProposalId()` returns the values, but I don't think we need this here.

modules/client/src/addresslistVoting/internal/client/estimation.ts
brickpop: If `decodeProposalId()` already checks for validity, then I would simply remove the `isProposalId()` check before the `decodeProposalId()` call

modules/client/examples.md
josemarinas: Yes this is the wallet address that you want to check if hast the permission to vote. This means the `who`so yes maybe is not the most straightforward naming and for sure it should be changed at least to `addressOrEns`, the change to `voterXXX` or `approverXXX` wouñlld make sense because in other parts of the code we have a reference of whats the address for, for example `daoAddressOrEns` or `pluginAddress`.

So yeah, I'm happy with the change, let me know if you wanna include it in this PR. If not create a task to review all the taming follows this pattern 👍 @brickpop

modules/client/examples/01-tokenVoting-client/06-can-vote.ts
brickpop: Probably because it's an `interface` (ensure a minimum set of fields/methods) and not a `type` (ensure an exact matching). There's a ticket to turn all remaining data-only interfaces into types (when reasonable), I guess that's why TS didn't complain.

modules/common/src/encoding.ts
```diff
@@ -102,6 +102,35 @@ export function decodeRatio(
...
+  return `${pluginAddress}_0x${id.toString(16)}`;
+}
+
+/** Decodes a proposalId from subgraph and returns the original pluginAddress and the nonce */
+export function decodeProposalId(
+  proposalId: string,
+): { pluginAddress: string; id: number } {
+  if (!/^0x[A-Fa-f0-9]{40}_(0x[A-Fa-f0-9]{1,64})$/.test(proposalId)) {
+    throw new Error("Invalid proposalId");
+  }
```
brickpop: Can we call `isProposalId()` and throw `InvalidProposalIdError` when applicable?

modules/common/src/encoding.ts
```diff
@@ -102,6 +102,35 @@ export function decodeRatio(
...
+
+  const matchedRegexResult =
+    proposalId.match(/^(0x[A-Fa-f0-9]{40})_(0x[A-Fa-f0-9]{1,64})$/) || [];
+  if (matchedRegexResult.length !== 3) {
+    throw new Error("Failed to deconstruct proposalId");
```
brickpop:
```suggestion
    throw new Error("Could not parse the proposal ID");
```

modules/client/src/addresslistVoting/internal/client/methods.ts
brickpop: In other words, if `proposalId` is a number or a bigint, things will work fine. But if the value is `0x55ff` then, `Number()` will fail

modules/client/test/integration/multisig-client/methods.test.ts
```diff
@@ -124,12 +124,11 @@ describe("Client Multisig", () => {
           expect(step.txHash).toMatch(/^0x[A-Fa-f0-9]{64}$/i);
           break;
         case ProposalCreationSteps.DONE:
-          expect(typeof step.proposalId).toBe("number");
+          expect(typeof step.proposalId).toBe("string");
+          expect(step.proposalId).toMatch(
+            /^0x[A-Fa-f0-9]{40}_0x[A-Fa-f0-9]{1,64}$/,
+          );
           return step.proposalId;
-          // TODO
```
brickpop, line 129: I know this is not ideal, but could we leave this comment here, to make things easier to find when we finally merge the metadata validation PR? 🙏

brickpop (commented):
Round 3

modules/client/examples.md
brickpop: @josemarinas ok, then let's have this as a separate task not to make things bigger :+1:

modules/client/src/multisig/internal/client/methods.ts
mathewmeconry: I removed the check because it was duplicate

modules/client/src/tokenVoting/internal/client/methods.ts
mathewmeconry: same

modules/client/src/addresslistVoting/internal/client/estimation.ts
mathewmeconry: done

modules/client/src/addresslistVoting/internal/client/methods.ts
mathewmeconry: the value will be a number but ethers returns it as `BigNumberish`. Meaning that the ts typing system would complain if we just use it directly. Thus the usage of `Number(proposalId)`.
If you want I can add a `parseInt()` just to do it another way

modules/client/src/addresslistVoting/internal/client/methods.ts
mathewmeconry: done

modules/common/src/encoding.ts
mathewmeconry: That doesn't follow the other functions that are implemented in `@aragon/sdk-common`. The common package defines custom errors but all it's functions throws "normal" errors with a message.
For example in `modules/common/src/encoding.ts` lines 1 - 10:
```typescript
/** Decodes a hex string and returns it as a buffer */
export function hexToBytes(hexString: string): Uint8Array {
  if (!hexString) return new Uint8Array();
  else if (!/^(0x)?[0-9a-fA-F]*$/.test(hexString)) {
    throw new Error("Invalid hex string");
  } else if (hexString.length % 2 !== 0) {
    throw new Error("The hex string has an odd length");
  }
```

modules/common/src/encoding.ts
mathewmeconry: done

modules/client/test/integration/multisig-client/methods.test.ts
mathewmeconry: This todo is done so I'd rather remove it and have clean code

modules/client/src/addresslistVoting/internal/client/estimation.ts
```diff
@@ -87,13 +93,20 @@ export class AddresslistVotingClientEstimation extends ClientCore
       throw new NoProviderError();
     }
 
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
+    }
```
brickpop: Redundant if decodeProposalId already does it

modules/client/src/addresslistVoting/internal/client/methods.ts
brickpop: Redundant

modules/client/src/multisig/internal/client/estimation.ts
```diff
@@ -90,50 +90,52 @@ export class MultisigClientEstimation extends ClientCore
     } else if (!signer.provider) {
       throw new NoProviderError();
     }
-    if (!isAddress(params.pluginAddress)) {
-      throw new InvalidAddressError();
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
     }
```
brickpop: Redundant

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -172,18 +171,19 @@ export class MultisigClientMethods extends ClientCore
...
       throw new NoProviderError();
     }
-    // TODO
-    // use yup
-    if (!isAddress(params.pluginAddress)) {
-      throw new InvalidAddressError();
+
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
     }
```
brickpop: Redundant?

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -256,48 +254,47 @@ export class MultisigClientMethods extends ClientCore
...
-    // use yup
     if (!isAddress(params.addressOrEns)) {
       throw new InvalidAddressOrEnsError();
     }
-    if (!isAddress(params.pluginAddress)) {
-      throw new InvalidAddressOrEnsError();
+
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
     }
```
brickpop: Redundant?

modules/client/src/multisig/internal/client/methods.ts
brickpop: I still see the two if's. I suggest we replace the "if" blocks by this one:

```ts
if (!isProposalId(proposalId)) { ... }
```

modules/client/src/tokenVoting/internal/client/estimation.ts
```diff
@@ -83,13 +89,20 @@ export class TokenVotingClientEstimation extends ClientCore
     } else if (!signer.provider) {
       throw new NoProviderError();
     }
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
```
brickpop: Redundant

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -184,13 +182,18 @@ export class TokenVotingClientMethods extends ClientCore
       throw new NoProviderError();
     }
 
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
+    }
```
brickpop: Redundant

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -247,46 +252,51 @@ export class TokenVotingClientMethods extends ClientCore
...
     if (!signer.provider) {
       throw new NoProviderError();
-    } else if (!isAddress(params.address) || !isAddress(params.pluginAddress)) {
+    } else if (!isAddress(params.address)) {
       throw new InvalidAddressError();
     }
 
+    if (!isProposalId(params.proposalId)) {
+      throw new InvalidProposalIdError();
+    }
```
brickpop: Redundant

modules/common/src/encoding.ts
brickpop: That's a good point, we have a low priority task for refactoring these generic errors into specific ones. 

At the very least, we should be calling `isProposalId()` instead of copying the regex every time

modules/client/src/multisig/internal/client/methods.ts
mathewmeconry: missed this one. it is fixed now

modules/client/src/addresslistVoting/internal/client/methods.ts
mathewmeconry: removed

modules/common/src/encoding.ts
mathewmeconry: `isProposalId()` is part of the common client in client and not part of the common package.

modules/client/src/addresslistVoting/internal/client/estimation.ts
mathewmeconry: removed

modules/client/src/multisig/internal/client/estimation.ts
mathewmeconry: removed

modules/client/src/multisig/internal/client/methods.ts
mathewmeconry: removed

modules/client/src/multisig/internal/client/methods.ts
mathewmeconry: removed

modules/client/src/tokenVoting/internal/client/estimation.ts
mathewmeconry: removed

modules/client/src/tokenVoting/internal/client/methods.ts
mathewmeconry: removed

modules/client/src/tokenVoting/internal/client/methods.ts
mathewmeconry: removed

modules/common/src/encoding.ts
brickpop: Then I would move the function and any related testing here instead. 

There should be only one source of truth, regarding valid/invalid proposal ID's, and the existing logic shouldn't be impacting any future plugins using hashing instead of nonce

modules/common/CHANGELOG.md
```diff
@@ -17,8 +17,18 @@ TEMPLATE:
 -->
 
 ## [UPCOMING]
+### Adds
```
brickpop, line 21:
```suggestion
## 0.11.0-beta
### Adds
```

modules/common/CHANGELOG.md
```diff
@@ -17,8 +17,18 @@ TEMPLATE:
...
+- Adds `isProposalId()` function to the package
+
+### Changed
+- Changes `encodeProposalId` and `decodeProposalId` function to support the new format
+
+## 0.11.0-beta
+- Adds `encodeProposalId(addr, nonce)` and `decodeProposalId(pid)`
```
brickpop: Move up, where `UPCOMING` is now. Versions are no longer automatic, right?

modules/common/src/encoding.ts
```diff
@@ -102,6 +102,35 @@ export function decodeRatio(
...
+  }
+
+  return `${pluginAddress}_0x${id.toString(16)}`;
+}
+
+/** Decodes a proposalId from subgraph and returns the original pluginAddress and the nonce */
+export function decodeProposalId(
+  proposalId: string,
+): { pluginAddress: string; id: number } {
+  if (!/^0x[A-Fa-f0-9]{40}_(0x[A-Fa-f0-9]{1,64})$/.test(proposalId)) {
```
brickpop: Using isProposalId() here now

Not super important

brickpop (approved):

modules/common/src/encoding.ts
```diff
@@ -102,6 +102,35 @@ export function decodeRatio(
   return Number(onChainValue) / (10 ** digits);
 }
 
+/** Encodes the particles of a proposalId into a globally unque value for subgraph */
+export function encodeProposalId(pluginAddress: string, id: number) {
+  if (!/^0x[A-Fa-f0-9]{40}$/.test(pluginAddress)) {
```
josemarinas, line 109: Just a small comment, we have an `isAddress` function and a `InvalidAddressError` on the common package so we can use them here

modules/common/src/encoding.ts
```diff
@@ -102,6 +102,35 @@ export function decodeRatio(
...
+    throw new Error("Invalid address");
+  }
+
+  return `${pluginAddress}_0x${id.toString(16)}`;
+}
+
+/** Decodes a proposalId from subgraph and returns the original pluginAddress and the nonce */
+export function decodeProposalId(
+  proposalId: string,
+): { pluginAddress: string; id: number } {
```
josemarinas, line 119: same here

modules/common/src/encoding.ts
mathewmeconry: `isAddress` is not in the common package but in the common-client. so it is not usable here. Also the custom error: I followed just how the other functions are working in the common package and all of them use "generic" errors

modules/common/src/encoding.ts
mathewmeconry: same

modules/common/src/encoding.ts
mathewmeconry: done

modules/common/CHANGELOG.md
mathewmeconry: moved

modules/common/CHANGELOG.md
mathewmeconry: done

modules/common/src/encoding.ts
mathewmeconry: done

modules/common/src/encoding.ts
brickpop: I think isAddress is an ethers.js exported function, you could pull it from there but not super important again

brickpop (approved):

modules/common/src/encoding.ts
josemarinas: okay 👍

PR #178, opened 2023-03-03
SDK Examples docs generation

heueristik:
## Description

The generated markdown was put into one single file and the file was not pushed to the builders-portal.

This PR 
- generates a separate markdown for each example 
- improves the workflow of updating the builders-portal by
  - copying the individual files while maintaining the folder structure using `rsync`
  - updating the GitHub workflow file
  - git-ignoring auto-generated files for the SDK versioning

Task: [APP-1967](https://aragonassociation.atlassian.net/browse/APP-1967), [APP-2012](https://aragonassociation.atlassian.net/browse/APP-2012)

[APP-1967]: https://aragonassociation.atlassian.net/browse/APP-1967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-2012]: https://aragonassociation.atlassian.net/browse/APP-2012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/documentation-update.yml
```diff
@@ -22,53 +22,40 @@ jobs:
...
-          git pull
-          git config --global user.name "Arabot-1"
-          git config --global user.email "arabot-1@users.noreply.github.com"
-          git add -A ./docs/
-          if [ $(git diff-index HEAD ./docs | wc -l) -gt 0 ]; then
-            git commit -am "Updates docs from commit $SHA_SHORT"
-            git push
-          fi
-        env:
-          SHA_SHORT: ${{ steps.hash.outputs.sha_short }}
```
brickpop, line 49: Why removing the step that commits the updated docs? Are they pushed to the DP instead of the repo itself?

Does this affect the examples somehow? Or just the deprecated `docs/` folder?

.github/workflows/documentation-update.yml
```diff
@@ -22,53 +22,40 @@ jobs:
...
           repository: aragon/builders-portal
           ref: staging
           path: builders-portal
           token: ${{ secrets.ARABOT_PAT }}
-      - name: Copy docs (Javascript)
-        run: |
-          mkdir -p $GITHUB_WORKSPACE/builders-portal/docs/sdk/javascript/
-          rm -rf $GITHUB_WORKSPACE/builders-portal/docs/sdk/javascript/*
-          cp -R ./docs/* $GITHUB_WORKSPACE/builders-portal/docs/sdk/javascript/
+      - name: Remove the docs/sdk folder in the builders-portal
```
brickpop:
```suggestion
      - name: Clean the docs/sdk folder in the builders-portal
```

modules/client/scripts/generate-markdown.js
```diff
@@ -0,0 +1,89 @@
+const fs = require("fs");
```
brickpop, line 1: The new file name sounds a bit ambiguous

modules/client/scripts/generate-markdown.js
```diff
@@ -0,0 +1,89 @@
...
+
+generateMarkdown().catch((err) => {
+  console.error(err);
+  process.exit(1);
+});
+
+async function generateMarkdown() {
+  const workingDirectory = process.argv[2];
+  if (!workingDirectory) {
+    console.error("Usage:  node generate-examples-md.js <working-directory>");
```
brickpop: Outdated script name

brickpop (commented):
If we are removing the examples.md file, the readme should be updated to point to the new location. Otherwise, we are losing effective contents

.github/workflows/documentation-update.yml
heueristik: > Are they pushed to the DP instead of the repo itself?

Yes, there is no need to store auto-generated files twice.

We use the same flow as for `aragon/osx` https://github.com/aragon/osx/blob/develop/.github/workflows/documentation-update.yml

modules/client/scripts/generate-markdown.js
heueristik: Fixed in 5154a78023ef8621c701a035fbfc81ba82db4bc8

modules/client/scripts/generate-markdown.js
heueristik: Hm, why? It can generate markdown files from any file with `/*MARKDOWN */` tokens inside, not only from examples.

.github/workflows/documentation-update.yml
heueristik: The workflow was copied over from `aragon/osx`.
As the bash command we apply is `rm`, I would like to keep it as is.

brickpop (approved):

brickpop (approved):

PR #181, opened 2023-03-07
Feature: Optimize subgraph calls

josemarinas:
## Description

Task: [APP-1932](https://aragonassociation.atlassian.net/browse/APP-1932)

## Type of change

## Checklist:

[APP-1932]: https://aragonassociation.atlassian.net/browse/APP-1932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -307,16 +307,18 @@ export class AddresslistVotingClientMethods extends ClientCore
     }
 
     try {
-      await this.graphql.ensureOnline();
+      this.graphql.assertClient();
       const client = this.graphql.getClient();
```
brickpop: Would there be a downside if we moved `assertClient()` into `getClient()` ?

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -307,16 +307,18 @@ export class AddresslistVotingClientMethods extends ClientCore
...
       });
       return response.addresslistVotingPlugin.members.map((
         member: { address: string },
       ) => member.address);
-    } catch {
-      throw new GraphQLError("AddressList members");
+    } catch (e) {
+      handleGraphQLError(e as Error, "adddresslistVoting members");
+      await this.graphql.ensureOnline();
+      return this.getMembers(pluginAddress);
```
brickpop: Careful because this can create an infinite recursion loop and also leak memory :warning:

brickpop: Isn't is possible to move the logic into the `request()` primitive and try new clients in loop until one works or until we have tried them all?

modules/client/src/client-common/utils.ts
```diff
@@ -117,3 +119,15 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+
+export function handleGraphQLError(error: Error, message: string) {
+  if (error instanceof ClientNotInitializedError) {
+    throw error;
+  } else if (error instanceof GraphQLClientError) {
+    const e = error as GraphQLClientError;
+    if (e.response.status < 500) {
+      throw new GraphQLError(message);
+    } 
+  }
```
brickpop: If the error is of another type, we do nothing?

modules/client/test/integration/constants.ts
```diff
@@ -37,6 +37,18 @@ const ipfsEndpoints = {
...
+        "https://httpstat.us/504?sleep=500",
+    },
+    {
+      url:
+        "https://httpstat.us/504?sleep=500",
+    },
+    {
+      url:
+        "https://httpstat.us/504?sleep=500",
+    },
```
brickpop: Are these "working" endpoints?

modules/client/src/client-common/utils.ts
josemarinas: no, because it means we have a 5xx meaning there is a server error and we want to change the endpoint

modules/client/src/addresslistVoting/internal/client/methods.ts
josemarinas: yes no problem with that 👍

modules/client/src/client-common/utils.ts
brickpop: ok, ok... but then they should be on `failing` and the test should be selecting the `failing` ones. Don't you think?

brickpop: oh ok :+1:
Adding some comment to clarify this for future maintenability would be nide

modules/common/src/timeout.ts
```diff
@@ -46,9 +46,21 @@ export function promiseWithTimeout<T>(
...
+
+export async function functionRetryWithCondition<T>(
+  func: () => Promise<T>,
+  // should not throw to execute recursivity
+  retryCondition: (e: Error) => Promise<void>,
+): Promise<T> {
+  return func().catch((e) => {
+    return retryCondition(e).then(() =>
+      functionRetryWithCondition(func, retryCondition)
+    );
```
brickpop: Wouldn't it be easier to have this?

```ts
async function tryUntil({
  func,
  onFail,
  shouldContinue
}) {
  do {
    try {
      const result = await func();
      // it worked
      return result;
    }
    catch (err) {
      onFail();
    }
  } while(shouldContinue())
}

// Usage:

await tryUntil({
  func: async () => {
    const result = await this.graphql.request(...);
    // ...
    return result;
  },
  onFail: () => {
    retries--;
    this.graphql.shiftClient()
  },
  shouldContinue: () => {
    return retries > 0;
  }
});
```

But actually, this logic could be abstracted inside of `request` and not need to be copied in every single method, every time.

modules/client/test/integration/constants.ts
josemarinas: Nope, these are endpoints simulating a 504 error, to test that the client shift is actually happening

modules/client/src/client-common/utils.ts
josemarinas: done, so resolving

modules/client/package.json
```diff
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@aragon/osx-ethers": "^1.0.0",
-    "@aragon/sdk-common": "^1.0.0",
+    "@aragon/sdk-common": "^1.0.1",
```
brickpop:
```suggestion
    "@aragon/sdk-common": "^1.1.0",
```

modules/common/src/promises.ts
```diff
@@ -16,3 +16,20 @@ function reflect<T>(prom: Promise<T>) {
...
+  shouldContinue: () => boolean;
+}) {
+  do {
+    try {
+      const result = await func();
+      // it worked
+      return result;
+    } catch (err) {
+      onFail(err as Error);
+      continue
```
brickpop: `continue` is it needed?

modules/common/src/promises.ts
```diff
@@ -16,3 +16,20 @@ function reflect<T>(prom: Promise<T>) {
     .then((value: T) => ({ value, status: "fulfilled" }))
     .catch((reason: T) => ({ reason, status: "rejected" }));
 }
+
+export async function tryUntil<T>({ func, onFail, shouldContinue }: {
```
brickpop: Still thinking of a better name here... maybe `runWithRetries`, `tryCallback`, `retryFailable`, ...

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -72,4 +74,33 @@ export class GraphqlModule implements IClientGraphQLCore {
     }
     throw new NoNodesAvailableError("graphql");
   }
+
+  public async request({ query, params, name }: {
+    query: string;
+    params: { [key: string]: any };
+    name?: string;
+  }) {
+    let retries = this.clients.length;
```
brickpop, line 86: I would throw an error right away if the length is `0`

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -72,4 +74,33 @@ export class GraphqlModule implements IClientGraphQLCore {
...
+    return tryUntil({
+      func: async () => {
+        const client = this.getClient();
+        return await client.request(query, params);
+      },
```
brickpop:
```suggestion
      func: () => {
        return this.getClient()
          .then(client => client.request(query, params));
      },
```

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -72,4 +74,33 @@ export class GraphqlModule implements IClientGraphQLCore {
...
+      onFail: (e: Error) => {
+        if (e instanceof ClientError) {
+          const err = e as ClientError;
```
brickpop: Needed?

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -72,4 +74,33 @@ export class GraphqlModule implements IClientGraphQLCore {
...
+          if (err.response.status < 500) {
+            throw new GraphQLError(name || "");
+          }
+        }
+        retries--;
+        this.shiftClient();
+      },
+      shouldContinue: () => {
+        return retries > 0;
+      },
```
brickpop:
```suggestion
      shouldContinue: () => retries > 0,
```

modules/common/src/promises.ts
josemarinas: nope, just me being verbose

modules/client/src/client-common/modules/graphql.ts
josemarinas: I already do this check on the getClient wich is called two lines later

brickpop: Right, but for this, we have to call the function, enter the async realm, wait for the promise to resolve, etc.

It's much cleaner to fail on the sync side

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -76,14 +66,40 @@ export class GraphqlModule implements IClientGraphQLCore {
...
     for (let i = 0; i < this.clients.length; i++) {
       if (await this.isUp()) return;
-
       this.shiftClient();
     }
-    throw new Error("No graphql nodes available");
+    throw new NoNodesAvailableError("graphql");
+  }
+
+  public async request({ query, params, name }: {
```
brickpop: Async is not needed with only one `return`

```suggestion
  public request({ query, params, name }: {
```

modules/client/src/client-common/modules/graphql.ts
```diff
@@ -76,14 +66,40 @@ export class GraphqlModule implements IClientGraphQLCore {
...
+          }
+        }
+        retries--;
+        this.shiftClient();
+      },
+      shouldRetry: () => {
+        if (retries <= 0) {
+          throw new NoNodesAvailableError("graphql");
+        }
+        return true;
```
brickpop:
```suggestion
        return retries > 0;
```

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -333,41 +332,39 @@ export class TokenVotingClientMethods extends ClientCore
...
-          );
-        }
+      },
+      name: "TokenVoting proposal",
+    });
+    if (!tokenVotingProposal) {
+      return null;
+    }
+    // format in the metadata field
+    try {
```
brickpop, line 359: Why does the `try` block only cover half of the function code that could fail?

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -400,55 +397,54 @@ export class TokenVotingClientMethods extends ClientCore
...
+            const metadataCid = resolveIpfsCid(proposal.metadata);
+            const stringMetadata = await this.ipfs.fetchString(metadataCid);
+            const metadata = JSON.parse(stringMetadata) as ProposalMetadata;
+            return toTokenVotingProposalListItem(proposal, metadata);
+          } catch (err) {
+            if (err instanceof InvalidCidError) {
               return toTokenVotingProposalListItem(
                 proposal,
-                UNAVAILABLE_PROPOSAL_METADATA,
+                UNSUPPORTED_PROPOSAL_METADATA_LINK,
```
brickpop, line 450: Is it easy to detect when a DAO simply has no metadata defined at all?

modules/client/src/tokenVoting/internal/client/methods.ts
```diff
@@ -464,38 +460,35 @@ export class TokenVotingClientMethods extends ClientCore
...
-        {
-          address: pluginAddress,
+    const { tokenVotingPlugin }: {
+      tokenVotingPlugin: SubgraphVotingSettings;
+    } = await this.graphql.request(
+      {
+        query: QueryTokenVotingSettings,
+        params: {
+          address: pluginAddress.toLowerCase(),
         },
```
brickpop: Really cosmetic, not important. But calls could be cleaner if they looked like:

```ts
const query = QueryTokenVotingSettings;
const params: {address: pluginAddress.toLowerCase()};
const name = "TokenVoting settings";

type T = { tokenVotingPlugin: SubgraphVotingSettings }
const { tokenVotingPlugin } = await this.graphql.request<T>({ query, params, name });



```

modules/client/test/integration/constants.ts
```diff
@@ -37,6 +37,18 @@ const ipfsEndpoints = {
 
 const grapqhlEndpoints = {
   working: [
```
brickpop: These don't seem to be `working`. If a test expects them to fail, the test should pick `failing` URI's

modules/client/package.json
```diff
@@ -61,7 +61,7 @@
   },
```
brickpop, line 61: Should we bump this version too? Or waiting for other PR too?

modules/common/src/promises.ts
```diff
@@ -16,3 +16,19 @@ function reflect<T>(prom: Promise<T>) {
...
+  do {
+    try {
+      const result = await func();
+      // it worked
+      return result;
+    } catch (err) {
+      onFail(err as Error);
+    }
+  } while (shouldRetry());
+}
```
brickpop: If `shouldRetry()` is false but the last result is an error, the function should be throwing the last known error.

modules/common/src/promises.ts
```diff
@@ -16,3 +16,19 @@ function reflect<T>(prom: Promise<T>) {
     .then((value: T) => ({ value, status: "fulfilled" }))
     .catch((reason: T) => ({ reason, status: "rejected" }));
 }
+
+export async function runAndRetry<T>({ func, onFail, shouldRetry }: {
```
brickpop, line 20: This function doesn't seem to have any new test on the common package. 

Something like this may do the trick:

```ts
let countFunc = 0;
let countOnFail = 0;

await runAndRetry({
  func: async () => {
    if(countFunc < 3) countFunc++;
    else throw new Error();
  },
  onFail: () => {
    countOnFail++;
  },
  shouldRetry: () => {
    return true;
  }
});

expect(countFunc).toBe(3);
expect(countOnFail).toBe(3);

// --------

expect(async () => {
  await runAndRetry({
    func: () => { throw new Error() },
    onFail: () => {},
    shouldRetry: () => false
  });
}).toThrow()



```

modules/common/src/promises.ts
```diff
@@ -16,3 +16,19 @@ function reflect<T>(prom: Promise<T>) {
...
+  onFail: (e: Error) => void;
+  shouldRetry: () => boolean;
+}) {
+  do {
+    try {
+      const result = await func();
+      // it worked
+      return result;
+    } catch (err) {
+      onFail(err as Error);
```
brickpop:
```suggestion
      if (typeof onFail === "function) {
        onFail(err as Error);
      }
```

modules/common/src/promises.ts
```diff
@@ -16,3 +16,19 @@ function reflect<T>(prom: Promise<T>) {
     .then((value: T) => ({ value, status: "fulfilled" }))
     .catch((reason: T) => ({ reason, status: "rejected" }));
 }
+
+export async function runAndRetry<T>({ func, onFail, shouldRetry }: {
+  func: () => Promise<T>;
+  onFail: (e: Error) => void;
```
brickpop:
```suggestion
  onFail?: (e: Error) => void;
```

modules/client/src/client-common/modules/graphql.ts
josemarinas: If we do it like this, the sdk will throw an error that will be like `cannot acces property XXX of undefined` which can be misleading

modules/client/src/tokenVoting/internal/client/methods.ts
josemarinas: Because it handles the ipfs errors, this will eventually be handled like the subgraph ones are handled on this PR

modules/client/src/tokenVoting/internal/client/methods.ts
josemarinas: Yes, we detect when there is no metadata, or at least a metadata that we can downloada and when we have metadata but it does not follow our schema

modules/client/test/integration/constants.ts
josemarinas: I dont expect them to fail, I expect it to iterate over the failing ones until it arrives to the working one. We can change this if you want

modules/common/src/promises.ts
josemarinas: Yes, but I wanted the shouldRetry function to thow the error so we dont lose the specific error for a generic one

modules/common/src/promises.ts
josemarinas: 👍

modules/client/src/internal/constants.ts
```diff
@@ -22,6 +22,11 @@ export const UNSUPPORTED_DAO_METADATA_LINK: DaoMetadata = {
   description: "(the metadata link is not supported)",
   links: [],
 };
+export const EMPTY_DAO_METADATA_LINK: DaoMetadata = {
+  name: "(the DAO has no metadata)",
+  description: "(the current DAO did not define any content)",
```
brickpop:
```suggestion
  description: "(the DAO did not define any content)",
```

modules/client/test/integration/addresslistVoting-client/methods.test.ts
```diff
@@ -513,7 +513,7 @@ describe("Client Address List", () => {
 
   describe("Data retrieval", () => {
     it("Should get the list of members that can vote in a proposal", async () => {
-      const ctx = new Context(contextParamsMainnet);
+      const ctx = new Context(contextParamsGraphqlWithTimeouts);
```
brickpop: In general

```suggestion
      const ctx = new Context(contextParamsOkWithGraphqlTimeouts);
```

modules/client/test/integration/constants.ts
```diff
@@ -106,3 +117,12 @@ export const contextParamsFailing: ContextParams = {
...
   graphqlNodes: grapqhlEndpoints.failing,
 };
+
+export const contextParamsGraphqlWithTimeouts: ContextParams = {
+  network: "mainnet",
+  signer: new Wallet(TEST_WALLET),
+  daoFactoryAddress: "0x0123456789012345678901234567890123456789",
+  web3Providers: web3endpoints.working,
+  ipfsNodes: ipfsEndpoints.working,
+  graphqlNodes: [...grapqhlEndpoints.timeout, ...grapqhlEndpoints.working],
```
brickpop:
```suggestion
  graphqlNodes: grapqhlEndpoints.timeout.concat(grapqhlEndpoints.working),
```

brickpop (approved):

PR #183, opened 2023-03-07
App 1643 sdk create a subgraph mock for isolated testing

mathewmeconry:
## Description

Task: [ID]()

## Type of change

## Checklist:


modules/client/src/client-common/interfaces/common.ts
```diff
@@ -1,5 +1,8 @@
-export const SupportedNetworksArray = ["mainnet", "goerli"] as const;
-export type SupportedNetworks = typeof SupportedNetworksArray[number];
+export type SupportedNetworks = "mainnet" | "goerli";
+export const SupportedNetworksArray: SupportedNetworks[] = [
+  "mainnet",
+  "goerli",
+];
```
brickpop: I would leave it as it was. In your way, you have to enter the values two times, whereas in the old one TS builds the type automatically.

modules/client/test/integration/addresslistVoting-client/methods.test.ts
```diff
@@ -1,19 +1,17 @@
-// @ts-ignore
-declare const describe, it, beforeAll, afterAll, expect;
```
brickpop, line 2: Wouldn't this leave the editor full of warnings and errors without any typing?

mathewmeconry: Only if your IDE isn't setup correctly. Also using `// @ts-ignore` has to be avoid as much as possible

modules/client/src/client-common/interfaces/common.ts
mathewmeconry: done

modules/client/test/integration/addresslistVoting-client/methods.test.ts
```diff
@@ -728,19 +843,27 @@ describe("Client Address List", () => {
...
+        addresslistVotingPlugin: {
+          minDuration: "10",
+          minProposerVotingPower: "200000",
+          minParticipation: "300000",
+          supportThreshold: "400000",
+          votingMode: VotingMode.STANDARD,
+        },
+      });
+
       const settings = await client.methods.getVotingSettings(pluginAddress);
```
josemarinas, line 940: Why not using the typecast to override the null value here?

modules/client/test/integration/constants.ts
```diff
@@ -45,6 +51,10 @@ const grapqhlEndpoints = {
   failing: [{ url: "https://bad-url-gateway.io/" }],
 };
 
+export const ADDRESS_ONE = "0x0000000000000000000000000000000000000001";
+export const ADDRESS_TWO = "0x0000000000000000000000000000000000000002";
+export const ADDRESS_THREE = "0x0000000000000000000000000000000000000003";
+
 export const TEST_WALLET_ADDRESS = "0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199";
 export const TEST_WALLET =
   "0xdf57089febbacf7ba0bc227dafbffa9fc08a93fdc68e1e42411a14efcf23656e";
```
josemarinas, line 71: The `TEST_TOKEN_VOTING_DAO_ADDRESS`, `TEST_TOKEN_VOTING_PLUGIN_ADDRESS`... constants dont make sense anymore so why wee dont delete them?

modules/client/test/integration/client/methods.test.ts
```diff
@@ -366,41 +384,55 @@ describe("Client", () => {
...
+          plugins: [{
+            id: TEST_MULTISIG_PLUGIN_ADDRESS,
+            __typename: SubgraphPluginTypeName.MULTISIG,
+          }],
+        };
+        mockedClient.request.mockResolvedValueOnce({
+          dao: subgraphDao,
+        });
+
         const dao = await client.methods.getDao(daoAddress);
```
josemarinas, line 402: Same here, why not using the type cast?
Personally I prefer the typecast to make it not null because I only have to write it once but if you prefer `!` go for it, but lets try to do it the same way on all places

modules/client/test/integration/addresslistVoting-client/methods.test.ts
mathewmeconry: because typecasting should be limited to a minimum and here it is not necessary

modules/client/test/integration/constants.ts
mathewmeconry: because even they don't fulfill a function anymore and can be any value that is a valid address. the tests still use them to issue the queries

modules/client/test/integration/client/methods.test.ts
mathewmeconry: typecasting should be avoided as much as possible. and here it is possible to avoid so I don't do it

modules/client/test/integration/constants.ts
josemarinas: Well the full purpose of this constants was to point to daoAddresses, pluginAddresses... that existend on the version of subgraph used, since the whole subgraph is now mocked, i downt se why we still need it.

Maybe i'm missing something but if we change all appearences by ADDRESS_ONE we should be good anyways

mathewmeconry: that is right. Let me change it

modules/client/test/integration/client/methods.test.ts
josemarinas: I mean both are assertions and should be avoided when possible imo, here it does not mater much because we can assume that the mock is going to return what we tell him to return. So feel free to leave it this way or change it to a `null` check

brickpop (approved):

josemarinas (approved):

PR #199, opened 2023-04-05
Feature: Provide plugin version

josemarinas:
## Description

Provide plugin versioning to daos obtained from subgraph

Task: [OS-215](https://aragonassociation.atlassian.net/browse/OS-215)

## Type of change

## Checklist:

[OS-215]: https://aragonassociation.atlassian.net/browse/OS-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/src/interfaces.ts
+++ b/modules/client/src/interfaces.ts
@@ -440,28 +440,15 @@ export enum DaoSortBy {
   // POPULARITY = "totalProposals", // currently defined as number of proposals
 }
 
-export enum SubgraphPluginTypeName {
-  TOKEN_VOTING = "TokenVotingPlugin",
-  ADDRESS_LIST = "AddresslistVotingPlugin",
-  ADMIN = "AdminPlugin",
-  MULTISIG = "MultisigPlugin",
-}
 
-export const SubgraphPluginTypeMap: Map<
-  SubgraphPluginTypeName,
-  string
-> = new Map([
-  [SubgraphPluginTypeName.TOKEN_VOTING, "token-voting.plugin.dao.eth"],
-  [SubgraphPluginTypeName.ADDRESS_LIST, "address-list-voting.plugin.dao.eth"],
-  [SubgraphPluginTypeName.ADMIN, "admin.plugin.dao.eth"],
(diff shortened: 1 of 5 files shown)


mathewmeconry (approved):

modules/client/src/interfaces.ts
```diff
@@ -384,7 +387,12 @@ export type DaoResourceLink = { name: string; url: string };
 export type InstalledPluginListItem = {
   id: string;
   instanceAddress: string;
-  version: string;
+  installations: PluginInstallation[];
+};
+
+export type PluginInstallation = {
```
brickpop: Wouldn't the type be `PluginVersion`? or if we don't need to export it, `PluginEntry`?
And the field be `plugins`?

brickpop: I see that subgraph uses these names, but face to the developer I don't think they are super accurate

josemarinas: mmm, i'd go with `pluginVersions`

modules/client/src/internal/utils.ts
```diff
@@ -96,24 +95,22 @@ export function toDaoListItem(
...
-          // update when subgraph returns version
-          version: "0.0.1",
-        };
+      ): InstalledPluginListItem[] => {
+        return plugin.installations.map((installation) => {
+          return {
+            instanceAddress: plugin.id,
+            id: SubgraphPluginTypeMap.get(
+              plugin.__typename,
+            ) as string,
```
brickpop: What is this value? Is it the pluginRepo ENS?

brickpop (approved):

modules/client/src/internal/utils.ts
josemarinas: Yes, this gets a plugin typename and returns the ens. Now the subdomain can also be queried from subgraph so maybe we can get it from there insead of getting it from a hardcoded map, wdyt?

PR #202, opened 2023-04-11
Feature: Injection of custom ens registry and refactor context

josemarinas:
## Description

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -22,6 +22,8 @@ TEMPLATE:
 - Fix `Status.DEFEATED` filter
 - Error handling if ENS cannot be resolved.
 - Renames `mainnet` to `homestead` in `LIVE_CONTRACTS`.
+- `Context.network` type chges from `Networkish` to `Network`
+- Sdk automatically sets the parameters when not provided 
 ### Changed
 - Uses `earlyExecuted` for proposal state calculation in Addresslistvoting and Tokenvoting.
 ## [1.4.1]
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -26,6 +26,8 @@ TEMPLATE:
 - Context and ContextPlugin now automatically use the environment settings by default (depending on the ETH network) when no values are provided
 ### Changed
 - Uses `earlyExecuted` for proposal state calculation in Addresslistvoting and Tokenvoting.
+- Removed `daoRegistry` from the context because it wasn't being used
+- Removed `pluginRepoRegistry` from the context because it wasn't being used
 ## [1.4.1]
 ## Fixed
 - Amending the type in subgraph's `minApprovals`
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -22,8 +22,8 @@ TEMPLATE:
 - Fix `Status.DEFEATED` filter
 - Error handling if ENS cannot be resolved.
 - Renames `mainnet` to `homestead` in `LIVE_CONTRACTS`.
-- `Context.network` type chges from `Networkish` to `Network`
-- Sdk automatically sets the parameters when not provided 
+- `Context.network` type changes from `Networkish` to `Network`
+- Context and ContextPlugin now automatically use the environment settings by default (depending on the ETH network) when no values are provided
 ### Changed
 - Uses `earlyExecuted` for proposal state calculation in Addresslistvoting and Tokenvoting.
 ## [1.4.1]
--- a/modules/client/examples/index.ts
+++ b/modules/client/examples/index.ts
@@ -46,24 +46,24 @@ export const contextParams: ContextParams = {
   signer: new Wallet("private-key"),
   // Optional on "rinkeby", "arbitrum-rinkeby" or "mumbai"
   // Pass the address of the  `DaoFactory` contract you want to use. You can find it here based on your chain of choice: https://github.com/aragon/core/blob/develop/active_contracts.json
-  // By default we will use the one in the ethers-contract package
-  daoFactoryAddress: "0x1234381072385710239847120734123847123",
-  // Pass the address of the ensRegistry, if its not provided, for mainnet and worli we will use the defualt ones
-  // For polygon and mumbai we will use the ones deployed by ourselves, you can check them here: https://github.com/aragon/core/blob/develop/active_contracts.json
+  // Optional. Leave it empty to use Aragon's DAO Factory contract and claim a dao.eth subdomain
+  daoFactoryAddress: "0x1234381072385710239847120734123847123", 
+  // Optional. Pass the address of the ensRegistry for networks other than Mainnet or Goerli. 
+  // It will default to the registry deployed by Aragon. You can check them here: https://github.com/aragon/osx/blob/develop/active_contracts.json
   ensRegistryAddress: "0x1234381072385710239847120734123847123",
   // Choose your Web3 provider: Cloudfare, Infura, Alchemy, etc.
-  // Remember to also change it if you change your network
-  web3Providers: ["https://rpc.ankr.com/eth_goerli"],
-  // You can use our enpoints by leaving this field empty.
-  // They will point to the correct one depending on the network
+  // Remember to change the list of providers if a different network is selected
+  web3Providers: ["https://rpc.ankr.com/eth_goerli"],  
+  // Optional. By default, it will use Aragon's provided endpoints. 
+  // They will switch depending on the network (production, development)
   ipfsNodes: [
     {
       url: "https://testing-ipfs-0.aragon.network/api/v0",
       headers: { "X-API-KEY": IPFS_API_KEY || "" },
     },
   ],
-  // You can use our enpoints by leaving this field empty.
-  // They will point to the correct one depending on the network
+  // Optional. By default it will use Aragon's provided endpoints.
+  // They will switch depending on the network (production, development)
   graphqlNodes: [
     {
       url: "https://subgraph.satsuma-prod.com/aragon/core-goerli/api",
@@ -71,8 +71,17 @@ export const contextParams: ContextParams = {
   ],
 };
 
+// Bare minimoum context that will use de default values
+const minimalContextParams:ContextParams =  {
+  network: 'mainnet',
+  web3Providers: 'https://example.com/your_mainnet_rpc',
+  signer: new Wallet('private-key')
+}
+
 // Instantiate the Aragon SDK context
 export const context: Context = new Context(contextParams);
+// Instantiate the Aragon SDK context
+export const minimalContext: Context = new Context(minimalContextParams);
 
 /* MARKDOWN
 Update the context with new parameters if you wish to throughout your app.
--- a/modules/client/examples/index.ts
+++ b/modules/client/examples/index.ts
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
   signer: new Wallet("private-key"),
   // Optional on "rinkeby", "arbitrum-rinkeby" or "mumbai"
   // Pass the address of the  `DaoFactory` contract you want to use. You can find it here based on your chain of choice: https://github.com/aragon/core/blob/develop/active_contracts.json
+  // By default we will use the one in the ethers-contract package
   daoFactoryAddress: "0x1234381072385710239847120734123847123",
+  // Pass the address of the ensRegistry, if its not provided, for mainnet and worli we will use the defualt ones
+  // For polygon and mumbai we will use the ones deployed by ourselves, you can check them here: https://github.com/aragon/core/blob/develop/active_contracts.json
+  ensRegistryAddress: "0x1234381072385710239847120734123847123",
   // Choose your Web3 provider: Cloudfare, Infura, Alchemy, etc.
+  // Remember to also change it if you change your network
   web3Providers: ["https://rpc.ankr.com/eth_goerli"],
+  // You can use our enpoints by leaving this field empty.
+  // They will point to the correct one depending on the network
   ipfsNodes: [
     {
       url: "https://testing-ipfs-0.aragon.network/api/v0",
       headers: { "X-API-KEY": IPFS_API_KEY || "" },
     },
   ],
-  // Don't change this line. This is how we connect your app to the Aragon subgraph.
+  // You can use our enpoints by leaving this field empty.
+  // They will point to the correct one depending on the network
   graphqlNodes: [
     {
       url: "https://subgraph.satsuma-prod.com/aragon/core-goerli/api",
@@ -72,4 +80,4 @@ Update the context with new parameters if you wish to throughout your app.
 
 context.set({ network: 1 });
 context.set({ signer: new Wallet("private key") }); // if you're using wagmi library, you can also get the signer through their [`useSigner` method](https://wagmi.sh/react/hooks/useSigner) inside a `useEffect` hook.
-context.setFull(contextParams);
+context.set(contextParams);
--- a/modules/client/examples/index.ts
+++ b/modules/client/examples/index.ts
@@ -43,18 +43,18 @@ export const contextParams: ContextParams = {
   // Choose the network you want to use. You can use "goerli" or "mumbai" for testing, "mainnet" for Ethereum.
   network: "goerli",
   // Depending on how you're configuring your wallet, you may want to pass in a `signer` object here.
-  signer: new Wallet("private-key"),
+  signer: Wallet.createRandom(),
   // Optional on "rinkeby", "arbitrum-rinkeby" or "mumbai"
   // Pass the address of the  `DaoFactory` contract you want to use. You can find it here based on your chain of choice: https://github.com/aragon/core/blob/develop/active_contracts.json
   // Optional. Leave it empty to use Aragon's DAO Factory contract and claim a dao.eth subdomain
-  daoFactoryAddress: "0x1234381072385710239847120734123847123", 
-  // Optional. Pass the address of the ensRegistry for networks other than Mainnet or Goerli. 
+  daoFactoryAddress: "0x1234381072385710239847120734123847123",
+  // Optional. Pass the address of the ensRegistry for networks other than Mainnet or Goerli.
   // It will default to the registry deployed by Aragon. You can check them here: https://github.com/aragon/osx/blob/develop/active_contracts.json
   ensRegistryAddress: "0x1234381072385710239847120734123847123",
   // Choose your Web3 provider: Cloudfare, Infura, Alchemy, etc.
   // Remember to change the list of providers if a different network is selected
-  web3Providers: ["https://rpc.ankr.com/eth_goerli"],  
-  // Optional. By default, it will use Aragon's provided endpoints. 
+  web3Providers: ["https://rpc.ankr.com/eth_goerli"],
+  // Optional. By default, it will use Aragon's provided endpoints.
   // They will switch depending on the network (production, development)
   ipfsNodes: [
     {
@@ -72,11 +72,11 @@ export const contextParams: ContextParams = {
 };
 
 // Bare minimoum context that will use de default values
-const minimalContextParams:ContextParams =  {
-  network: 'mainnet',
-  web3Providers: 'https://example.com/your_mainnet_rpc',
-  signer: new Wallet('private-key')
-}
+const minimalContextParams: ContextParams = {
+  network: "mainnet",
+  web3Providers: "https://eth.llamarpc.com",
+  signer: Wallet.createRandom(),
+};
 
 // Instantiate the Aragon SDK context
 export const context: Context = new Context(contextParams);
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -29,29 +29,34 @@ export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
   maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
 };
 
+const getGraphqlNode = (netowrk: string): string => {
+  return `https://subgraph.satsuma-prod.com/${
+    process.env.SATSUMA_API_KEY || ""
+  }/aragon/${netowrk}/api`;
+};
+
 export const GRAPHQL_NODES: { [K in SupportedNetworks]: { url: string }[] } = {
-  homestead: [
+  homestead: [{ url: getGraphqlNode("osx-mainnet") }],
+  goerli: [{ url: getGraphqlNode("osx-goerli") }],
+  matic: [{ url: getGraphqlNode("osx-polygon") }],
+  maticmum: [{ url: getGraphqlNode("osx-mumbai") }],
+};
+
+const IPFS_ENDPOINTS = {
+  prod: [
     {
-      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
-        "" + "/aragon/osx-mainnet/api",
+      url: "https://ipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+    {
+      url: "https://ipfs-1.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
     },
   ],
-  goerli: [
+  test: [
     {
-      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
-        "" + "/aragon/osx-goerli/api",
-    },
-  ],
-  matic: [
-    {
-      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
-        "" + "/aragon/osx-polygon/api",
-    },
-  ],
-  maticmum: [
-    {
-      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
-        "" + "/aragon/osx-mumbai/api",
+      url: "https://testingipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_TEST_API_KEY || "" },
     },
   ],
 };
@@ -62,38 +67,10 @@ export const IPFS_NODES: {
     headers?: Record<string, string> | undefined;
   }[];
 } = {
-  homestead: [
-    {
-      url: "https://ipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-    {
-      url: "https://ipfs-1.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-  ],
-  goerli: [
-    {
-      url: "https://testingipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-  ],
-  matic: [
-    {
-      url: "https://ipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-    {
-      url: "https://ipfs-1.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-  ],
-  maticmum: [
-    {
-      url: "https://testingipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
-    },
-  ],
+  homestead: IPFS_ENDPOINTS.prod,
+  goerli: IPFS_ENDPOINTS.test,
+  matic: IPFS_ENDPOINTS.prod,
+  maticmum: IPFS_ENDPOINTS.test,
 };
 
 export const LIVE_CONTRACTS: { [K in SupportedNetworks]: NetworkDeployment } = {
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -23,7 +23,7 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
 
 // TODO change rpcs to aragon ones
 export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
-  homestead: ["https://mainnet.eth.aragon.network/"],
+  homestead: ["https://mainnet.eth.aragon.network"],
   goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
   matic: ["https://polygon.llamarpc.com"],
   maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -23,7 +23,7 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
 
 // TODO change rpcs to aragon ones
 export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
-  homestead: ["https://eth.llamarpc.com"],
+  homestead: ["https://mainnet.eth.aragon.network/"],
   goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
   matic: ["https://polygon.llamarpc.com"],
   maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -21,6 +21,49 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
   resources: [],
 };
 
+// TODO, fill me with data
+export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
+  homestead: [],
+  goerli: [],
+  matic: [],
+  maticmum: [],
+};
+
+export const GRAPHQL_NODES: { [K in SupportedNetworks]: { url: string }[] } = {
+  homestead: [{
+    url: "",
+  }],
+  goerli: [{
+    url: "",
+  }],
+  matic: [{
+    url: "",
+  }],
+  maticmum: [{
+    url: "",
+  }],
+};
+
+export const IPFS_NODES: {
+  [K in SupportedNetworks]: {
+    url: string;
+    headers?: Record<string, string> | undefined;
+  }[];
+} = {
+  homestead: [{
+    url: "",
+  }],
+  goerli: [{
+    url: "",
+  }],
+  matic: [{
+    url: "",
+  }],
+  maticmum: [{
+    url: "",
+  }],
+};
+
 export const LIVE_CONTRACTS: { [K in SupportedNetworks]: NetworkDeployment } = {
   homestead: {
     daoFactory: activeContractsList.mainnet.DAOFactory,
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -21,34 +21,54 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
   resources: [],
 };
 
-const getGraphqlNode = (netowrk: string): string => {
+const getGraphqlNode = (netowrk: SupportedNetworks): string => {
   return `https://subgraph.satsuma-prod.com/${
     process.env.SATSUMA_API_KEY || "qHR2wGfc5RLi6"
-  }/aragon/${netowrk}/api`;
+  }/aragon/osx-${SupportedNetworksToGraphqlNetworks[netowrk]}/api`;
+};
+
+type GraphqlNetworks = "mainnet" | "goerli" | "polygon" | "mumbai";
+
+const SupportedNetworksToGraphqlNetworks: {
+  [K in SupportedNetworks]: GraphqlNetworks;
+} = {
+  homestead: "mainnet",
+  goerli: "goerli",
+  matic: "polygon",
+  maticmum: "mumbai",
 };
 
 export const GRAPHQL_NODES: { [K in SupportedNetworks]: { url: string }[] } = {
-  homestead: [{ url: getGraphqlNode("osx-mainnet") }],
-  goerli: [{ url: getGraphqlNode("osx-goerli") }],
-  matic: [{ url: getGraphqlNode("osx-polygon") }],
-  maticmum: [{ url: getGraphqlNode("osx-mumbai") }],
+  homestead: [{ url: getGraphqlNode("homestead") }],
+  goerli: [{ url: getGraphqlNode("goerli") }],
+  matic: [{ url: getGraphqlNode("matic") }],
+  maticmum: [{ url: getGraphqlNode("maticmum") }],
 };
 
 const IPFS_ENDPOINTS = {
   prod: [
     {
       url: "https://ipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt" },
+      headers: {
+        "X-API-KEY": process.env.IPFS_API_KEY ||
+          "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt",
+      },
     },
     {
       url: "https://ipfs-1.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt" },
+      headers: {
+        "X-API-KEY": process.env.IPFS_API_KEY ||
+          "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt",
+      },
     },
   ],
   test: [
     {
       url: "https://testing-ipfs-0.aragon.network",
-      headers: { "X-API-KEY": process.env.IPFS_TEST_API_KEY || "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt" },
+      headers: {
+        "X-API-KEY": process.env.IPFS_TEST_API_KEY ||
+          "b477RhECf8s8sdM7XrkLBs2wHc4kCMwpbcFC55Kt",
+      },
     },
   ],
 };
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -21,27 +21,39 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
   resources: [],
 };
 
-// TODO, fill me with data
+// TODO change rpcs to aragon ones
 export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
-  homestead: [],
-  goerli: [],
-  matic: [],
-  maticmum: [],
+  homestead: ["https://eth.llamarpc.com"],
+  goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
+  matic: ["https://polygon.llamarpc.com"],
+  maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
 };
 
 export const GRAPHQL_NODES: { [K in SupportedNetworks]: { url: string }[] } = {
-  homestead: [{
-    url: "",
-  }],
-  goerli: [{
-    url: "",
-  }],
-  matic: [{
-    url: "",
-  }],
-  maticmum: [{
-    url: "",
-  }],
+  homestead: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
+        "" + "/aragon/osx-mainnet/api",
+    },
+  ],
+  goerli: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
+        "" + "/aragon/osx-goerli/api",
+    },
+  ],
+  matic: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
+        "" + "/aragon/osx-polygon/api",
+    },
+  ],
+  maticmum: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
+        "" + "/aragon/osx-mumbai/api",
+    },
+  ],
 };
 
 export const IPFS_NODES: {
@@ -50,18 +62,38 @@ export const IPFS_NODES: {
     headers?: Record<string, string> | undefined;
   }[];
 } = {
-  homestead: [{
-    url: "",
-  }],
-  goerli: [{
-    url: "",
-  }],
-  matic: [{
-    url: "",
-  }],
-  maticmum: [{
-    url: "",
-  }],
+  homestead: [
+    {
+      url: "https://ipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+    {
+      url: "https://ipfs-1.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+  ],
+  goerli: [
+    {
+      url: "https://testingipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+  ],
+  matic: [
+    {
+      url: "https://ipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+    {
+      url: "https://ipfs-1.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+  ],
+  maticmum: [
+    {
+      url: "https://testingipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
+  ],
 };
 
 export const LIVE_CONTRACTS: { [K in SupportedNetworks]: NetworkDeployment } = {
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -23,9 +23,9 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
 
 // TODO change rpcs to aragon ones
 export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
-  homestead: ["https://mainnet.eth.aragon.network"],
+  homestead: ["https://mainnet.eth.aragon.network", "https://eth.llamarpc.com", "https://rpc.mevblocker.io"],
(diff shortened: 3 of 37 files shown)


modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,81 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
   resources: [],
 };
 
+// TODO change rpcs to aragon ones
```
brickpop: TO DO?

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,81 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
+export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
+  homestead: ["https://mainnet.eth.aragon.network"],
+  goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
+  matic: ["https://polygon.llamarpc.com"],
+  maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
+};
+
+export const GRAPHQL_NODES: { [K in SupportedNetworks]: { url: string }[] } = {
```
brickpop, line 41: The type could simple be a string[], not an object with url

Also, we would be the ones using these details, not sure the rest of the world would require that we export everywhere

modules/client/src/client-common/interfaces/core.ts
```diff
@@ -4,24 +4,22 @@ import { Signer } from "@ethersproject/abstract-signer";
...
+import { GasFeeEstimation } from "./common";
 
 export interface IClientWeb3Core {
-  useSigner: (signer: Signer) => void;
   shiftProvider: () => void;
-  getSigner: () => Signer | null;
+  getSigner: () => Signer;
   getConnectedSigner: () => Signer;
-  getProvider: () => JsonRpcProvider | null;
+  getProvider: () => JsonRpcProvider;
```
brickpop, line 14: Lovely removal of `| null`

Is there any place where we could remove existing null checks?

modules/client/src/client-common/interfaces/context.ts
```diff
@@ -30,19 +35,22 @@ export type ContextParams = Web3ContextParams &
 // Context state data
 
 type Web3ContextState = {
-  network: Networkish;
-  signer?: Signer;
-  daoFactoryAddress?: string;
-  daoRegistryAddress?: string;
-  pluginRepoRegistryAddress?: string;
+  network: Network;
```
brickpop, line 41: What's the difference between Networkish and Network?

I wonder wether we want to couple things with Ethers. It might be inevitable

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
-const defaultState: ContextState = {
-  network: "mainnet",
-  web3Providers: [],
-  gasFeeEstimationFactor: DEFAULT_GAS_FEE_ESTIMATION_FACTOR,
-};
-
 export class Context {
-  protected state: ContextState = Object.assign({}, defaultState);
+  protected state: ContextState = {} as ContextState;
+  protected manual: ManualState = {
```
brickpop: `manualState` instead of `manual`?

I wonder whether it would be cleaner to have a `state` and a boolean map to track if certain fields were changed or not. But having a single source of truth, in regards of the settings

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
-        contextParams.gasFeeEstimationFactor
+        contextParams.gasFeeEstimationFactor,
       );
+      this.manual.gasFeeEstimationFactor = true;
     }
+  }
 
-    if (contextParams.ipfsNodes?.length) {
-      this.state.ipfs = Context.resolveIpfs(contextParams.ipfsNodes);
+  setDefault() {
```
brickpop: Is `setDefault()` still needed, after we provide the defaults ourselves? Is someone using it?

brickpop (commented):
Round 1

modules/client/src/client-common/interfaces/context.ts
josemarinas: Networkish is basically `string | number | Network`, `Network` is the object used by ethers the structure is teh following:

```
{
    chainId: number,
    name: string,
    ensAddress?: string,
}
```
So the idea is that we receive a `Networkish`, which can be a network, a string with the name or the chain id, and we change it to a `Netowrk` type so it can be usable by the `JsonRPCProvider`

modules/client/src/client-common/interfaces/core.ts
josemarinas: Yes we need another ticket to remove all the signer and signer.provider checks because now this is done automatically done on the context and if you call `getConnectedSigner` it'll return a signer with a provider asocciated or throw an error, way mor elegant than the way we were doing it

modules/client/src/client-common/constants.ts
josemarinas: Yes, rn we only have RPCs for mainnet and I believe goerli, for polygon and mumbai we must use external RPCs while we dont have ours.

We can create a ticket for devops to deploy this rpcs or trust the public rpcs

modules/client/src/client-common/constants.ts
josemarinas: Yes it could be a simple string, but we left it as `{ url: string }` following the pattern of the ipfs nodes where we add a header with the api key

About exporting it, we need to export it here, but we mus remove the export of the constants from the `ìndex.ts` file if we dont want to export it to the exterior

modules/client/src/client-common/context.ts
josemarinas: Having an unique variable is going to make the types messier, besides, the `network` and `signer` properties dont have manual

modules/client/src/client-common/context.ts
josemarinas: We can make it private, but this is called after each `set`that includes a change of the property `network`.

modules/client/src/client-common/constants.ts
brickpop: We can use the publicly available ones. Having ours and other alternatives is the best approach for decentralization

modules/client/src/client-common/interfaces/core.ts
brickpop: Isn't it easy to do it on this PR?

modules/client/src/client-common/context.ts
brickpop: I was rather meaning something like:

```ts
protected state: ContextState = Object.assign({}, defaultState);

protected overridden = {
  web3Nodes: false,
  ipfsNodes: false,
  subgraphEndpoints: false,
 ...
}


```

modules/client/src/client-common/context.ts
brickpop: Then shouldn't setDefault include the network as a parameter, to do a full wipe?

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,81 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
+  ],
+  goerli: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
+        "" + "/aragon/osx-goerli/api",
+    },
+  ],
+  matic: [
+    {
+      url: "https://subgraph.satsuma-prod.com/" + process.env.SATSUMA_API_KEY ||
```
brickpop: Instead of repeating the same code for each network, we could have 2 internal constants, one for production and one for development

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,81 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
+  [K in SupportedNetworks]: {
+    url: string;
+    headers?: Record<string, string> | undefined;
+  }[];
+} = {
+  homestead: [
+    {
+      url: "https://ipfs-0.aragon.network",
+      headers: { "X-API-KEY": process.env.IPFS_API_KEY || "" },
+    },
```
brickpop: Same (recycle)

modules/client/src/client-common/index.ts
```diff
@@ -7,4 +7,9 @@ export * from "./context";
 export * from "./context-plugin";
 export * from "./utils";
 export * from "./encoding";
-export * from "./constants";
+export {
+  EMPTY_PROPOSAL_METADATA_LINK,
+  LIVE_CONTRACTS,
+  UNAVAILABLE_PROPOSAL_METADATA,
+  UNSUPPORTED_PROPOSAL_METADATA_LINK,
+} from "./constants";
```
brickpop, line 15: Aren't we going to break anything if we leave explicit exports only?

modules/client/src/client-common/interfaces/core.ts
```diff
@@ -4,24 +4,22 @@ import { Signer } from "@ethersproject/abstract-signer";
...
   getMaxFeePerGas: () => Promise<bigint>;
   isUp: () => Promise<boolean>;
   ensureOnline: () => Promise<void>;
-  attachContract: <T>(
+  attachContract:(
     address: string,
     abi: ContractInterface,
-  ) => Contract & T;
+  ) => Contract;
```
brickpop: Why removing the generic type? People won't be able to attach to a custom contract

modules/client/src/client-common/modules/web3.ts
```diff
@@ -152,53 +88,36 @@ export class Web3Module implements IClientWeb3Core {
    * @param abi The Application Binary Inteface of the contract
    * @return A contract instance attached to the given address
    */
-  public attachContract<T>(
+  public attachContract(
```
brickpop: Allowing to type the contract you attach to is very helpful for consumers

modules/client/test/integration/addresslistVoting-client/decoding.test.ts
```diff
@@ -5,12 +5,18 @@ import {
...
   VotingSettings,
 } from "../../../src";
-import { contextParamsLocalChain } from "../constants";
+import { ADDRESS_ONE, contextParamsLocalChain } from "../constants";
+
+jest.spyOn(SupportedNetworksArray, 'includes').mockReturnValue(true)
 
 describe("Client Address List", () => {
+  beforeAll(() => {
+    contextParamsLocalChain.ensRegistryAddress = ADDRESS_ONE
```
brickpop: Won't this make DAO creations fail? Why is it added in the context of decoding tests?

modules/client/test/integration/addresslistVoting-client/encoding.test.ts
```diff
@@ -6,11 +6,18 @@ import {
...
 } from "../../../src";
 import { bytesToHex, InvalidAddressError } from "@aragon/sdk-common";
-import { contextParamsLocalChain } from "../constants";
+import { ADDRESS_ONE, contextParamsLocalChain } from "../constants";
+
+jest.spyOn(SupportedNetworksArray, 'includes').mockReturnValue(true)
+
 describe("Client Address List", () => {
+  beforeAll(() => {
+    contextParamsLocalChain.ensRegistryAddress = ADDRESS_ONE
```
brickpop: same

modules/client/test/integration/addresslistVoting-client/methods.test.ts
```diff
@@ -60,6 +61,8 @@ import {
 import { AddresslistVotingPluginPrepareInstallationParams } from "../../../src/addresslistVoting/interfaces";
 import { LIVE_CONTRACTS } from "../../../src/client-common/constants";
 
+jest.spyOn(SupportedNetworksArray, 'includes').mockReturnValue(true)
```
brickpop: Don't we need to clean this up after these tests have run? Won't this affect the rest of test files?

modules/client/test/integration/client/index.test.ts
```diff
@@ -31,24 +31,5 @@ describe("Client", () => {
       const web3Status = await client.web3.isUp();
       expect(web3Status).toEqual(false);
     });
-    it("Should create a client, fail and shift to a working endpoint", async () => {
-      contextParamsMainnet.web3Providers = web3endpoints.failing.concat(
```
brickpop: Why dropping this test?

modules/client/test/mocks/ethers-providers.ts
```diff
@@ -0,0 +1,8 @@
+const original = jest.requireActual("@ethersproject/providers");
+export const mockedEthersProviders = {
+  ...original,
+  getNetwork: jest.fn().mockReturnValue({ chainId: 31337, name: "goerli" }),
```
brickpop: Q: no ensRegistry?

modules/client/test/unit/context-plugins.test.ts
```diff
@@ -27,29 +36,47 @@ describe("ContextPlugin instances", () => {
...
+    context.graphql.map((graphqlClient) =>
+      expect(graphqlClient).toBeInstanceOf(GraphQLClient)
+    );
   });
   it("Should create a context and have the correct values", () => {
     const context = new ContextPlugin(contextParams);
 
     expect(context).toBeInstanceOf(ContextPlugin);
-    expect(context.network).toEqual("mainnet");
+    expect(context.network.name).toEqual("homestead");
```
brickpop, line 66: If this is a breaking change, the changelog should mention it

modules/client/test/unit/context-plugins.test.ts
```diff
@@ -67,15 +94,55 @@ describe("ContextPlugin instances", () => {
...
-    expect(context.network).toEqual("goerli");
-    expect(context.daoFactoryAddress).toEqual(
-      activeContractsList.goerli.DAOFactory,
+    expect(context).toBeInstanceOf(Context);
+    expect(context.network.name).toBe("matic");
+    expect(context.network.chainId).toBe(137);
+    expect(context.daoFactoryAddress).toBe(LIVE_CONTRACTS.matic.daoFactory);
+    expect(context.ensRegistryAddress).toBe(LIVE_CONTRACTS.matic.ensRegistry);
+    expect(context.gasFeeEstimationFactor).toBe(0.625);
+    expect(context.web3Providers.length).toBe(WEB3_NODES.matic.length);
```
brickpop: I would do a deep equal, instead of just the length

modules/client/test/unit/context.test.ts
```diff
@@ -27,54 +36,119 @@ describe("Context instances", () => {
...
   });
   it("Should create a context and have the correct values", () => {
     const context = new Context(contextParams);
 
     expect(context).toBeInstanceOf(Context);
-    expect(context.network).toEqual("mainnet");
-    expect(context.signer).toBeInstanceOf(Wallet);
-    expect(context.daoFactoryAddress).toEqual("0x1234");
-    context.web3Providers?.map((provider) =>
+    expect(context.network.name).toBe("homestead");
```
brickpop, line 62: Why don't we try with values that are actually different to the default ones?

modules/client/test/unit/context.test.ts
```diff
@@ -27,54 +36,119 @@ describe("Context instances", () => {
...
     );
+    context.ipfs?.map((ipfsClient) =>
+      expect(ipfsClient).toBeInstanceOf(IpfsClient)
+    );
+    context.graphql?.map((graphqlClient) =>
+      expect(graphqlClient).toBeInstanceOf(GraphQLClient)
+    );
     expect(context.gasFeeEstimationFactor).toEqual(0.1);
   });
+  it("Should Change the network and update all the parameters", () => {
```
brickpop, line 172: I'm missing certain test cases:

- Create a context with empty params. Set the network to mumbai and see that everything changed
- Create a context by passing only the network (goerli). See that all params are from development. Then set the network to matic. See that everything changed again

modules/common/src/errors.ts
```diff
@@ -153,3 +153,24 @@ export class DataDecodingError extends Error {
     super("Cannot decode data: " + message);
   }
 }
+
+export class EnsUnsuportedByNetworkError extends Error {
```
brickpop: Aren't e providing our own? When is this used?

modules/common/src/errors.ts
```diff
@@ -153,3 +153,24 @@ export class DataDecodingError extends Error {
...
+  constructor(network: string) {
+    super("Network " + network + " dows not support ens");
```
brickpop:
```suggestion
    super("Network " + network + " does not support ens");
```

modules/common/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-common",
   "author": "Aragon Association",
-  "version": "1.1.1",
+  "version": "1.1.2",
```
brickpop: If we are adding:

```suggestion
  "version": "1.2.0",
```

brickpop (commented):
Round 2

modules/client/src/client-common/interfaces/core.ts
josemarinas: Sure I can do it on this PR, there is a lot of them though, so its going to be a big one

modules/client/src/client-common/interfaces/core.ts
josemarinas: Deleted it while refactoring, forgot to add it back, my bad

modules/client/src/client-common/constants.ts
josemarinas: Yes, but the selected node is randomly selected between one of the nodes in the list. This means that it wont always select the aragon node, if we are ok with thiss i'll add them

modules/client/src/client-common/index.ts
josemarinas: Nope, I exported everything except the new ones that we dont want to export

modules/client/test/integration/addresslistVoting-client/decoding.test.ts
josemarinas: The problem with this is that in testing the network name is `unknown` so the tests crash when trying to initialize the context, seting this manually avoids the tests crashing

modules/client/test/integration/addresslistVoting-client/methods.test.ts
josemarinas: This only affects this test file, and we are okay with having this includes to true because the network in all tests is `unknown`

modules/client/test/integration/client/index.test.ts
josemarinas: Because is not deterministic since the web3provider is selected randomly from the available ones

modules/client/test/mocks/ethers-providers.ts
josemarinas: nope this was a file that i created for testing but it ended up not working up well because the JSONRpcProvider detect the network changes and crashes

modules/client/test/unit/context-plugins.test.ts
josemarinas: It shouldnt break anything on the app becuse i believe they dont use the netowrk param of the context however, I'll mention it in the changelog just in case

modules/client/test/unit/context-plugins.test.ts
josemarinas: We cannot do that, one object is just the url the other one is a `JSONRpcProvider`

modules/common/src/errors.ts
josemarinas: yes this is an edge case, where for some reason the ensAddress is empty in the provider and we dont have one to fill it with, we can delete the check if it does not make sense to you

modules/client/src/client-common/constants.ts
brickpop: I guess that's the main benefit. We don't always use our own node, so you get a fail safe client by default

modules/client/src/client-common/interfaces/core.ts
brickpop: mmmm ok, then let's create a separate task

modules/client/test/integration/client/index.test.ts
brickpop: Can we temporarily override `Math.random()` so that we know which one is going to happen?

modules/client/test/mocks/ethers-providers.ts
brickpop: Right, but the mock value we are returning has no ensRegistry. Maybe we can prevent all the manual mocks by setting this to include an ensRegistry?

modules/client/test/unit/context-plugins.test.ts
brickpop: How about comparing the URL values then?

modules/common/src/errors.ts
brickpop: Not sure I follow. If we provide our own ENS always, would such an error ever be used?

modules/client/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ TEMPLATE:
 ### Fixed
 - Error handling if ENS cannot be resolved.
 - Renames `mainnet` to `homestead` in `LIVE_CONTRACTS`.
+- `Context.network` type chges from `Networkish` to `Network`
+- Sdk automatically sets the parameters when not provided
```
```suggestion
- Context and ContextPlugin now automatically use the environment settings by default (depending on the ETH network) when no values are provided
```

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,58 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
+export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
+  homestead: ["https://mainnet.eth.aragon.network"],
+  goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
+  matic: ["https://polygon.llamarpc.com"],
+  maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
+};
+
+const getGraphqlNode = (netowrk: string): string => {
+  return `https://subgraph.satsuma-prod.com/${
+    process.env.SATSUMA_API_KEY || ""
```
brickpop: `process.env.XXX` is a NodeJS-only feature, not standard JS. We should receive an explicit parameter or use a hardcoded constant provided by us

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,58 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
 
+// TODO change rpcs to aragon ones
+export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
+  homestead: ["https://mainnet.eth.aragon.network"],
+  goerli: ["https://endpoints.omniatech.io/v1/eth/goerli/public"],
+  matic: ["https://polygon.llamarpc.com"],
+  maticmum: ["https://endpoints.omniatech.io/v1/matic/mumbai/public"],
+};
+
+const getGraphqlNode = (netowrk: string): string => {
```
brickpop: `network` should be accepting `SupportedNetwork` values only, and we should be returning the result based on these known values... not any random string

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
+  LIVE_CONTRACTS,
+  WEB3_NODES,
+} from "./constants";
+import { SupportedNetworks, SupportedNetworksArray } from "./interfaces/common";
+import { isAddress } from "@ethersproject/address";
+import { Signer } from "@ethersproject/abstract-signer";
 export { ContextParams } from "./interfaces/context";
 
 const DEFAULT_GAS_FEE_ESTIMATION_FACTOR = 0.625;
+type ManualState = {
```
brickpop:
```suggestion
type OverriddenState = {
```

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
-const defaultState: ContextState = {
-  network: "mainnet",
-  web3Providers: [],
-  gasFeeEstimationFactor: DEFAULT_GAS_FEE_ESTIMATION_FACTOR,
-};
-
 export class Context {
-  protected state: ContextState = Object.assign({}, defaultState);
+  protected state: ContextState = {} as ContextState;
+  protected manualState: ManualState = {
```
brickpop:
```suggestion
  protected overridden: OverriddenState = {
```

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
+      this.manualState.gasFeeEstimationFactor = true;
     }
+  }
 
-    if (contextParams.ipfsNodes?.length) {
-      this.state.ipfs = Context.resolveIpfs(contextParams.ipfsNodes);
+  private setDefault() {
+    const networkName = this.state.network.name as SupportedNetworks;
+    if (
+      WEB3_NODES[networkName] && WEB3_NODES[networkName].length &&
```
brickpop: If the length was zero I would rather throw an internal error

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
     if (contextParams.signer) {
       this.state.signer = contextParams.signer;
     }
-    if (contextParams.web3Providers) {
+    if (
+      contextParams.web3Providers ||
+      (Array.isArray(contextParams.web3Providers) &&
+        contextParams.web3Providers.length)
+    ) {
       this.state.web3Providers = Context.resolveWeb3Providers(
```
brickpop, line 76: Should we change that the given network matches or throw an Error?

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
-        contextParams.gasFeeEstimationFactor
+        contextParams.gasFeeEstimationFactor,
       );
+      this.manualState.gasFeeEstimationFactor = true;
     }
+  }
 
-    if (contextParams.ipfsNodes?.length) {
-      this.state.ipfs = Context.resolveIpfs(contextParams.ipfsNodes);
+  private setDefault() {
```
brickpop:
```suggestion
  private setNetworkDefaults() {
```

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
+    ) {
+      this.state.web3Providers = Context.resolveWeb3Providers(
+        WEB3_NODES[networkName],
+        this.state.network,
+      );
     }
-    if (contextParams.graphqlNodes?.length) {
-      this.state.graphql = Context.resolveGraphql(contextParams.graphqlNodes);
+    if (
+      GRAPHQL_NODES[networkName] && GRAPHQL_NODES[networkName].length &&
```
brickpop: Same

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
+    ) {
+      this.state.graphql = Context.resolveGraphql(GRAPHQL_NODES[networkName]);
+    }
+    if (
+      IPFS_NODES[networkName] && IPFS_NODES[networkName].length &&
+      !this.manualState.ipfs
+    ) {
+      this.state.ipfs = Context.resolveIpfs(IPFS_NODES[networkName]);
+    }
+    if (LIVE_CONTRACTS[networkName]) {
```
brickpop: Same, about throwing an error instead of failing silently

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,151 @@
...
+      }
+      if (!this.manualState.ensRegistry) {
+        let ensRegistry = LIVE_CONTRACTS[networkName].ensRegistry;
+        if (!ensRegistry) {
+          ensRegistry = this.network.ensAddress;
+        }
+        this.state.ensRegistryAddress = ensRegistry;
+      }
+    }
+    if (!this.manualState.gasFeeEstimationFactor) {
```
brickpop: no override check here?

modules/client/src/client-common/context.ts
```diff
@@ -211,48 +232,61 @@ export class Context {
...
+    const network = getNetwork(networkish);
+    const networkName = network.name as SupportedNetworks;
+    if (!SupportedNetworksArray.includes(networkName)) {
+      throw new UnsupportedNetworkError(networkName);
     }
-    if (params.signer) {
-      defaultState.signer = params.signer;
+    if (ensRegistryAddress) {
+      if (!isAddress(ensRegistryAddress)) {
+        console.warn("invalid address, using default");
```
brickpop: Why not throwing an error? If it was empty, it's fine to fall back to the default. But if it's incorrect, the developer should be aware of that

modules/client/src/client-common/context.ts
```diff
@@ -211,48 +232,61 @@ export class Context {
...
+        network.ensAddress = ensRegistryAddress;
+      }
     }
+    if (!network.ensAddress) {
+      const ensAddress = LIVE_CONTRACTS[networkName].ensRegistry;
+      if (!ensAddress) {
+        throw new EnsUnsuportedByNetworkError(networkName);
+      }
+      network.ensAddress = ensAddress;
+    }
```
brickpop: Why not combining these 2 blocks into a single one?

brickpop (commented):
Round 3

modules/client/test/unit/context.test.ts
josemarinas: I'mean they are different this is using the contextParams defined at the top of the file, in the next test we test with another parameters

modules/client/src/client-common/context.ts
josemarinas: Mmm, i dont think so, it wouldnt make much sense changing the network since all the default parameters depend on this

modules/client/src/client-common/context.ts
josemarinas: Sure :)

modules/client/src/client-common/context.ts
josemarinas: We check it, not sure what you mean

modules/client/src/client-common/context.ts
josemarinas: okay, I lef it as an option but we can throw an error

modules/client/src/client-common/context.ts
josemarinas: We cannot combine it it would change the code works:

1. The network object is generated and can contain an `ensAddress`
2. Check if the ensRegistryAddress is specified
3. If its specified and is incorrect, throw an error
4. If is correct, override the value in `network.ensAddress`

If we try to unify this we will override the value of `network.ensAddress` even when the `ensRegistryAddress` is not specified

modules/client/src/client-common/constants.ts
josemarinas: Ok, I added a buynch of endpoints, feel free to sugest any public rpc

modules/client/src/client-common/constants.ts
josemarinas: Okay, I'll create a couple of constants for this, I didn't want to expose the satsuma api keys and ifs nodes keys but if there is no other way :(

modules/client/src/client-common/constants.ts
josemarinas: nope, because this is called only by the the `GRAPHQL_NODES` constants, and the networks in satsuma have diffeerent names (osx-mainnet, osx-polygon...) so I'd leave it like this or create a new enum with the satsuma network names, because they are not the same as the `SupportedNetworks`

modules/client/test/unit/context-plugins.test.ts
josemarinas: I tried but is not that easy, because the graphql and web nodes dont not expose the url, and the ipfs exposes it but modifies it

modules/common/src/errors.ts
josemarinas: I just used it in one check, changed the error to 'UnsupportedNetworkError'

modules/client/src/client-common/constants.ts
brickpop: You can remove the `TODO` line that originated the comment, as well :)

modules/client/src/client-common/constants.ts
brickpop: For Satsuma, we definitely need a given API key 🤷 
Developers will use process.env to pass it on their end and give it to us, but we should be 100% unopinionated on our side

modules/client/src/client-common/context.ts
brickpop: ping

modules/client/src/client-common/constants.ts
```diff
@@ -21,6 +21,58 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
   resources: [],
 };
 
+// TODO change rpcs to aragon ones
+export const WEB3_NODES: { [K in SupportedNetworks]: string[] } = {
```
brickpop: They are not exactly nodes, they are RPC endpoints that talk to actual ETH nodes.
`WEB3_RPCS` or `WEB3_ENDPOINTS`?

modules/client/src/client-common/constants.ts
brickpop: But, can't we do a `switch` to convert this? or have an auxiliary function that translates this and validates that only correct values are handled?

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,162 @@
...
+    if (WEB3_NODES[networkName] && WEB3_NODES[networkName].length) {
+      if (!this.overriden.web3Nodes) {
+        this.state.web3Providers = Context.resolveWeb3Providers(
+          WEB3_NODES[networkName],
+          this.state.network,
+        );
+      }
+    } else {
+      throw new UnsupportedNetworkError(networkName);
     }
```
brickpop:
```suggestion
    if (!WEB3_NODES[networkName]?.length) {
      throw new UnsupportedNetworkError(networkName);
    }
    if (!this.overriden.web3Nodes) {
      this.state.web3Providers = Context.resolveWeb3Providers(
        WEB3_NODES[networkName],
        this.state.network,
      );
    }
```

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,162 @@
...
-    if (contextParams.graphqlNodes?.length) {
-      this.state.graphql = Context.resolveGraphql(contextParams.graphqlNodes);
+
+    if (GRAPHQL_NODES[networkName] && GRAPHQL_NODES[networkName].length) {
+      if (!this.overriden.graphqlNodes) {
+        this.state.graphql = Context.resolveGraphql(GRAPHQL_NODES[networkName]);
+      }
+    } else {
+      throw new UnsupportedNetworkError(networkName);
+    }
```
brickpop: Move to a unified early error return at the top

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,162 @@
...
+
+    if (IPFS_NODES[networkName] && IPFS_NODES[networkName].length) {
+      if (
+        !this.overriden.ipfsNodes
+      ) {
+        this.state.ipfs = Context.resolveIpfs(IPFS_NODES[networkName]);
+      }
+    } else {
+      throw new UnsupportedNetworkError(networkName);
+    }
```
brickpop: Early return, above

modules/client/src/client-common/context.ts
```diff
@@ -1,116 +1,162 @@
...
+      if (!this.overriden.ensRegistryAddress) {
+        let ensRegistry = LIVE_CONTRACTS[networkName].ensRegistry;
+        if (!ensRegistry) {
+          ensRegistry = this.network.ensAddress;
+        }
+        this.state.ensRegistryAddress = ensRegistry;
+      }
+    } else {
+      throw new UnsupportedNetworkError(networkName);
+    }
```
brickpop: Same

modules/client/src/client-common/interfaces/context.ts
```diff
@@ -30,19 +35,22 @@ export type ContextParams = Web3ContextParams &
...
+  signer: Signer;
+  // always going to exist because i'll default to LIVE_CONTRACTS
+  daoFactoryAddress: string;
+  // may exist depending on the network
+  ensRegistryAddress?: string;
+  // TODO
+  // delete these 2 arent used
   web3Providers: JsonRpcProvider[];
   gasFeeEstimationFactor: number;
```
brickpop: How won't these be used??

modules/client/src/client-common/interfaces/core.ts
brickpop: `OS-377`

modules/client/src/client-common/modules/web3.ts
```diff
@@ -154,51 +90,34 @@ export class Web3Module implements IClientWeb3Core {
...
-      return contract as Contract & T;
-    } else if (signer instanceof Wallet) {
-      return contract.connect(signer.connect(provider)) as Contract & T;
-    }
-
-    return contract.connect(signer) as Contract & T;
+    if (!address || !isAddress(address)) throw new InvalidAddressError();
+    else if (!abi) throw new InvalidContractAbiError();
+    const signer = this.getConnectedSigner();
+    return new Contract(address, abi, signer) as Contract & T;
```
brickpop, line 98: Are we sure that this line of code covers all the `connect` logic that existed before? That is a bit tricky from ethers, everything has to be sort of double connected for things to work properly

modules/client/src/client-common/modules/web3.ts
```diff
@@ -207,23 +126,9 @@ export class Web3Module implements IClientWeb3Core {
...
-      throw new NoPluginRepoRegistry();
-    }
-    return this.pluginRepoRegistryAddress;
-  }
-
-  public getDaoRegistry(): string {
-    if (!this.daoRegistryAddress) {
-      throw new NoDaoRegistry();
-    }
-    return this.daoRegistryAddress;
```
brickpop, line 227: Why removing these?

modules/client/test/integration/client/index.test.ts
brickpop: ping

modules/client/test/unit/context-plugins.test.ts
brickpop: Is it done?

modules/client/test/unit/context-plugins.test.ts
brickpop: Any check we can realistically do? Otherwise, resolve

modules/client/test/unit/context.test.ts
brickpop: Right, but what I like about tests is the fact that you can do a clear A/B test within the same test case. If you have to look 3 different scopes to realize that this is covered, that's hard to follow and make sense of

modules/client/test/unit/context.test.ts
brickpop: ping

modules/common/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aragon/sdk-common",
   "author": "Aragon Association",
-  "version": "1.1.1",
+  "version": "1.2.0",
```
brickpop, line 4: changelog too?

modules/client/src/client-common/interfaces/context.ts
josemarinas: old comment, deleting it

modules/client/src/client-common/interfaces/core.ts
josemarinas: ok, I'm resolving this then

modules/client/src/client-common/modules/web3.ts
josemarinas: - The address is checked
- The abi is checked
- The getConnectedSigner calls this.getSigner that thows if its null, so the signer is checked
- The getConnectedSigner also calls this.getProvider that throws if its null, so the provider is checked too
- Then we checked connected the signer with the provider, using the `getConnectedSigner` returns a signer connected with the provider
- We created the contract with the `provider` and then we connected to the signer, here we provide a connected signer so we can skip that. 

I think we are good

modules/client/src/client-common/constants.ts
josemarinas: Yes, we can do it, but i dont see the point, I mean, this is a private helper function to avoid repeating the same thing over and over

modules/client/src/client-common/constants.ts
josemarinas: Not sure what do you mean here. The deveolpers can pass another node if they want, this was supposed to be our api key for satsuma, thinking about it it does not make sense to hide it, since is fairly public, you can find it bu checking the nerwork tab in app.argaon.org

We can leave the process env in case we want to override it and default to our key

modules/client/test/integration/client/index.test.ts
josemarinas: Sorry I didnt see it, sure we can mock it, totally forgot about that, my bad

modules/client/test/unit/context-plugins.test.ts
josemarinas: yep 

- `Context.network` type changes from `Networkish` to `Network`

modules/client/test/unit/context.test.ts
josemarinas: Im not sure I'm following what you mean. We have 3 tests to test values:

- First we pass an empty value and check that sets the deaults
- Second que pass specific contextParams and check that the context sets them properly
- Third we have a context with some parameters and we set new params and check that they were set properly

You want to group this tests in one?

modules/client/src/client-common/context.ts
josemarinas: I suppose you mean with the provider?

This brings some doubts, if the user has multiple providers and some of them are from `goerli`and others are from `mainnet` do we delete the ones that dont match the network or whats the behaivour here. Plus I believe ethers already complains about this when you create the provider throwing a NETWORK_CHANGED_ ERROR

modules/client/test/unit/context.test.ts
josemarinas: I believe this test cases are covered, theres is a test where we start with an empty context and then we set the network, and there is a test where we specify a network and then we change it

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
   signer: new Wallet("private-key"),
   // Optional on "rinkeby", "arbitrum-rinkeby" or "mumbai"
   // Pass the address of the  `DaoFactory` contract you want to use. You can find it here based on your chain of choice: https://github.com/aragon/core/blob/develop/active_contracts.json
+  // By default we will use the one in the ethers-contract package
```
brickpop:
```suggestion
  // Optional. Leave it empty to use Aragon's DAO Factory contract and claim a dao.eth subdomain
```

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
...
   daoFactoryAddress: "0x1234381072385710239847120734123847123",
+  // Pass the address of the ensRegistry, if its not provided, for mainnet and worli we will use the defualt ones
+  // For polygon and mumbai we will use the ones deployed by ourselves, you can check them here: https://github.com/aragon/core/blob/develop/active_contracts.json
+  ensRegistryAddress: "0x1234381072385710239847120734123847123",
```
brickpop:
```suggestion
  // Optional. Pass the address of the ensRegistry for networks other than Mainnet or Goerli. 
  // It will default to the registry deployed by Aragon. You can check them here: https://github.com/aragon/osx/blob/develop/active_contracts.json
  ensRegistryAddress: "0x1234381072385710239847120734123847123",
```

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
...
   // Choose your Web3 provider: Cloudfare, Infura, Alchemy, etc.
+  // Remember to also change it if you change your network
```
brickpop:
```suggestion
  // Remember to change the list of providers if a different network is selected
```

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
...
   web3Providers: ["https://rpc.ankr.com/eth_goerli"],
+  // You can use our enpoints by leaving this field empty.
+  // They will point to the correct one depending on the network
```
brickpop:
```suggestion
  // Optional. By default, it will use Aragon's provided endpoints. 
  // They will switch depending on the network (production, development)
```

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
...
   ipfsNodes: [
     {
       url: "https://testing-ipfs-0.aragon.network/api/v0",
       headers: { "X-API-KEY": IPFS_API_KEY || "" },
     },
   ],
-  // Don't change this line. This is how we connect your app to the Aragon subgraph.
+  // You can use our enpoints by leaving this field empty.
+  // They will point to the correct one depending on the network
```
brickpop:
```suggestion
  // Optional. By default it will use Aragon's provided endpoints.
  // They will switch depending on the network (production, development)
```

modules/client/examples/index.ts
```diff
@@ -46,16 +46,24 @@ export const contextParams: ContextParams = {
   signer: new Wallet("private-key"),
```
brickpop: Should we have first an example with the very, very minimal working parameters?
Just a signer, the network and a free Web3 endpoint example.

modules/client/src/client-common/constants.ts
brickpop: Such a function being internal doesn't prevent from using clean code practices :)

modules/client/src/client-common/context.ts
brickpop: I had trouble understanding my own question back... 

Yes, I mean having some check to make sure that developers don't just enter a bunch of mixed network URL's.
I guess we won't check them until we try to use a given endpoint, in which case the error users will be getting will be completely different (bad network, contract not found, etc)

modules/client/src/client-common/interfaces/context.ts
```diff
@@ -1,26 +1,31 @@
...
+import { JsonRpcProvider, Network, Networkish } from "@ethersproject/providers";
 import { Client as IpfsClient } from "@aragon/sdk-ipfs";
 import { GraphQLClient } from "graphql-request";
 
 // Context input parameters
 
 type Web3ContextParams = {
-  network: Networkish;
-  signer?: Signer;
+  // Network should be provided
```
brickpop: Could we use these types of comments for struct fields, so that everyone gets code completion with tooltips for free when developing?

```suggestion
  /** Network should be provided */
```

modules/client/src/client-common/modules/web3.ts
brickpop: ping

modules/client/test/unit/context-plugins.test.ts
```diff
@@ -23,33 +30,51 @@ describe("ContextPlugin instances", () => {
...
     };
   });
-  it("Should create an empty context", () => {
-    const context = new ContextPlugin({});
-
-    expect(context).toBeInstanceOf(ContextPlugin);
-  });
   it("Should create an empty context and have default values", () => {
-    const context = new ContextPlugin({});
+    const context = new ContextPlugin();
```
brickpop, line 41: Is it still possible to create a context without a list of web3 endpoints? I thought there were some required params, at least

modules/client/test/unit/context.test.ts
brickpop: Where comes again the question whether an empty context should be possible, or rather having some mandatory parameters

modules/client/src/client-common/context.ts
josemarinas: No, I believe I'll throw way before that, when you create the JsonRpcProvider, If the RPCs and the delcared networks dont match, ethers will start complaining

modules/client/src/client-common/modules/web3.ts
josemarinas: We dont use them in the code so it made no sense to leave them there, besides we are going to have multiple pluginRepos depending on the plugin type

modules/client/test/unit/context-plugins.test.ts
josemarinas: Yes, I didnt make it required because that was the aproach that we used with the signer, if we want the context to crash when empty, I'd do it equally for both cases.

modules/client/test/unit/context.test.ts
josemarinas: Yes, check previous comment, but if we add mandatory params I'd add `Signer` and `web3Providers`

keeping them non mandatory can be useful in some UI cases when you whant to have initialized the client, but you dont have a signer because the user is not logged in. For example aragon APP landing page where it wheries the sdk for showing daos and proposals but the signer and web3providers can be undefined.

brickpop: I see, makes sense. Then, do we have checks that make the Client fail whenever a Context without the required fields is passed?

modules/client/src/client-common/context.ts
brickpop: Do we have any test for that?

modules/client/src/client-common/modules/web3.ts
brickpop: Do the examples reflect this? Maybe the changelog?

modules/client/examples/index.ts
```diff
@@ -46,33 +46,42 @@ export const contextParams: ContextParams = {
...
       url: "https://subgraph.satsuma-prod.com/aragon/core-goerli/api",
     },
   ],
 };
 
+// Bare minimoum context that will use de default values
+const minimalContextParams:ContextParams =  {
+  network: 'mainnet',
+  web3Providers: 'https://example.com/your_mainnet_rpc',
+  signer: new Wallet('private-key')
```
brickpop: Could we use a real (free) endpoint and `Wallet.createRandom()` so that the code can potentially be run?

modules/client/src/client-common/constants.ts
```diff
@@ -21,34 +21,54 @@ export const UNAVAILABLE_PROPOSAL_METADATA: ProposalMetadata = {
...
+
+type GraphqlNetworks = "mainnet" | "goerli" | "polygon" | "mumbai";
+
+const SupportedNetworksToGraphqlNetworks: {
+  [K in SupportedNetworks]: GraphqlNetworks;
+} = {
+  homestead: "mainnet",
+  goerli: "goerli",
+  matic: "polygon",
+  maticmum: "mumbai",
```
brickpop, line 38: Wasn't the value osx-mainnet, osx-goerli... before the last change?

modules/client/test/unit/context.test.ts
josemarinas: What makes sense to you, keeping them optional or making them mandatory?

ATM we do no check for this fields, 1 use case could be the described, since for getting info from daos you dont need neither signer or web3providers

modules/client/src/client-common/constants.ts
josemarinas: Yes but all were preceeded by `osx-`, so I moved to the url constant, I can change itr back if you prefer it

modules/client/examples/index.ts
josemarinas: Sure 👍

modules/client/src/client-common/context.ts
josemarinas: Just checked this, ethers does not autoimatically check this, if we make a call to `getNetwork` and the `expectdNectwork`and the `detectedNetowrk` are different it'll throw

PR #206, opened 2023-04-25
Feature: Add getPlugin function

josemarinas:
## Description

Add getPlugin function

Task: [OS-10](https://aragonassociation.atlassian.net/browse/OS-10)

## Type of change

## Checklist:

[OS-10]: https://aragonassociation.atlassian.net/browse/OS-10?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/interfaces.ts
```diff
@@ -38,6 +41,8 @@ export interface IClientMethods extends IClientCore {
   getDao: (daoAddressOrEns: string) => Promise<DaoDetails | null>;
   /** Retrieves metadata for many daos */
   getDaos: (params: IDaoQueryParams) => Promise<DaoListItem[]>;
+  /** Retrieves metadata for many daos */
```
brickpop:
```suggestion
  /** Retrieves the list of plugins available on the PluginRegistry */
```

modules/client/src/internal/client/methods.ts
```diff
@@ -625,4 +637,62 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
     );
   }
+
+  public async getPlugins({
+    limit = 10,
+    skip = 0,
+    direction = SortDirection.ASC,
+    sortBy = PluginSortBy.SUBDOMAIN,
+    subdomain,
+  }: PluginQueryParams): Promise<PluginRepo[]> {
```
brickpop: Like this, I think we are expecting `{}` even is no params are passed

```suggestion
  public async getPlugins({
    limit = 10,
    skip = 0,
    direction = SortDirection.ASC,
    sortBy = PluginSortBy.SUBDOMAIN,
    subdomain,
  }: PluginQueryParams = {}): Promise<PluginRepo[]> {
```

modules/client/src/internal/client/methods.ts
```diff
@@ -625,4 +637,62 @@ export class ClientMethods extends ClientCore implements IClientMethods {
       ),
     );
   }
+
+  public async getPlugins({
+    limit = 10,
+    skip = 0,
+    direction = SortDirection.ASC,
+    sortBy = PluginSortBy.SUBDOMAIN,
```
brickpop, line 666: Why having a sortBy parameter if only one field is available?

modules/client/src/internal/constants.ts
```diff
@@ -33,3 +42,19 @@ export const UNAVAILABLE_DAO_METADATA: DaoMetadata = {
   description: "(the DAO metadata is not available)",
   links: [],
 };
+export const UNSUPPORTED_RELEASE_METADATA_LINK: PluginRepoReleaseMetadata = {
```
brickpop, line 49: These names are already long, but it's not clear that this applies to a plugin. Probably fine as it is, just pointing out (resolve)

modules/client/src/internal/client/methods.ts
josemarinas: Yes but this does not solve the issue of needing to call it with `{}`. One solution would be having an interface with the optional parameter that calls the function with the mandatory parameter:

```ts
interface Test {
  someFunction: (params?: SomeType) => Promise<SomeTypeResponse>;
}

class Methods implements Test {
  someFunction(params?: SomeType): Promise<SomeTypeResponse> {
    return this.someFunctionImplementation(params || {});
  }

  private someFunctionImplementation(
    { prop1 = "ss", prop2 = "bb" }: SomeType,
  ): Promise<SomeTypeResponse> {
    // DO Stuff
  }
}
```

modules/client/src/internal/client/methods.ts
josemarinas: In the future we may have more than one, like, ratings, recomended, most installed... Thats why I left it there

modules/client/src/internal/client/methods.ts
brickpop: It does solve the problem. If the parameter has a default value `= {}` then you don't need to pass anything. If you don't, `{}` will be used and all parameters are going to be empty or have the default value

josemarinas: Typescript complains, I tested it. Maybe its because this should be done in the interface?

modules/client/src/interfaces.ts
```diff
@@ -524,3 +529,44 @@ export type ContractPermissionWithConditionParams = [
...
+export type PluginRepoReleaseMetadata = {
+  name: string;
+  description: string;
+  images: Object; // TODO specify parameters
+};
+
+export type PluginRepoRelease = {
+  release: number;
+  metadata: PluginRepoReleaseMetadata;
+  builds: number[];
```
brickpop: why `number[]`? Can't we add any other data here?
Having an array like `[1,2, 3, 4, 5, 6, 7]` would seem a bit redundant

modules/client/src/internal/client/methods.ts
```diff
@@ -625,4 +637,75 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+    }
+    const query = QueryPlugins;
+    const params = {
+      where,
+      limit,
+      skip,
+      direction,
+      sortBy,
+    };
+    const name = "plugin repos";
```
brickpop, line 681: why this value and this variable?

modules/client/src/internal/client/methods.ts
```diff
@@ -625,4 +637,75 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+          for (const release of pluginRepo.releases) {
+            let metadata: PluginRepoReleaseMetadata;
+            if (!release.metadata) {
+              metadata = EMPTY_RELEASE_METADATA_LINK;
+            } else {
+              try {
+                const metadataCid = resolveIpfsCid(release.metadata);
+                const stringMetadata = await this.ipfs.fetchString(metadataCid);
+                const resolvedMetadata = JSON.parse(stringMetadata);
+                metadata = resolvedMetadata;
```
brickpop, line 703: Do we want to fetch every release's metadata? Feature-wise it makes total sense, but it also becomes so easy to stage a DOS attack if you pin 20 releases of a dummy plugin with an IPFS CiD pointing to a 50Gb file

josemarinas: I mean this could be done literally with anything that has metadata, any proposal or dao has the same issue because we fetch the metadata. So if we want to protect against it i'd do it in the ipfs module, adding a size limit or similar

modules/client/src/internal/client/methods.ts
josemarinas: iss for the graphql error, by default the graphql request returns a generic error, here we just pass what caused the error

modules/client/src/interfaces.ts
josemarinas: There is not much to add, a release can have multiple builds, but all the fields are common

brickpop: But every build has its own metadata as well. I would personally change this to a field like `lastBuildNumber: 7` and then add a separate functio like `getPlugin()` which did the fetching of all the metadata's (releases and builds)

modules/client/src/internal/client/methods.ts
brickpop: The point is that we can only control our IPFS node, but nothing prevents someone from using Pinata and pinning a 1G file and setting it as the metadata everywhere.

We could limit the amount of bytes that our IPFS proxy will accept but independently, I would only fetch the metadata of the very last release and build on `getPlugins()` and then add a separate `getPlugin()` function, which attempts to fetch more data, only for the given plugin

josemarinas: No, I was refering to the IPFS module in the SDK, the client can know how much data has downloaded for an IPFS request, if it exceeds lets say 100KB, we throw and return the default metadata

modules/client/src/interfaces.ts
josemarinas: Mmm, i didnt know that ecah build has different metadata, I though it was shared between bulds. That messes things a bit, but we can follow the strategy of having a shorter response for `getPlugins` and the full response for `getPlugin`

modules/client/src/interfaces.ts
```diff
@@ -41,8 +41,10 @@ export interface IClientMethods extends IClientCore {
   getDao: (daoAddressOrEns: string) => Promise<DaoDetails | null>;
   /** Retrieves metadata for many daos */
   getDaos: (params: IDaoQueryParams) => Promise<DaoListItem[]>;
+  /** retrieves the plugin details given an address, release and build */
+  getPlugin: (params: GetPluginParams) => Promise<PluginRepo>;
   /** Retrieves the list of plugins available on the PluginRegistry */
-  getPlugins: (params?: PluginQueryParams) => Promise<PluginRepo[]>;
+  getPlugins: (params?: PluginQueryParams) => Promise<PluginRepoListItem[]>;
```
brickpop: `GetPluginParams` vs `PluginQueryParams` seems inconsistent. Is there a reason behind?

modules/client/src/interfaces.ts
```diff
@@ -553,20 +555,57 @@ export type SubgraphPluginRepoListItem = {
...
+  release: {
+    release: number;
+    metadata: string;
+  };
+  pluginRepo: {
+    id: string;
+    subdomain: string;
+  };
+};
+
```
brickpop: I'm starting to wonder if it would make sense to move all the subgraph related (internal) types into a separate file. External developers shouldn't need to be aware of this

modules/client/src/interfaces.ts
```diff
@@ -553,20 +555,57 @@ export type SubgraphPluginRepoListItem = {
...
+  ui: string;
+  change: string;
+  pluginSetupABI: Object;
+};
 
 export type PluginRepoRelease = {
   release: number;
   metadata: PluginRepoReleaseMetadata;
-  builds: number[];
+  latestBuild: number;
```
brickpop: Soft suggestion, no strong feelings

```suggestion
  currentBuild: number;
```

modules/client/src/interfaces.ts
```diff
@@ -553,20 +555,57 @@ export type SubgraphPluginRepoListItem = {
...
 };
+
+export type PluginRepo = {
+  address: string;
+  subdomain: string;
+  buildDetails: {
+    metadata: PluginRepoBuildMetadata;
+    build: number;
+  };
+  releaseDetails: {
```
brickpop: Why `xxxDetails` instead of just `xxx`?

modules/client/src/interfaces.ts
```diff
@@ -553,20 +555,57 @@ export type SubgraphPluginRepoListItem = {
...
 };
 
-export type PluginRepo = {
+export type PluginRepoListItem = {
   address: string;
   subdomain: string;
   releases: PluginRepoRelease[];
 };
+
+export type PluginRepo = {
```
brickpop, line 588: Not sure I follow. A plugin repo contains a list of releases. Each release contains a list of builds. How can the plugin repo contain a release and build at the root?

If we are providing the latest version for convenience, then let's make it explicit:

```ts
  current: {
    release: {
      number,
      metadata
    },
    build: {
      number
      metadata
    }
  }
```

modules/client/src/internal/client/methods.ts
```diff
@@ -692,20 +703,75 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      } catch (err) {
+        releaseMetadata = UNAVAILABLE_RELEASE_METADATA;
+        if (err instanceof InvalidCidError) {
+          releaseMetadata = UNSUPPORTED_RELEASE_METADATA_LINK;
+        }
+      }
+    }
+    // get build metadata
+    let buildMetadata: PluginRepoBuildMetadata;
+    if (!pluginVersion.metadata) {
```
brickpop: How are we using `pluginVersion.release.metadata` above and `pluginVersion.metadata` here?  Shouldn't it be `pluginVersion.build.metadata`?

modules/client/src/internal/utils.ts
```diff
@@ -248,21 +251,39 @@ export function toPluginRepoRelease(
...
+): PluginRepo {
+  return {
+    address: pluginVersion.pluginRepo.id,
+    subdomain: pluginVersion.pluginRepo.subdomain,
+    buildDetails: {
+      build: pluginVersion.build,
+      metadata: buildMetadata,
+    },
+    releaseDetails: {
+      release: pluginVersion.release.release,
```
brickpop: `pluginVersion.build` vs `release: pluginVersion.release.release` isn't this inconsistent?

modules/client/src/internal/client/methods.ts
brickpop: This would require introducing the AbortController and rework the client so we can cancel after having downloaded just 100kb, not after the whole download

Anyway, given the new implementation, I think this can be resolved

modules/client/src/internal/constants.ts
```diff
@@ -33,3 +46,36 @@ export const UNAVAILABLE_DAO_METADATA: DaoMetadata = {
...
+};
+
+export const UNAVAILABLE_RELEASE_METADATA: PluginRepoReleaseMetadata = {
+  name: "(unavailable metadata)",
+  description: "(the release metadata is not available)",
+  images: {},
+};
+
+export const UNSUPPORTED_BUILD_METADATA_LINK: PluginRepoBuildMetadata = {
+  ui: "(unsupported metadata link)",
```
brickpop: What does `ui` mean in the context of a plugin build?

modules/client/src/internal/utils.ts
```diff
@@ -237,6 +245,48 @@ export function toTokenTransfer(transfer: SubgraphTransferListItem): Transfer {
...
+    address: pluginRepo.id,
+    subdomain: pluginRepo.subdomain,
+    current: {
+      build: {
+        metadata: buildMetadata,
+        number: pluginRepo.releases[0].builds[0].build,
+      },
+      release: {
+        metadata: releaseMetadata,
+        number: pluginRepo.releases[0].release,
```
brickpop: It would be nice to use the conditional syntax, just in case:

`obj?.prop?.[0]?.value`

modules/client/src/internal/utils.ts
```diff
@@ -237,6 +245,48 @@ export function toTokenTransfer(transfer: SubgraphTransferListItem): Transfer {
...
+  releaseMetadata: PluginRepoReleaseMetadata,
+  buildMetadata: PluginRepoBuildMetadata,
+): PluginRepo {
+  return {
+    address: pluginRepo.id,
+    subdomain: pluginRepo.subdomain,
+    current: {
+      build: {
+        metadata: buildMetadata,
+        number: pluginRepo.releases[0].builds[0].build,
```
brickpop: Are we doing any assumptions here regarding the content?

modules/client/test/integration/client/methods.test.ts
```diff
@@ -954,6 +959,139 @@ describe("Client", () => {
...
+        );
+        expect(plugin.current.release.number).toBe(1);
+        expect(plugin.current.release.metadata.name).toBe("Name");
+        expect(typeof plugin.current.release.metadata.name).toBe("string");
+        expect(plugin.current.release.metadata.description).toBe("Description");
+        expect(typeof plugin.current.release.metadata.description).toBe(
+          "string",
+        );
+        expect(typeof plugin.current.release.metadata.images).toBe("object");
+      });
```
brickpop, line 1119: No test for getPlugin()?

modules/client/src/internal/constants.ts
josemarinas: I suppose It refers to the url to the url where the ui of the plugin is stored but I'm not sure, I followed the metadata schema of the plugins that are deployed atm

modules/client/src/internal/utils.ts
josemarinas: No we are not the query specifies, the sorting and the limit of both releases and builds so only one release and only one build are returned

modules/client/src/internal/utils.ts
josemarinas: I can add it but this wont be undefined since one plugin hast at least one release and a release has at least one build

modules/client/test/integration/client/methods.test.ts
josemarinas: It is there, line 1029

modules/client/src/interfaces.ts
josemarinas: Now the getplugin only receives a plugin address, so resolving

modules/client/src/interfaces.ts
josemarinas: @mathewmeconry  sugested me a tool a couple of months ago that generates TS types from a graphql schema. Maybe its not a bad a idea to take a look into it

modules/client/src/internal/client/methods.ts
josemarinas: I was because I was queryin `PluginVersion`instead of `PluginRepo`, now its solved

modules/client/src/internal/utils.ts
josemarinas: Nope It followed the `PluginVersion` schema, now its back to the `PluginRepo`one so this is no longer relevant

modules/client/src/internal/client/methods.ts
josemarinas: The new implementation still has the same issue and introducing the AbortController in the `IpfsModule` of the client is not a bad idea but is a bigger problem that I'd address in other PR

modules/client/src/interfaces.ts
brickpop: That's interesting, but this would probably change the type names we already have. Anyway, the short term suggestion is to split the types we create/export from the types that are given to us by the subgraph model

modules/client/src/internal/client/methods.ts
brickpop: Agree. Anyway, I would address this on the IPFS proxy side, rather than on the client.
Let's resolve this issue

modules/client/src/internal/constants.ts
brickpop: Then, I would not enter plan text like this. Rather empty strings so that the UI doesn't attempt to load something strange

modules/client/src/internal/utils.ts
brickpop: Better be safe than sorry, a new subgraph version might potentially alter something

modules/client/src/internal/utils.ts
brickpop: If so, let's add a comment explaining that we assume that there will always be one release and build entry. 
BTW, are we 100% sure that the query will always give us the latest release/build?

modules/client/test/integration/client/methods.test.ts
brickpop: Sorry, with the formatting I missed it.

Sometimes it's OK to add an empty line of spacing, just to make things more readable

modules/client/src/internal/utils.ts
josemarinas: Yes it sorts by release / build in descending order and que only pick the first one, so it should be fine

modules/client/src/interfaces.ts
josemarinas: Okay we can make a task to isolate the subgraph types into its own file. I think we should give it a test in the future because I remember that you can modify the output of the generated types so it maches our naming conventions

modules/client/test/integration/client/methods.test.ts
josemarinas: Sorry, my bad, added the empty lines for readability

brickpop (approved):

PR #207, opened 2023-05-02
Feature: Add plugin uninstallation

josemarinas:
## Description

Task: [OS-391](https://aragonassociation.atlassian.net/browse/OS-391)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

[OS-391]: https://aragonassociation.atlassian.net/browse/OS-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -21,12 +21,16 @@ const client: Client = new Client(context);
 const prepareUninstallationParams: PrepareUninstallationParams = {
   daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
   pluginAddress: "0x2345678901234567890123456789012345678901",
+  uninstallationParams: [[
+    "casa",
+  ]],
 };
 
 // Estimate how much gas the transaction will cost.
-const estimatedGas: GasFeeEstimation = await client.estimation.prepareUninstallation(
-  prepareUninstallationParams,
-);
+const estimatedGas: GasFeeEstimation = await client.estimation
+  .prepareUninstallation(
+    prepareUninstallationParams,
+  );
 console.log({ avg: estimatedGas.average, max: estimatedGas.max });
 
 // Deposit the ERC20 tokens.
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -69,14 +69,12 @@ Returns:
       operation: 1, // revoke
       where: "0x12345...",
       who: "0x23456...",
-      condition: "0x78910...",
       permissionId: "0x1234567890..."
     },
     {
       operation: 1, //REVOKE
       where: "0x3456...",
       who: "0x4567...",
-      condition: "0x5678...",
       permissionId: "0x1234567890..."
     }
   ],
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -22,13 +22,10 @@ const prepareUninstallationParams: PrepareUninstallationParams = {
   daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
   pluginAddress: "0x2345678901234567890123456789012345678901",
   uninstallationParams: [ // Parameters needed by the prepare uninstall abi
-    "test",
-    ["testArrayItem", "testArrayItem2"],
-    {
-      testObjectProp: "testObjectValue",
-    },
+    1234,
+    "0x1234567890123456789012345678901234567890",
   ],
-  uninastallationAbi: ["uint256", "string", "string[]", "object"], // optional, if not specified, the abi will be fetched from the plugin metadata
+  uninstallationAbi: ["uint256", "adress"],
 };
 
 // Estimate how much gas the transaction will cost.
@@ -69,7 +66,7 @@ Returns:
 {
   permissions: [
     {
-      operation: 1,
+      operation: 1, // revoke
       where: "0x12345...",
       who: "0x23456...",
       condition: "0x78910...",
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -67,7 +67,22 @@ Returns:
   txhash: "0xb1c14a49...3e8620b0f5832d61c"
 }
 {
-  permissions: ["0x123...", "0x123..."],
+  permissions: [
+    {
+      operation: 1,
+      where: "0x12345...",
+      who: "0x23456...",
+      condition: "0x78910...",
+      permissionId: [0,10,20...]
+    },
+    {
+      operation: 1,
+      where: "0x3456...",
+      who: "0x4567...",
+      condition: "0x5678...",
+      permissionId: [0,10,20...]
+    }
+  ],
   pluginAddress: "0x1234567890...",
   pluginRepo: "0x1234567890..."
   versionTag: {
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -28,6 +28,7 @@ const prepareUninstallationParams: PrepareUninstallationParams = {
       testObjectProp: "testObjectValue",
     },
   ],
+  uninastallationAbi: ["uint256", "string", "string[]", "object"], // optional, if not specified, the abi will be fetched from the plugin metadata
 };
 
 // Estimate how much gas the transaction will cost.
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -70,14 +70,14 @@ Returns:
       where: "0x12345...",
       who: "0x23456...",
       condition: "0x78910...",
-      permissionId: [0,10,20...]
+      permissionId: "0x1234567890..."
     },
     {
-      operation: 1,
+      operation: 1, //REVOKE
       where: "0x3456...",
       who: "0x4567...",
       condition: "0x5678...",
-      permissionId: [0,10,20...]
+      permissionId: "0x1234567890..."
     }
   ],
   pluginAddress: "0x1234567890...",
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -9,13 +9,9 @@ title: PrepareInstallation
 
 import {
   Client,
-  DaoDepositSteps,
-  DepositParams,
   GasFeeEstimation,
   PrepareUninstallationParams,
   PrepareUninstallationSteps,
-  SetAllowanceSteps,
-  TokenType,
 } from "@aragon/sdk-client";
 import { context } from "../index";
 
@@ -24,10 +20,6 @@ const client: Client = new Client(context);
 
 const prepareUninstallationParams: PrepareUninstallationParams = {
   daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
-  versionTag: {
-    build: 1,
-    release: 1,
-  },
   pluginAddress: "0x2345678901234567890123456789012345678901",
 };
 
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -1,0 +1,78 @@
+/* MARKDOWN
+---
+title: PrepareInstallation
+---
+
+### Prepare Uninstallation of a plugin
+
+*/
+
+import {
+  Client,
+  DaoDepositSteps,
+  DepositParams,
+  GasFeeEstimation,
+  PrepareUninstallationParams,
+  PrepareUninstallationSteps,
+  SetAllowanceSteps,
+  TokenType,
+} from "@aragon/sdk-client";
+import { context } from "../index";
+
+// Instantiate the general purpose client from the Aragon OSx SDK context.
+const client: Client = new Client(context);
+
+const prepareUninstallationParams: PrepareUninstallationParams = {
+  daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
+  versionTag: {
+    build: 1,
+    release: 1,
+  },
+  pluginAddress: "0x2345678901234567890123456789012345678901",
+};
+
+// Estimate how much gas the transaction will cost.
+const estimatedGas: GasFeeEstimation = await client.estimation.prepareUninstallation(
+  prepareUninstallationParams,
+);
+console.log({ avg: estimatedGas.average, max: estimatedGas.max });
+
+// Deposit the ERC20 tokens.
+const steps = client.methods.prepareUninstallation(prepareUninstallationParams);
+for await (const step of steps) {
+  try {
+    switch (step.key) {
+      case PrepareUninstallationSteps.PREPARING:
+        console.log({ txHash: step.txHash });
+        break;
+      case PrepareUninstallationSteps.DONE:
+        console.log({
+          permissions: step.permissions,
+          pluginAddress: step.pluginAddress,
+          pluginRepo: step.pluginRepo,
+          versionTag: step.versionTag,
+        });
+        break;
+    }
+  } catch (err) {
+    console.error(err);
+  }
+}
+
+/* MARKDOWN
+Returns:
+```tsx
+{
+  txhash: "0xb1c14a49...3e8620b0f5832d61c"
+}
+{
+  permissions: ["0x123...", "0x123..."],
+  pluginAddress: "0x1234567890...",
+  pluginRepo: "0x1234567890..."
+  versionTag: {
+    build: 1,
+    release: 1
+  }
+}
+```
+*/
--- a/modules/client/examples/01-client/12-prepare-uninstallation.ts
+++ b/modules/client/examples/01-client/12-prepare-uninstallation.ts
@@ -21,9 +21,13 @@ const client: Client = new Client(context);
 const prepareUninstallationParams: PrepareUninstallationParams = {
   daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
   pluginAddress: "0x2345678901234567890123456789012345678901",
-  uninstallationParams: [[
-    "casa",
-  ]],
+  uninstallationParams: [ // Parameters needed by the prepare uninstall abi
+    "test",
+    ["testArrayItem", "testArrayItem2"],
+    {
+      testObjectProp: "testObjectValue",
+    },
+  ],
 };
 
 // Estimate how much gas the transaction will cost.
--- a/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
@@ -36,7 +36,7 @@ const applyInstallationParams: ApplyInstallationParams = {
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
     operation: PermissionOperationType.GRANT_WITH_CONDITION,
-    permissionId: PermissionIds.EXECUTE_PERMISSION_ID, // hexToBytes("PermissionIds.EXECUTE_PERMISSION_ID")
+    permissionId: PermissionIds.EXECUTE_PERMISSION_ID, // hash("EXECUTE_PERMISSION_ID")
     where: "0x1234567890123456789012345678901234567890",
     who: "0x2345678901234567890123456789012345678901",
   }],
--- a/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
@@ -15,6 +15,7 @@ import {
   Client,
   DaoAction,
   DecodedApplyInstallationParams,
+  PermissionOperationType,
 } from "@aragon/sdk-client";
 import { hexToBytes } from "@aragon/sdk-common";
 import { context } from "../index";
@@ -34,7 +35,7 @@ const applyInstallationParams: ApplyInstallationParams = {
   ],
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
-    operation: 1,
+    operation: PermissionOperationType.GRANT_WITH_CONDITION,
     permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
     where: "0x1234567890123456789012345678901234567890",
     who: "0x2345678901234567890123456789012345678901",
--- a/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-installation.ts
@@ -36,7 +36,7 @@ const applyInstallationParams: ApplyInstallationParams = {
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
     operation: PermissionOperationType.GRANT_WITH_CONDITION,
-    permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
+    permissionId: PermissionIds.EXECUTE_PERMISSION_ID, // hexToBytes("PermissionIds.EXECUTE_PERMISSION_ID")
     where: "0x1234567890123456789012345678901234567890",
     who: "0x2345678901234567890123456789012345678901",
   }],
--- a/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
@@ -90,7 +90,6 @@ Returns:
 { decodedParams:
   {
     permissions: [{
-      condition: "0x1234567890123456789012345678901234567890",
       operation: 1,
       permissionId: Uint8Array([10,20,30]),
       where: "0x1234567890123456789012345678901234567890",
--- a/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
@@ -28,7 +28,7 @@ const client: Client = new Client(context);
 const applyUninstallationParams: ApplyUninstallationParams = {
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
-    operation: PermissionOperationType.GRANT,
+    operation: PermissionOperationType.REVOKE,
     permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
     where: "0x1234567890123456789012345678901234567890",
     who: "0x2345678901234567890123456789012345678901",
--- a/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
+++ b/modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
@@ -27,7 +27,6 @@ const client: Client = new Client(context);
 // This variable contains the values received on the prepareUninstallation() method
 const applyUninstallationParams: ApplyUninstallationParams = {
   permissions: [{
-    condition: "0x1234567890123456789012345678901234567890",
     operation: PermissionOperationType.REVOKE,
     permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
     where: "0x1234567890123456789012345678901234567890",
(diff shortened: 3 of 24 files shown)


modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// Instantiate the general purpose client from the Aragon OSx SDK context.
+const client: Client = new Client(context);
+
+const prepareUninstallationParams: PrepareUninstallationParams = {
+  daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
+  versionTag: {
+    build: 1,
+    release: 1,
+  },
```
brickpop: Why do we need the version tag? Can't it be resolved from the given plugin address or from subgraph?

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,78 @@
...
+/* MARKDOWN
+Returns:
+```tsx
+{
+  txhash: "0xb1c14a49...3e8620b0f5832d61c"
+}
+{
+  permissions: ["0x123...", "0x123..."],
+  pluginAddress: "0x1234567890...",
+  pluginRepo: "0x1234567890..."
```
brickpop, line 82: Do we need to return these 2?

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/* MARKDOWN
+Returns:
+```tsx
+{
+  txhash: "0xb1c14a49...3e8620b0f5832d61c"
+}
+{
+  permissions: ["0x123...", "0x123..."],
```
brickpop: @Rekard0 don't the permissions to revoke come from subgraph? 
@josemarinas There can be many things to undo, not necessarily always revoking. And we also need to know who/where

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,78 @@
...
+  pluginAddress: "0x1234567890...",
+  pluginRepo: "0x1234567890..."
+  versionTag: {
+    build: 1,
+    release: 1
+  }
```
brickpop, line 86: Why returning the tag?

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,78 @@
...
+}
```
brickpop, line 87: We should be returning the uninstallationId

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+import { context } from "../index";
+import { PermissionIds } from "../../dist/interfaces";
+
+// Instantiates an Aragon OSx SDK client.
+const client: Client = new Client(context);
+
+// This variable contains the values received on the prepareUninstallation() method
+const applyUninstallationParams: ApplyUninstallationParams = {
+  permissions: [{
+    condition: "0x1234567890123456789012345678901234567890",
```
brickpop: @Rekard0 @josemarinas is the condition needed to apply an uninstallation?

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+    operation: 1,
```
brickpop: I would use the constant

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+    permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
+    where: "0x1234567890123456789012345678901234567890",
+    who: "0x2345678901234567890123456789012345678901",
+  }],
```
brickpop: If I am not mistaken, applyUninstallation should work pretty much like an installation. There is an uninstallation prepared, and the only thing missing is to apply it on the PSP. 

Not sure why these parameters are still needed here

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: We dont but they are part of the `applyUninstallation` encoder so I dont see a reason for not leaving them there

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: same as pluginRepo and pluginAddress

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: uninstallations are not indexed so there is not an unistallation Id

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
josemarinas: Reviewed with sarkawt they are needed

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -24,17 +20,17 @@ const client: Client = new Client(context);
...
 const prepareUninstallationParams: PrepareUninstallationParams = {
   daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
-  versionTag: {
-    build: 1,
-    release: 1,
-  },
   pluginAddress: "0x2345678901234567890123456789012345678901",
+  uninstallationParams: [[
+    "casa",
+  ]],
```
brickpop: Why an array of arrays?

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -27,7 +28,7 @@ const client: Client = new Client(context);
 const applyUninstallationParams: ApplyUninstallationParams = {
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
-    operation: 1,
+    operation: PermissionOperationType.GRANT,
```
brickpop: are we granting a permission when uninstalling?

modules/client/src/interfaces.ts
```diff
@@ -539,26 +536,31 @@ export type ContractWithdrawParams = [string, string, BigNumber, string];
 
 export type PrepareUninstallationParams = {
   daoAddressOrEns: string;
-  versionTag: VersionTag;
   pluginAddress: string;
   pluginInstallationIndex?: number;
+  uninstallationParams?: any;
```
brickpop: Shouldn't it be a `Uint8Array`? empty by default

modules/client/src/interfaces.ts
```diff
@@ -539,26 +536,31 @@ export type ContractWithdrawParams = [string, string, BigNumber, string];
...
 };
 
 export type SubgraphPluginVersion = {
   release: {
     release: number;
   };
+  metadata: string;
```
brickpop, line 620: Which metadata? The release or the build's?

modules/client/src/internal/client/methods.ts
```diff
@@ -410,17 +410,41 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-      },
-    )[pluginInstallationIndex];
-    if (!filteredInstallation) {
+    const selectedInstallation = iplugin.installations[pluginInstallationIndex];
+    if (!selectedInstallation) {
       throw new InstallationNotFoundError();
     }
+    // check if metadata is defined
+    if (!selectedInstallation.appliedVersion.metadata) {
+      throw new MissingMetadataError();
```
brickpop: In case a plugin has no metadata, I think the SDK should allow for a way to manually specify it

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: Just an example this must be anything that matches the abi on the plugin metadata, from a single value, to an array of values or object

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
josemarinas: true, my bad

modules/client/src/interfaces.ts
josemarinas: no because the encoding is done inside the plugin uninstallation using the abi that comes from the ipfs metadata, the other approach would be asume that the user is going to do de encoding outside of the prepareUninstallation and ask for an Uint8Array

modules/client/src/interfaces.ts
josemarinas: build's, you want me to specify it?

modules/client/src/internal/client/methods.ts
josemarinas: I think this wont be the case, because all plugins deployed with our deploy script have metadata, however one solution would be allowing to pass the pluginUninstallationAbi as an optional parameter the other one is abstracting the encoding functionality from the `prepareUninstallation` function and receive a data parameter

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: No, they come from the pluginUninstallation function, atm subgraph does not store the permissions needed to apply the uninstallation. The `prepareUninstallation` emits a n `UninstallationPrepared` event with all the permissions

brickpop: But in addition to the permission ID's themselves, don't we also need the `who` and the `where`?

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
brickpop: Ping

modules/client/src/interfaces.ts
brickpop: Not sure I follow. Prepare uninstallation params is used by a developer, and in my view the developer should have used an encoder to produce the bytes to pass to the prepare function.

Do you mean that you prefer us doing this instead? My concern is that by defining `any`, developers can just pass anything and they will get little clue from what is going wrong. If there was a more explicit way, we should consider it... since this is a sensitive flow

modules/client/src/interfaces.ts
brickpop: oh wait, that's what comes from subgraph
Adding a comment like `/** build metadata URI */`  before the line, will help us here

modules/client/src/internal/client/methods.ts
brickpop: Our plugins will be fine, but nothing prevents anyone else from publishing a metadata-less plugin. In such a case, a DAO could not uninstall such plugin through the SDK

Adding an optional parameter sounds good. Alternatively, having a specific function receiving the ABI instead of fetching it.

modules/client/src/interfaces.ts
josemarinas: Yeah its going to be hard to be specific here, because the params of the encoder can be literally anything. I dont have a strong opinion on doing the encoding ourselves or leave it to the user, in one case theey call `prepareUninstallation` with the params and in the other you call `getPlugin`to get the ABI, encode the params and you pass them to `prepareUninstallation`. Let me know how you invision it and i'll implement it

modules/client/src/interfaces.ts
josemarinas: 👍

modules/client/src/internal/client/methods.ts
josemarinas: Ok, first lets decide which implementation are we going to do, because if we decide that the user needs to pass the data this optinal parameter wont make sense

modules/client/examples/01-client/12-prepare-uninstallation.ts
brickpop: ping

modules/client/src/interfaces.ts
brickpop: If we're reasonably sure that parameters will have to be an array, I would at least type it like that. For the rest, if we delegate the data checking to the encoding library, it's fine :+1:

modules/client/src/internal/client/methods.ts
brickpop: If the end user facing parameters don't look strange, I would add an optional parameter to receive the ABI in case it was not available (for some reason).

modules/client/src/multisig/interfaces.ts
```diff
@@ -204,4 +204,3 @@ export type SubgraphMultisigVotingSettings = {
   minApprovals: number;
   onlyListed: boolean;
 };
-
```
brickpop, line 207:
```suggestion
};

```

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: We are using the full permission as it is returned from the contracts, we even check them in the tests

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
josemarinas: Yes, the ether package requires an array of `MultiTargetPermissionStruct` defiuned as

```ts

    type MultiTargetPermissionStruct = {
        operation: BigNumberish;
        where: string;
        who: string;
        condition: string;
        permissionId: BytesLike;
    };
```

modules/client/src/multisig/interfaces.ts
josemarinas: idk why it has the wron formatting in gh in mylocal seems ok with an empty line at the end

modules/client/src/internal/client/methods.ts
```diff
@@ -388,6 +402,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      }
+      // get unistallation abi
+      uninstallationAbi = buildMetadata?.pluginSetupABI
+        ?.prepareUninstallation;
+  
+      // TODO remove this when metadata is fixed
+      if (uninstallationAbi.length === 1 && uninstallationAbi[0] === "") {
+        uninstallationAbi = [];
+      }
+      if (!uninstallationAbi) {
```
brickpop: This is never going to be the case. Shouldn't this check be on line 461, before the other `if`?

Like this, if it was `null` execution would fail before we get to this `if`

modules/client/src/multisig/interfaces.ts
brickpop: I'm making the opposite point, there was an empty line, now being removed. Git works better when there's an empty line at the end

modules/client/test/integration/client/decoding.test.ts
```diff
@@ -504,6 +505,77 @@ describe("Client", () => {
...
       );
     });
+    it("Should decode an apply uninstallation action", async () => {
+      const networkSpy = jest.spyOn(JsonRpcProvider.prototype, "network", "get");
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+
+      const applyUninstallationParams: ApplyUninstallationParams = {
+        permissions: [{
+          condition: "0x1234567890123456789012345678901234567890",
```
brickpop: Why do we need a condition?

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -457,5 +458,81 @@ describe("Client", () => {
...
       }
     });
+    it("Should encode an applyUninstallation action", async () => {
+      const networkSpy = jest.spyOn(JsonRpcProvider.prototype, "network", "get");
+      const context = new Context(contextParamsLocalChain);
+      const client = new Client(context);
+
+      const applyUninstallationParams: ApplyUninstallationParams = {
+        permissions: [{
+          condition: "0x1234567890123456789012345678901234567890",
```
brickpop: Same

modules/client/test/integration/client/decoding.test.ts
josemarinas: It is required by the ethers-contracts parameters, they expect a `MultiTargetPermissionStruct` defined as:

```ts
 type MultiTargetPermissionStruct = {
        operation: BigNumberish;
        where: string;
        who: string;
        condition: string;
        permissionId: BytesLike;
    };
```

modules/client/src/multisig/interfaces.ts
josemarinas: yes but on my machine the empty line is still there, idk why is it removed on gh
imagen

modules/client/examples/01-client/12-prepare-uninstallation.ts
Rekard0: > @Rekard0 don't the permissions to revoke come from subgraph? 


No, 
When an event of prepareXYZ happens we index PluginPermission entities inside PluginPreparation entity
when applyXYZ happen we set the PluginPreparation as applied

the same pattern is true for un-installation, somebody need to prepareUninstall, this will tell subgraph what revoke or even grant permission need to be indexed, to be shown on UI or whatever.

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
Rekard0: > @Rekard0 @josemarinas is the condition needed to apply an uninstallation?

yes, but it can be address zero()

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,94 @@
...
+/* MARKDOWN
+Returns:
+```tsx
+{
+  txhash: "0xb1c14a49...3e8620b0f5832d61c"
+}
+{
+  permissions: [
+    {
+      operation: 1,
```
brickpop: I would add a comment to the right, with the constant that `1` refers to

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,94 @@
...
+      where: "0x12345...",
+      who: "0x23456...",
+      condition: "0x78910...",
+      permissionId: [0,10,20...]
```
brickpop: Is that a Uint8Array? I would return it as a hex string, as it is commonly represented elsewhere. There's no data processing that needs to be done with it and this value needs to be potentially compared

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,94 @@
...
+{
+  txhash: "0xb1c14a49...3e8620b0f5832d61c"
+}
+{
+  permissions: [
+    {
+      operation: 1,
+      where: "0x12345...",
+      who: "0x23456...",
+      condition: "0x78910...",
```
brickpop: I'm almost certain that the condition here is not relevant. I would either remove it (make it optional at the type) or worst case scenario display it with a `0x000...` value

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
brickpop: According to applyUninstallation > applyMultiTargetPermissions, the condition won't be used when revoking the permissions:

https://github.com/aragon/osx/blob/527474bb14529f2892e8277f6d7a1ca2da637a55/packages/contracts/src/core/permission/PermissionManager.sol#L173-L177

modules/client/examples/01-client/12-prepare-uninstallation.ts
```diff
@@ -0,0 +1,94 @@
...
+  daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
+  pluginAddress: "0x2345678901234567890123456789012345678901",
+  uninstallationParams: [ // Parameters needed by the prepare uninstall abi
+    "test",
+    ["testArrayItem", "testArrayItem2"],
+    {
+      testObjectProp: "testObjectValue",
+    },
+  ],
+  uninastallationAbi: ["uint256", "string", "string[]", "object"], // optional, if not specified, the abi will be fetched from the plugin metadata
```
brickpop: The ABI is defined in terms of Solidity, there is no such data type called `object`. It should rather be a `tuple()` but this is being defined right now. So I would probably remove the last parameter and just leave a number, and an address.

To another note, the format that we will probably use is not Ether's but rather the standard solidity one (a bit more verbose)

modules/client/examples/01-client/12-prepare-uninstallation.ts
brickpop: ping

modules/client/examples/01-client/12-prepare-uninstallation.ts
brickpop:
```suggestion
```

modules/client/examples/01-client/12-prepare-uninstallation.ts
brickpop: ping

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+import { PermissionIds } from "../../dist/interfaces";
+
+// Instantiates an Aragon OSx SDK client.
+const client: Client = new Client(context);
+
+// This variable contains the values received on the prepareUninstallation() method
+const applyUninstallationParams: ApplyUninstallationParams = {
+  permissions: [{
+    operation: PermissionOperationType.REVOKE,
+    permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
```
brickpop: can't we accept it as a hex string? This is a fixed length value with no data processing applied

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+console.log({ decodedParams });
+
+/* MARKDOWN
+Returns:
+
+```json
+{ decodedParams:
+  {
+    permissions: [{
+      condition: "0x1234567890123456789012345678901234567890",
```
brickpop:
```suggestion
```

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
```diff
@@ -0,0 +1,109 @@
...
+      operation: 1,
```
brickpop: Shouldn't the value correspond to revoke?
I would add a comment on the right side with the constant

modules/client/src/multisig/interfaces.ts
brickpop: Very weird...
image

modules/client/test/integration/client/decoding.test.ts
brickpop:
```suggestion
```

modules/client/test/integration/client/decoding.test.ts
```diff
@@ -504,6 +505,77 @@ describe("Client", () => {
...
+          operation: 1,
```
brickpop: Isn't `1` GRANT? I would use the constant to perform a REVOKE instead

modules/client/test/integration/client/encoding.test.ts
brickpop:
```suggestion
```

modules/client/src/multisig/interfaces.ts
josemarinas: well, i'll commit it and hopefully it fixes

modules/client/test/integration/client/decoding.test.ts
josemarinas: nope, 1 is revoke

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
josemarinas: Yes, the contracts accept `BytesLike` which is basically `Uint8Array | number[] | string`, so we can accept the string for this if you want

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: The contracts will return it if it exists

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: yes, then i'll change the typing to `Uint8Array | string` ?

brickpop: This is a result, we will only be able to return one value, and I think that the permissionId should be a hex string, which is the hash of the permission name

modules/client/examples/01-client/12-prepare-uninstallation.ts
brickpop: Maybe there's a misunderstanding here. One thing is reporting the current status of how permissions are, but the way I see it, the method should be returning the steps to apply for installing. So, for uninstalling these steps should be a `REVOKE` operation, with the `who`, `where` and `permissionId`... which are the only relevant parameters for revoke().

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
brickpop: ok, but I mean about the only-one-correct-way of receiving data on the SDK side (the internal Ethers behaviour should be independent)

modules/client/src/internal/client/methods.ts
```diff
@@ -433,41 +430,8 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-      }
-      // TODO remove this when metadata is fixed
-      if (uninstallationAbi.length === 1 && uninstallationAbi[0] === "") {
-        uninstallationAbi = [];
-      }
-    }
-
     // encode uninstallation params
-    const { uninstallationParams = [] } = params;
+    const { uninstallationParams = [], uninstallationAbi = []} = params;
```
brickpop, line 435: Shouldn't the uninstallationAbi be required? Even if it's an empty array

modules/client/examples/05-encoders-decoders/12-apply-uninstallation.ts
josemarinas: We are generally accepoting `Uint8Array` for all this kind of hex data, we can change to string in this case or Uint8Array | string

modules/client/src/internal/client/methods.ts
josemarinas: then uninstallation params should be required to?

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: 👍 then i'll set the condition to `AddressZero`

modules/client/examples/01-client/12-prepare-uninstallation.ts
josemarinas: okay 👍

modules/client/examples/05-encoders-decoders/12-apply-installation.ts
```diff
@@ -34,8 +35,8 @@ const applyInstallationParams: ApplyInstallationParams = {
   ],
   permissions: [{
     condition: "0x1234567890123456789012345678901234567890",
-    operation: 1,
-    permissionId: hexToBytes(PermissionIds.EXECUTE_PERMISSION_ID),
+    operation: PermissionOperationType.GRANT_WITH_CONDITION,
+    permissionId: PermissionIds.EXECUTE_PERMISSION_ID, // hexToBytes("PermissionIds.EXECUTE_PERMISSION_ID")
```
brickpop:
```suggestion
    permissionId: PermissionIds.EXECUTE_PERMISSION_ID, // hash("EXECUTE_PERMISSION_ID")
```

modules/client/src/internal/utils.ts
```diff
@@ -309,7 +325,7 @@ export function applyInstallatonParamsFromContract(
     permissions: params.permissions.map((permission: any) => {
       return {
         ...permission,
-        permissionId: hexToBytes(permission.permissionId),
+        permissionId: permission.permissionId,
       };
```
brickpop: Can't this `map` simply disappear now?

brickpop (approved):

modules/client/src/internal/utils.ts
josemarinas: yep

brickpop (approved):

PR #211, opened 2023-05-10
Feature: Refactor interfaces into types

josemarinas:
## Description

Changed remaining interfaces to types and put them in a separate file `types.ts`
Moved the supbgraph and contract types from `ìnterfaces.ts`to `internal/types.ts`
Exported everything outside of internal
Moved and reefactor some types to avoid the plugins depending from `@aragon/sdk-client`
Updated tests and examples

Task: [OS-235](https://aragonassociation.atlassian.net/browse/OS-235)

## Type of change

## Checklist:

[OS-235]: https://aragonassociation.atlassian.net/browse/OS-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -75,6 +68,15 @@ import {
...
 import { AddresslistVotingClientEncoding } from "./encoding";
+import {
+  AddresslistVotingPluginPrepareInstallationParams,
+  AddresslistVotingProposal,
+  AddresslistVotingProposalListItem,
+} from "../../types";
+import {
+  SubgraphAddresslistVotingProposal,
+  SubgraphAddresslistVotingProposalListItem,
+} from "../types";
```
brickpop, line 79: Isn't it a bit confusing to have two `types` files like this?
IMO the subgraph only types could be in a dependencies (or similar) file

modules/client/src/client-common/context.ts
```diff
@@ -1,4 +1,4 @@
-import { ContextParams, ContextState } from "./interfaces/context";
+import { ContextParams, ContextState } from "./types/context";
```
brickpop, line 1: don't we have any interface left on the clients side?

modules/client/src/addresslistVoting/internal/client/encoding.ts
```diff
@@ -32,16 +30,16 @@ export class AddresslistVotingClientEncoding extends ClientCore
...
    * @memberof AddresslistVotingClientEncoding
    */
   static getPluginInstallItem(
-    params: IAddresslistVotingPluginInstall,
+    params: AddresslistVotingPluginInstall,
     network: Networkish,
-  ): IPluginInstallItem {
-    const networkName = getNetwork(network).name as SupportedNetworks
+  ): PluginInstallItem {
+    const networkName = getNetwork(network).name as SupportedNetworks;
```
brickpop: Unrelated to this file/change but the type of something like `goerli` would be better written like `SupportedNetwork` rather than in plural.

modules/client/src/types.ts
```diff
@@ -0,0 +1,376 @@
...
+};
+
+const PermissionIds = Object.entries(Permissions).reduce(
+  (acc, [k, v]) => ({ ...acc, [k + "_ID"]: keccak256(toUtf8Bytes(v)) }),
+  {} as { [k: string]: string },
+);
+Object.freeze(Permissions);
+export { Permissions };
+Object.freeze(PermissionIds);
+export { PermissionIds };
```
brickpop: Should this really be part of the types file? 
This feels more like a constant

modules/client/src/addresslistVoting/internal/client/encoding.ts
josemarinas: Yes, I agree, `SupportedNetwork` its a better name. I'll change it

modules/client/src/client-common/context.ts
josemarinas: yes, we have some, but are in the interfaces.ts file

modules/client/src/addresslistVoting/internal/client/methods.ts
josemarinas: We have the internal types that are only used by the sdk, and the ones outside external which are all exported. Maybe moving the subgraph types to its own package makes sense, but we will still have some internal types that we dont want to export

modules/client/src/types.ts
josemarinas: it is true, thank you for pointng it out

brickpop (commented):
Changes look good

brickpop (commented):
Aren't we updating the package.json file with a new version?

brickpop (approved):

PR #215, opened 2023-05-23
Feature: Add generic prepare installation

josemarinas:
## Description

Task: [OS-407]()

## Type of change

## Checklist:

[OS-407]: https://aragonassociation.atlassian.net/browse/OS-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/examples/01-client/13-prepare-installation.ts
```diff
@@ -0,0 +1,72 @@
...
+const client: Client = new Client(context);
+
+const prepareInstallationParams: PrepareInstallationParams = {
+  daoAddressOrEns: "0x1234567890123456789012345678901234567890", // my-dao.dao.eth
+  pluginRepo: "0x2345678901234567890123456789012345678901",
+  installationParams: [ // Parameters needed by the prepare install abi
+    1234,
+    "0x1234567890123456789012345678901234567890",
+  ],
+  installationAbi: ["uint256", "adress"],
```
brickpop: Is this the solidity ABI that we are using?
https://github.com/aragon/simple-storage-example-plugin/blob/main/contracts/release1/build3/build-metadata-R1B3.json#L7-L20

modules/client/examples/01-client/13-prepare-installation.ts
```diff
@@ -0,0 +1,72 @@
...
+      build: 1,
+      release: 1
+    },
+    permissions: [
+      {
+        condition: "0x12345...",
+        who: "0x12345...",
+        where: "0x12345...",
+        operation: 1, // GRANT
+      permissionId: "0x1234567890..."
```
brickpop, line 69: Indentation

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -261,85 +258,35 @@ export class AddresslistVotingClientMethods extends ClientCore
...
-    yield {
-      key: PrepareInstallationStep.DONE,
-      pluginAddress,
+    const {
+      votingMode,
+      supportThreshold,
+      minParticipation,
+      minDuration,
+      minProposerVotingPower,
+    } = votingSettingsToContract(params.settings.votingSettings);
```
brickpop: Shouldn't this simply return the params array to pass below? We are not using any of these elsewhere

modules/client/src/addresslistVoting/internal/constants.ts
```diff
@@ -14,3 +14,8 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
     "removeAddresses",
   ).format("minimal"),
 ];
+
+export const INSTALLATION_ABI: string[] = [
+  "tuple(uint8, uint64, uint64, uint64, uint256)",
+  "address[]",
+];
```
brickpop: Same, the ABI format published on the metadata will look like this:
https://github.com/aragon/simple-storage-example-plugin/blob/main/contracts/release1/build3/build-metadata-R1B3.json#L7-L20

modules/client/src/client-common/types/common.ts
```diff
@@ -87,4 +89,37 @@ export enum TokenType {
...
+
+export type PrepareInstallationParams = {
+  daoAddressOrEns: string;
+  pluginRepo: string;
+  version?: {
+    build: number;
+    release: number;
+  };
+  installationParams?: any[];
+  installationAbi?: string[];
```
brickpop: https://github.com/aragon/simple-storage-example-plugin/blob/main/contracts/release1/build3/build-metadata-R1B3.json#L7-L20

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -256,85 +250,25 @@ export class MultisigClientMethods extends ClientCore
...
-    const parsedLog = pspContractInterface.parseLog(log);
-    const pluginAddress = parsedLog.args["plugin"];
-    const preparedSetupData = parsedLog.args["preparedSetupData"];
-    if (!(pluginAddress || preparedSetupData)) {
-      throw new PluginInstallationPreparationError();
-    }
-    yield {
-      key: PrepareInstallationStep.DONE,
-      pluginAddress,
+    // todo params
```
brickpop, line 250: todo?

modules/client/src/multisig/internal/constants.ts
```diff
@@ -13,3 +13,8 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
     "updateMultisigSettings",
   ).format("minimal"),
 ];
+
+export const INSTALLATION_ABI: string[] = [
+  "address[]",
```
brickpop: Same, the format

modules/client/src/tokenVoting/internal/constants.ts
```diff
@@ -10,3 +10,9 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
   IERC20MintableUpgradeable__factory.createInterface().getFunction("mint")
     .format("minimal"),
 ];
+
+export const INSTALLATION_ABI: string[] = [
+  "tuple(uint8, uint64, uint64, uint64, uint256)",
```
brickpop: ABI format

modules/client/test/integration/client/methods.test.ts
```diff
@@ -393,6 +394,60 @@ describe("Client", () => {
...
+          Promise.resolve({
+            name: "goerli",
+            chainId: 31337,
+          }),
+        );
+        const steps = client.methods.prepareInstallation(
+          {
+            daoAddressOrEns: dao,
+            pluginRepo: deployment.multisigRepo.address,
+            installationAbi: [ "address[]", "tuple(bool, uint16)"],
```
brickpop: the format

modules/client/examples/01-client/13-prepare-installation.ts
josemarinas: the problem with that abi is that is not compatible with `Array<ParamType | string>`which is accepted by the `defaultAbiEncoder`, so either we accept it as a string or we accept it with the metadata format and we parse it internally to the string format

```ts
export class ParamType {

    // The local name of the parameter (of null if unbound)
    readonly name: string;

    // The fully qualified type (e.g. "address", "tuple(address)", "uint256[3][]"
    readonly type: string;

    // The base type (e.g. "address", "tuple", "array")
    readonly baseType: string;

    // Indexable Paramters ONLY (otherwise null)
    readonly indexed: boolean;

    // Tuples ONLY: (otherwise null)
    //  - sub-components
    readonly components: Array<ParamType>;

    // Arrays ONLY: (otherwise null)
    //  - length of the array (-1 for dynamic length)
    //  - child type
    readonly arrayLength: number;
    readonly arrayChildren: ParamType;

    readonly _isParamType: boolean;
```

brickpop: Still using the other ABI format here.

I would actually declare this value as a separate `const`, to illustrate that this ABI comes from the plugin metadata:

```
// Obtained from the plugin metadata
const installationAbi = [ ... ];
```

modules/client/src/addresslistVoting/internal/client/encoding.ts
```diff
@@ -43,27 +46,12 @@ export class AddresslistVotingClientEncoding extends ClientCore
...
-      ],
-      [
-        [
-          votingMode,
-          supportThreshold,
-          minParticipation,
-          minDuration,
-          minProposerVotingPower,
-        ],
+        Object.values(
```
brickpop: Do we have the certainty that object.values will return us the values in the intended order? in JS there is no guarantee about the order of a JSON object's keys

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -245,81 +243,23 @@ export class AddresslistVotingClientMethods extends ClientCore
...
+      daoAddressOrEns: params.daoAddressOrEns,
       pluginRepo: LIVE_CONTRACTS[networkName].addresslistVotingRepo,
-      versionTag: versionTag!,
-      permissions: preparedSetupData.permissions,
-      helpers: preparedSetupData.helpers,
-    };
+      version: params.versionTag,
+      installationAbi: INSTALLATION_ABI,
+      installationParams: [
+        Object.values(
```
brickpop: Same question as above

modules/client/src/addresslistVoting/internal/constants.ts
```diff
@@ -14,3 +15,54 @@ export const AVAILABLE_FUNCTION_SIGNATURES: string[] = [
     "removeAddresses",
   ).format("minimal"),
 ];
+
+export const INSTALLATION_ABI: MetadataAbiInput[] = [
```
brickpop, line 19: Is there a way to make this be imported from the plugin metadata?
I guess that the npm package doesn't export that?

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +146,135 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+): string[] {
+  return inputs.map((input) => {
+    if (input.type.startsWith("tuple")) {
+      const tupleResult = getNamedTypesFromMetadata(input.components).join(
+        ", ",
+      );
+
+      let tupleString = `tuple(${tupleResult})`;
+
+      if (input.type.endsWith("[]")) {
```
brickpop: Would this work if something has a matrix? `[][]`

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +146,135 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+  return {
+    helpersHash: params.helpersHash,
+    permissions: params.permissions,
+    versionTag: params.pluginSetupRef.versionTag,
+    pluginAddress: params.plugin,
+    pluginRepo: params.pluginSetupRef.pluginSetupRepo,
+  };
+}
+
+export function getNamedTypesFromMetadata(
```
brickpop: Do we have a clearer name? 
- `toStringAbi(jsonAbi)`
- `jsonAbiToNamedTypes`
- ...?

modules/client/src/multisig/internal/client/methods.ts
brickpop: what is still to do?

modules/client/src/addresslistVoting/internal/client/encoding.ts
josemarinas: didnt think about the order, my. bad i'll try to implement it in other way then

modules/client/src/addresslistVoting/internal/constants.ts
josemarinas: That would be the best solution, atm the `osx-ethers` does not provide this constant, but last publish was 2 months ago

modules/client/src/client-common/utils.ts
josemarinas: I'm not sure if this can ever happen, because in the case of a tuple[] you would have components, and then you would have another tuple[] inside the components of this tuple and this recursive case is covered, but Im not sure if `tuple[][]` is a supported type. Since @Michael-A-Heuer wrote this maybe he can give us the response to this

modules/client/src/client-common/utils.ts
josemarinas: I used the name used in the contracts, can be changed to one that makes more sense in the SDK, for me `toStringAbi` makes more sense in this context since we have functions that use the same pattern `toXXX`

So if you are ok with this one i'll prceed and change it

modules/client/src/addresslistVoting/internal/constants.ts
brickpop: OS-486 created

modules/client/src/client-common/utils.ts
heueristik: I haven't considered nested arrays.
Check out the tests. I would include the tests and add this case. https://github.com/aragon/simple-storage-example-plugin/blob/9f222b3d9606b44000b4f22a7901cb5b6ad8b356/test/helpers/types.ts#L39

This method is probably a candidate for the tooling package we want to have.

brickpop: OS-513 created to tackle this

Rekard0 (approved):

PR #217, opened 2023-05-24
Feature: Add prevention of failing proposals

josemarinas:
## Description

Task: [OS-189]()

## Type of change

## Checklist:

[OS-189]: https://aragonassociation.atlassian.net/browse/OS-189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -94,6 +96,10 @@ export class AddresslistVotingClientMethods extends ClientCore
   ): AsyncGenerator<ProposalCreationStepValue> {
     const signer = this.web3.getConnectedSigner();
 
+    if (isFailingProposal(params.actions || [])) {
```
brickpop: I would move the default value `|| []` to the function itself, not here

modules/client/src/client-common/constants.ts
```diff
@@ -138,3 +143,24 @@ export const LIVE_CONTRACTS: { [K in SupportedNetwork]: NetworkDeployment } = {
     ensRegistry: activeContractsList.polygon.ENSRegistry,
   },
 };
+
+export const FAILING_PROPOSAL_AVAILABLE_FUNCTION_SIGNATURES = [
+  Multisig__factory.createInterface().getFunction("addAddresses")
```
brickpop: Refactor the repeated `Multisig__factory.createInterface()` into a local constant

modules/client/src/client-common/constants.ts
```diff
@@ -138,3 +143,24 @@ export const LIVE_CONTRACTS: { [K in SupportedNetwork]: NetworkDeployment } = {
     ensRegistry: activeContractsList.polygon.ENSRegistry,
   },
 };
+
+export const FAILING_PROPOSAL_AVAILABLE_FUNCTION_SIGNATURES = [
```
brickpop: Shouldn't this be on the `internal` side? only exporting the function but not this

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -89,6 +91,10 @@ export class MultisigClientMethods extends ClientCore
   ): AsyncGenerator<ProposalCreationStepValue> {
     const signer = this.web3.getConnectedSigner();
 
+    if (isFailingProposal(params.actions || [])) {
+      throw new InvalidActionsOrderError();
```
brickpop: Suggested name

```suggestion
      throw new AlwaysFailingProposalError();
```

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +126,50 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
       throw new InvalidVotingModeError();
   }
 }
+
+export function isFailingProposal(actions: DaoAction[]): boolean {
+  const functionNames: string[] = [];
+  // store the function names of the actions
+  for (const action of actions) {
```
brickpop: Why not doing an `actions.map`?

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +126,50 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+    try {
+      const fragment = getFunctionFragment(
+        action.data,
+        FAILING_PROPOSAL_AVAILABLE_FUNCTION_SIGNATURES,
+      );
+      console.log(fragment);
```
brickpop:
```suggestion
```

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +126,50 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+      functionNames.push(
+        fragment.name,
+      );
+    } catch (e) {
+      console.log(e);
```
brickpop: Why?

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -76,3 +121,40 @@ describe("client-common utils", () => {
     });
   });
 });
+
+describe("Compute failing proposals", () => {
+  it("Should return true because the proposal changes the min approvals and then updates the addresses", async () => {
```
brickpop:
```suggestion
describe("Detect failing proposals", () => {
  it("isFailingProposal should return true because the proposal changes the min approvals and then updates the addresses", async () => {
```

modules/client/src/client-common/utils.ts
```diff
@@ -121,3 +126,50 @@ export function votingModeFromContracts(votingMode: number): VotingMode {
...
+    }
+  }
+  for (const [i, functionName] of functionNames.entries()) {
```
brickpop, line 141: What would happen if updateSettings was unchanged and an address was removed/added (while settings being still valid)?

Are we reproducing the exact same logic that would fail on the plugins side? Shouldn't we somehow check the current minApprovals and the new one?

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -76,3 +121,40 @@ describe("client-common utils", () => {
...
+    const actions: DaoAction[] = [
+      UPDATE_MULTISIG_SETTINGS_ACTION,
+      ADD_ADDRESSES_ACTION,
+    ];
+    expect(isFailingProposal(actions)).toBe(
+      true,
+    );
+  });
+  it("Should return true because the proposal removes addresses and then updates the settings", async () => {
```
brickpop:
```suggestion
  it("isFailingProposal should return true because the proposal removes addresses and then updates the settings", async () => {
```

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -76,3 +121,40 @@ describe("client-common utils", () => {
...
+    const actions: DaoAction[] = [
+      REMOVE_ADDRESSES_ACTION,
+      UPDATE_MULTISIG_SETTINGS_ACTION,
+    ];
+    expect(isFailingProposal(actions)).toBe(
+      true,
+    );
+  });
+  it("Should return false because the proposal adds addresses, changes the min approvals and then removes addresses", async () => {
```
brickpop:
```suggestion
  it("isFailingProposal should return false because the proposal adds addresses, changes the min approvals and then removes addresses", async () => {
```

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -76,3 +121,40 @@ describe("client-common utils", () => {
...
+    expect(isFailingProposal(actions)).toBe(
+      false,
+    );
+  });
+  it("Should return false because ther is no actions", async () => {
+    const actions: DaoAction[] = [];
+    expect(isFailingProposal(actions)).toBe(
+      false,
+    );
+  });
```
brickpop, line 180: I would also add some other random actions and see that it still returns `false`, even with an add/remove in place

modules/common/src/errors.ts
```diff
@@ -216,3 +216,8 @@ export class DepositNativeTokenError extends SdkError {
     super("Please, use the token's transfer function directly", cause);
   }
 }
```
brickpop, line 218: package.json version bump? Maybe after merging other PR's

modules/client/src/client-common/utils.ts
josemarinas: nope, that will mean is unvalid with the current implementation, but we cannot know how many addresses are added / removed without decoding the data

modules/client/src/client-common/utils.ts
josemarinas: The linter complais if there is a map that does not return something in one of it iterations, this happens here when the function is not on the list `FAILING_PROPOSAL_AVAILABLE_FUNCTION_SIGNATURES`

modules/client/src/client-common/utils.ts
josemarinas: Because if the function signature is not in the available ones it'll throw

modules/client/src/client-common/utils.ts
brickpop: Then we should to a filter+map:
`actions.filter(item => ...).map(item => ...)`

modules/client/src/client-common/utils.ts
brickpop: If a given action is unknown to us, that's not an error. We should only be printing exceptionsl situations if they are crucial to prevent an error or to help diagnosing what went wrong. 
If the `catch` should simply continue iterating, then I would leave `catch() {}`

modules/client/src/client-common/utils.ts
brickpop: And what if there was some other setting (non-wallet related) that was changed? I still believe we should inspect a bit deeper what's actually changing, but I may be missing context

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -44,6 +44,18 @@ const REMOVE_ADDRESSES_ACTION = {
     ),
   ),
 };
+const UPGRADE_TO_ACTION = {
```
brickpop, line 47: Shouldn't this encoding receive any parameters? Or they are irrelevant here?

modules/client/test/unit/client-common/utils.test.ts
```diff
@@ -141,17 +153,26 @@ describe("Compute failing proposals", () => {
...
+      REMOVE_ADDRESSES_ACTION,
+      REMOVE_ADDRESSES_ACTION,
       REMOVE_ADDRESSES_ACTION,
     ];
     expect(isFailingProposal(actions)).toBe(
       false,
     );
   });
-  it("Should return false because ther is no actions", async () => {
+  it("Should return false because there is no actions", async () => {
```
brickpop:
```suggestion
  it("Should return false because there are no actions", async () => {
```

modules/client/src/client-common/utils.ts
josemarinas: I mean there is not much more to do that can break a proposal, the whole reasoning behind this was avoiding proposals that increase or reduce the number of adddresses without changing the settings leading to a bricked dao ora proposal creation error. We can go deeper and decode the data and check if the settings would make sense instead of just checking the order and name functions

modules/client/test/unit/client-common/utils.test.ts
josemarinas: irrelevant here, since we are only checking the name of the function what we need are the first 8 chars of the encoding, the rest does not matter to us. Even if we decode the actions it wouldnt matter because we will only decode the `updateSettings`, `addAddresses` and `removeAdddresses`actions

modules/client/src/client-common/utils.ts
brickpop: As long as we are sure that we are not giving any false positives, then let's go with it and improve over time.
(resolve if so)

brickpop (approved):

modules/common/src/errors.ts
brickpop: PR 215 is already bumping this. If they are published together, then we can resolve this and merge

brickpop (approved):

PR #218, opened 2023-05-25
Feature: Add getNetworkName to web3

josemarinas:
## Description

Encapsulate the getting the network name in the web3 methods

Task: [OS-315]()

## Type of change

## Checklist:

[OS-315]: https://aragonassociation.atlassian.net/browse/OS-315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -14,7 +14,7 @@ TEMPLATE:
 -->
 ## [UPCOMING]
 ## Changed
-- Errors now throw caouse
+- Errors now report the cause more clearly
 - Internal refactor of signer and network
 ## [1.8.1]
 ## Fixes
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -13,6 +13,10 @@ TEMPLATE:
 
 -->
 ## [UPCOMING]
+## Changed
+- Errors now throw caouse
+- Internal refactor of signer and network
+## [1.8.1]
 ## Fixes
 - List all the plugins installed on a DAO by using an updated subgraph query
 ## [1.8.0]
--- a/modules/client/src/addresslistVoting/internal/client/methods.ts
+++ b/modules/client/src/addresslistVoting/internal/client/methods.ts
@@ -14,7 +14,6 @@ import {
   PluginInstallationPreparationError,
   ProposalCreationError,
   resolveIpfsCid,
-  UnsupportedNetworkError,
 } from "@aragon/sdk-common";
 import { isAddress } from "@ethersproject/address";
 import { IAddresslistVotingClientMethods } from "../../interfaces";
@@ -36,8 +35,6 @@ import {
   SortDirection,
   SubgraphMembers,
   SubgraphVotingSettings,
-  SupportedNetwork,
-  SupportedNetworksArray,
   VersionTag,
   VoteProposalParams,
   VoteProposalStep,
@@ -246,12 +243,7 @@ export class AddresslistVotingClientMethods extends ClientCore
     params: AddresslistVotingPluginPrepareInstallationParams,
   ): AsyncGenerator<PrepareInstallationStepValue> {
     const signer = this.web3.getConnectedSigner();
-
-    const network = await this.web3.getProvider().getNetwork();
-    const networkName = network.name as SupportedNetwork;
-    if (!SupportedNetworksArray.includes(networkName)) {
-      throw new UnsupportedNetworkError(networkName);
-    }
+    const networkName = this.web3.getNetworkName();
     // connect to psp contract
     const pspContract = PluginSetupProcessor__factory.connect(
       LIVE_CONTRACTS[networkName].pluginSetupProcessor,
--- a/modules/client/src/client-common/constants.ts
+++ b/modules/client/src/client-common/constants.ts
@@ -32,17 +32,21 @@ type GraphqlNetworks = "mainnet" | "goerli" | "polygon" | "mumbai";
 const SupportedNetworksToGraphqlNetworks: {
   [K in SupportedNetwork]: GraphqlNetworks;
 } = {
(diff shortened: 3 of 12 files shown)


modules/client/src/client-common/interfaces.ts
```diff
@@ -5,13 +5,14 @@ import { Contract, ContractInterface } from "@ethersproject/contracts";
...
 import { GraphQLClient } from "graphql-request";
-import { GasFeeEstimation } from "./types/common";
+import { GasFeeEstimation, SupportedNetwork } from "./types/common";
 
 export interface IClientWeb3Core {
   shiftProvider: () => void;
   getSigner: () => Signer;
   getConnectedSigner: () => Signer;
   getProvider: () => JsonRpcProvider;
+  getNetworkName: () => SupportedNetwork;
```
brickpop, line 15: Would it make sense to turn these values into an `enum`? The string values could still be the same, but people would be dealing with `SupportedNetworks.MUMBAI` instead of `maticmum`

modules/client/src/internal/client/methods.ts
```diff
@@ -393,8 +392,7 @@ export class ClientMethods extends ClientCore implements IClientMethods {
     params: PrepareUninstallationParams,
   ): AsyncGenerator<PrepareUninstallationStepValue> {
     const signer = this.web3.getConnectedSigner();
-    const provider = this.web3.getProvider();
-    const networkName = (await provider.getNetwork()).name as SupportedNetwork;
+    const networkName = this.web3.getNetworkName();
```
brickpop, line 395: Why no await now?  or why did we have `await provider.getNetwork()).name` before?

modules/client/src/addresslistVoting/internal/client/methods.ts
```diff
@@ -246,12 +243,7 @@ export class AddresslistVotingClientMethods extends ClientCore
     params: AddresslistVotingPluginPrepareInstallationParams,
   ): AsyncGenerator<PrepareInstallationStepValue> {
     const signer = this.web3.getConnectedSigner();
-
-    const network = await this.web3.getProvider().getNetwork();
-    const networkName = network.name as SupportedNetwork;
-    if (!SupportedNetworksArray.includes(networkName)) {
-      throw new UnsupportedNetworkError(networkName);
-    }
+    const networkName = this.web3.getNetworkName();
```
brickpop, line 246: Why no await? Or why `await this.web3.getProvider().getNetwork()` before?

modules/client/src/internal/client/methods.ts
josemarinas: Because we network name without consulting the provider because its already in the context

modules/client/src/addresslistVoting/internal/client/methods.ts
josemarinas: Same, we didnt had a way of accessing the network in the context so the only way was through the provider, now we have a way without async/await

modules/client/src/client-common/interfaces.ts
josemarinas: Yep that would make sense and will cover us in case we upgrade to ethers 6 and the names change again

modules/client/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ TEMPLATE:
 
 -->
 ## [UPCOMING]
+## Changed
+- Errors now throw caouse
```
```suggestion
- Errors now report the cause more clearly
```

modules/client/src/client-common/constants.ts
```diff
@@ -39,10 +39,10 @@ const SupportedNetworksToGraphqlNetworks: {
...
 
 export const GRAPHQL_NODES: { [K in SupportedNetwork]: { url: string }[] } = {
-  homestead: [{ url: getGraphqlNode("homestead") }],
-  goerli: [{ url: getGraphqlNode("goerli") }],
-  matic: [{ url: getGraphqlNode("matic") }],
-  maticmum: [{ url: getGraphqlNode("maticmum") }],
+  homestead: [{ url: getGraphqlNode(SupportedNetwork.MAINNET) }],
+  goerli: [{ url: getGraphqlNode(SupportedNetwork.GOERLI) }],
+  matic: [{ url: getGraphqlNode(SupportedNetwork.POLYGON) }],
+  maticmum: [{ url: getGraphqlNode(SupportedNetwork.MUMBAI) }],
```
brickpop:
```suggestion
  [SupportedNetwork.MAINNET]: [{ url: getGraphqlNode(SupportedNetwork.MAINNET) }],
  [SupportedNetwork.GOERLI]: [{ url: getGraphqlNode(SupportedNetwork.GOERLI) }],
  SupportedNetwork.POLYGON: [{ url: getGraphqlNode(SupportedNetwork.POLYGON) }],
  SupportedNetwork.MUMBAI: [{ url: getGraphqlNode(SupportedNetwork.MUMBAI) }],
```

modules/client/src/client-common/types/common.ts
```diff
@@ -1,10 +1,13 @@
...
-  "maticmum",
-] as const;
-export type SupportedNetwork = typeof SupportedNetworksArray[number];
+export enum SupportedNetwork {
+  MAINNET = "homestead",
+  GOERLI = "goerli",
+  POLYGON = "matic",
+  MUMBAI = "maticmum",
+}
+
```
brickpop, line 7: Will there be any difference in TypeScript checking? Will the strict "homestead" (etc) values remain strictly checked?

josemarinas: Yes the main difference is that if you want to use the sting values I believe you need to use `"homestead" as SupportedNetwork`

brickpop (approved):

PR #219, opened 2023-05-25
Feature: Add potentiallyExecutable property and update the status definition

josemarinas:
## Description

Fix statuses for early executable proposals

Task: [OS-446](https://aragonassociation.atlassian.net/browse/OS-446)

## Type of change

## Checklist:

[OS-446]: https://aragonassociation.atlassian.net/browse/OS-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/client-common/utils.ts
```diff
@@ -41,12 +41,12 @@ export function computeProposalStatus(
...
     return ProposalStatus.PENDING;
   }
-  if (proposal.potentiallyExecutable || proposal.earlyExecutable) {
-    return ProposalStatus.SUCCEEDED;
-  }
   if (endDate >= now) {
     return ProposalStatus.ACTIVE;
   }
+  if (proposal.potentiallyExecutable || proposal.earlyExecutable) {
+    return ProposalStatus.SUCCEEDED;
```
brickpop, line 48: what if potentiallyExecutable is true but the endDate is still in the future and early executable is false?

brickpop: And what if `endDate > now` (the above condition) but the proposal has been approved? We shouldn't return `ACTIVE`

modules/client/src/multisig/internal/utils.ts
```diff
@@ -67,6 +67,7 @@ export function toMultisigProposal(
     approvals: proposal.approvers.map(
       (approver) => approver.id.slice(0, 42),
     ),
+    potentiallyExecutable: proposal.potentiallyExecutable,
```
brickpop, line 70: Why do we need to export this parameter? What cannot be done using the existing fields?

brickpop:
Closing since this alters the expected logic of the SDK itself and other users would get an unexpected behavior with these changes.

PR #226, opened 2023-06-19
Featrue: Add `initializeFrom` encoders and decoders

josemarinas:
## Description

Task: [OS-512](https://aragonassociation.atlassian.net/browse/OS-512)

## Type of change

## Checklist:

[OS-512]: https://aragonassociation.atlassian.net/browse/OS-512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/examples/01-client/01-create-dao.ts
```diff
@@ -13,8 +13,8 @@ import {
   CreateDaoParams,
   DaoCreationSteps,
   DaoMetadata,
-  TokenVotingPluginInstall,
   TokenVotingClient,
+  TokenVotingPluginInstall,
```
heueristik, line 17: Since a lot of imports sorting is happening:

How about using `"@trivago/prettier-plugin-sort-imports": "^4.1.1",` (like we do here https://github.com/aragon/osx-plugin-template-hardhat/blob/bc914889cbd286b14d4c4a8a48f870a00ed8526e/package.json#L21) in another PR?

modules/client/examples/05-encoders-decoders/14-initialize-from.ts
```diff
@@ -0,0 +1,79 @@
+/* MARKDOWN
+---
+title: Apply Uninstallation
```
heueristik: `initializeFrom` will be called on an after an update, not an uninstallation. 
Inside `PluginSetupProcessor` update will call the `upgradeToAndCall` (here https://github.com/aragon/osx/blob/873cf252ec92e5f54662d0364e141e7b16348217/packages/contracts/src/framework/plugin/setup/PluginSetupProcessor.sol#L531, which then goes here https://github.com/aragon/osx/blob/873cf252ec92e5f54662d0364e141e7b16348217/packages/contracts/src/framework/plugin/setup/PluginSetupProcessor.sol#L671C20-L671C20 if `initData` is provided).

Accordingly, this example must be adapted to show an plugin update.

heueristik (changes requested):
Please use a plugin update as an example for when initializeFrom is used

modules/client/examples/01-client/01-create-dao.ts
josemarinas: We are using deno atm, and I ran a denofmt and made some changes, if we want to change to prettier or start using this for sorting I would say to leave it for another PR

heueristik (approved):

heueristik (dismissed):

heueristik (dismissed):

heueristik (approved):

PR #250, opened 2023-07-10
Fix: Proposal status and filters for multisig

josemarinas:
## Description

The subgraph sets potentiallyExecutable = true for any multisig proposal where minApproval is 1. So the filters turn up wrong on such cases on this line https://github.com/aragon/sdk/blob/develop/modules/client/src/client-common/utils.ts#L80 (for multisig with minApproval = 1, the defeated condition will be wrong) Also for "Succeeded" in multisig, if the proposal goes past the proposal end date, it should be considered as "Defeated".

An example DAO to check this https://staging-app.aragon.org/#/daos/polygon/0x517b7b401bb0b2c9af9663d0bf7b3688b8cb4fa2/governance

Task ID: [OS-561](https://aragonassociation.atlassian.net/browse/OS-561)

## Type of change

## Checklist:

[OS-561]: https://aragonassociation.atlassian.net/browse/OS-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/src/client-common/utils.ts
+++ b/modules/client/src/client-common/utils.ts
@@ -76,13 +76,12 @@ export function computeProposalStatusFilter(
       where = { executed: true };
       break;
     case ProposalStatus.SUCCEEDED:
-      where = { potentiallyExecutable: true };
+      where = { potentiallyExecutable: true, endDate_lt: now };
       break;
     case ProposalStatus.DEFEATED:
       where = {
         endDate_lt: now,
         executed: false,
-        potentiallyExecutable: false,
       };
       break;
     default:
--- a/modules/client/src/client-common/utils.ts
+++ b/modules/client/src/client-common/utils.ts
@@ -49,8 +49,8 @@ export function computeProposalStatus(
     return ProposalStatus.PENDING;
   }
   if (
-    (proposal.potentiallyExecutable || proposal.earlyExecutable) &&
-    endDate >= now
+    (now > endDate || proposal.earlyExecutable) &&
+    proposal.potentiallyExecutable
   ) {
     return ProposalStatus.SUCCEEDED;
   }
@@ -67,21 +67,22 @@ export function computeProposalStatusFilter(
   const now = Math.round(new Date().getTime() / 1000).toString();
   switch (status) {
     case ProposalStatus.PENDING:
-      where = { startDate_gte: now };
+      where = { startDate_gte: now, potentiallyExecutable: false, executed: false };
       break;
     case ProposalStatus.ACTIVE:
-      where = { startDate_lt: now, endDate_gte: now, executed: false };
+      where = { startDate_lt: now, endDate_gte: now, executed: false, potentiallyExecutable: false };
       break;
     case ProposalStatus.EXECUTED:
       where = { executed: true };
       break;
     case ProposalStatus.SUCCEEDED:
-      where = { potentiallyExecutable: true, endDate_lt: now };
+      where = { potentiallyExecutable: true };
       break;
     case ProposalStatus.DEFEATED:
       where = {
         endDate_lt: now,
         executed: false,
+        potentiallyExecutable: false,
       };
       break;
     default:
--- a/modules/client/src/client-common/utils.ts
+++ b/modules/client/src/client-common/utils.ts
@@ -70,7 +70,7 @@ export function computeProposalStatusFilter(
       where = { startDate_gte: now, potentiallyExecutable: false, executed: false };
       break;
     case ProposalStatus.ACTIVE:
-      where = { startDate_lt: now, endDate_gte: now, executed: false, potentiallyExecutable: false };
+      where = { startDate_lt: now, endDate_gte: now, executed: false };
       break;
     case ProposalStatus.EXECUTED:
       where = { executed: true };
--- a/modules/client/src/client-common/utils.ts
+++ b/modules/client/src/client-common/utils.ts
@@ -48,7 +48,10 @@ export function computeProposalStatus(
   if (startDate >= now) {
     return ProposalStatus.PENDING;
   }
-  if (proposal.potentiallyExecutable || proposal.earlyExecutable) {
+  if (
+    (proposal.potentiallyExecutable || proposal.earlyExecutable) &&
+    endDate >= now
+  ) {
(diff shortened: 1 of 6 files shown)


modules/client/src/client-common/utils.ts
```diff
@@ -48,7 +48,10 @@ export function computeProposalStatus(
...
     return ProposalStatus.PENDING;
   }
-  if (proposal.potentiallyExecutable || proposal.earlyExecutable) {
+  if (
+    (proposal.potentiallyExecutable || proposal.earlyExecutable) &&
+    endDate >= now
+  ) {
     return ProposalStatus.SUCCEEDED;
   }
   if (endDate >= now) {
```
brickpop: But now, this means that a passed proposal that was not executed will be considered as DEFEATED, which is not correct. Why would an executable proposal be displayed as defeated?

If the plugin won't allow it to be executed after expired, then we would need another state

modules/client/src/client-common/utils.ts
```diff
@@ -77,7 +80,6 @@ export function computeProposalStatusFilter(
       break;
     case ProposalStatus.DEFEATED:
       where = {
-        potentiallyExecutable: false,
```
brickpop, line 80: Why isn't this line required anymore?

josemarinas: Basically on the multisig subgraph if a proposal has minApprovals of 1, potentiallyExecutable is set to true by default. This causes that some issues while filtering

modules/client/src/client-common/utils.ts
josemarinas: I dont think so, if is potentially executable and the date expired it will appear as succeeded. Maybe I should give it a go with the tests ant test more edge cases but i think tthe change is ok

modules/client/src/client-common/utils.ts
brickpop: But how about the cases where minApprovals is not 1?

modules/client/src/client-common/utils.ts
brickpop: Well, the execution of an executable proposal that already ended would not enter any of the "if" clauses. It would keep trying until the end of the function, when it would simply return DEFEATED

modules/client/src/client-common/utils.ts
```diff
@@ -64,22 +67,22 @@ export function computeProposalStatusFilter(
...
     case ProposalStatus.ACTIVE:
-      where = { startDate_lt: now, endDate_gte: now, executed: false };
+      where = { startDate_lt: now, endDate_gte: now, executed: false, potentiallyExecutable: false };
       break;
     case ProposalStatus.EXECUTED:
       where = { executed: true };
       break;
     case ProposalStatus.SUCCEEDED:
-      where = { potentiallyExecutable: true, endDate_lt: now };
+      where = { potentiallyExecutable: true };
```
brickpop: What if the proposal is potentially executable but the date expired? Isn't this condition tied to any other variable? `executed`, `executable`, `endDate` ?

modules/client/src/client-common/utils.ts
```diff
@@ -48,7 +48,10 @@ export function computeProposalStatus(
   if (startDate >= now) {
     return ProposalStatus.PENDING;
   }
```
brickpop, line 50: Should there be any check for the special case of minApprovals = 1 ?

modules/client/src/client-common/utils.ts
josemarinas: I mean the end date only shows you when you can and cant vote, there is not such thing as expired. If now is greater then endDate and potentiallyExecutable is true, congrats, your proposal passed :)

modules/client/src/client-common/utils.ts
josemarinas: Nope, i don't think so, the thing was that minApprovals = 1 triggered the potentiallyExecutable = true

Fabricevladimir: But without a check doesn't this mean that we will get s`ucceeded` by default for those proposals where `minApprovals` = 1 without anyone actually having voted? Perhaps we could add a check to see if there exists at least 1 vote when `potentiallyExecutable` = true?

modules/client/src/client-common/utils.ts
```diff
@@ -77,7 +80,6 @@ export function computeProposalStatusFilter(
       break;
     case ProposalStatus.DEFEATED:
```
Fabricevladimir, line 81: This doesn't quite work for the TokenVoting proposals. While the Multisig proposal is expired regardless of whether the approval threshold is met when the `endDate` has passed, this is not the case with the TokenVoting Proposal; we need to know whether it is `potentiallyExecutable`.

modules/client/src/client-common/utils.ts
```diff
@@ -64,7 +67,7 @@ export function computeProposalStatusFilter(
   const now = Math.round(new Date().getTime() / 1000).toString();
   switch (status) {
     case ProposalStatus.PENDING:
-      where = { startDate_gte: now };
+      where = { startDate_gte: now, potentiallyExecutable: false, executed: false };
```
Fabricevladimir, line 70: Hmm this means we won't get any pending Multisig proposals with `minApproval` = 1, since by default `potentiallyExecutable` is set to true for those.

Fabricevladimir (commented):
Thanks, Jose. One thing I fear is that it might be quite difficult to set these statuses and generate the filters without knowing the type of proposal since the current two types are similar yet also behave slightly differently

modules/client/src/client-common/utils.ts
josemarinas: True, good catch

modules/client/src/client-common/utils.ts
josemarinas: yep then we are not going to be able to fix it.

Either we wait for the split where each plugin has its own compute status function or we create a compute status filter function for the multisig

josemarinas:
Closing this one and starting a new one to split the computeProposalStatus and computeProposalStatusFilter for each plugin

PR #258, opened 2023-07-20
Fix: Split status functions

josemarinas:
## Description
The statuses shoud be computed differently accreoss plugins this removes the function from the common utils and splits  it in 3 different functions one for each plugin

Task ID: [OS-561](https://aragonassociation.atlassian.net/browse/OS-561)

## Type of change

## Checklist:

[OS-561]: https://aragonassociation.atlassian.net/browse/OS-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/src/multisig/internal/utils.ts
+++ b/modules/client/src/multisig/internal/utils.ts
@@ -1,11 +1,15 @@
-import { getCompactProposalId, hexToBytes } from "@aragon/sdk-common";
-import { computeProposalStatus, SubgraphAction } from "../../client-common";
+import { InvalidProposalStatusError, getCompactProposalId, hexToBytes } from "@aragon/sdk-common";
+import { SubgraphAction } from "../../client-common";
 import { MultisigProposal, MultisigProposalListItem } from "../types";
 import {
   SubgraphMultisigProposal,
   SubgraphMultisigProposalListItem,
 } from "./types";
-import { DaoAction, ProposalMetadata } from "@aragon/sdk-client-common";
+import {
+  DaoAction,
+  ProposalMetadata,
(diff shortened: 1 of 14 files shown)


modules/client/src/multisig/internal/utils.ts
```diff
@@ -98,3 +102,54 @@ export function toMultisigProposalListItem(
...
+    parseInt(proposal.startDate) * 1000,
+  );
+  const endDate = new Date(parseInt(proposal.endDate) * 1000);
+  if (proposal.executed) {
+    return ProposalStatus.EXECUTED;
+  }
+  if (now < startDate) {
+    return ProposalStatus.PENDING;
+  }
+  if (proposal.potentiallyExecutable && now < endDate) {
```
Fabricevladimir, line 120: Are we expressly not taking into account the `minApproval` = 1 case? All non Pending proposals will be successful by default for a plugin with `minApproval` = 1

josemarinas: Its not the intended behaivour? I mean if you create a proposal witn minApprovals = 1, any membber of the multisig can just approve an execute, we can add the check to see if the proposal minApproval is 1 and the length of approvals is 0 to return it as Active

Fabricevladimir: I will leave this to your discretion

Fabricevladimir (dismissed):

brickpop (dismissed):

brickpop (approved):

PR #259, opened 2023-07-24
Feature: Add support nft deposits and withdrawals

josemarinas:
## Description
Add support for ERC721 deposits and withdraws

Task ID: [OS-184](https://aragonassociation.atlassian.net/browse/OS-184)

## Type of change

## Checklist:

[OS-184]: https://aragonassociation.atlassian.net/browse/OS-184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/internal/abi/erc721.ts
```diff
@@ -0,0 +1,350 @@
+import { ContractInterface } from "@ethersproject/contracts";
+
+export const ERC721_ABI: ContractInterface = [
```
brickpop: Can't we source the ABI from an official place?

modules/client/src/internal/client/methods.ts
```diff
@@ -277,76 +279,116 @@ export class ClientMethods extends ClientCore implements IClientMethods {
   ): AsyncGenerator<DaoDepositStepValue> {
     const signer = this.web3.getConnectedSigner();
 
-    if (params.type !== TokenType.NATIVE && params.type !== TokenType.ERC20) {
+    if (
+      params.type !== TokenType.NATIVE && params.type !== TokenType.ERC20 &&
+      params.type !== TokenType.ERC721
+    ) {
       throw new UseTransferError();
```
brickpop: Is this error name meaningful?

modules/client/src/internal/client/methods.ts
```diff
@@ -277,76 +279,116 @@ export class ClientMethods extends ClientCore implements IClientMethods {
   ): AsyncGenerator<DaoDepositStepValue> {
```
brickpop, line 280: Isn't this function becoming way too large? I would recommend splitting it into smaller chunks

modules/client/src/internal/client/methods.ts
```diff
@@ -277,76 +279,116 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+        parsedLog.args["tokenId"].toString() !== params.tokenId.toString()
+      ) {
+        throw new FailedDepositError();
+      }
+      yield {
+        key: DaoDepositSteps.DONE,
+        tokenId: params.tokenId,
+      };
+    } else {
+      throw new NotImplementedError("Token type not implemented");
```
brickpop: I would rather mention "invalid"

But the "if" at the top of the function makes this code unreachable. I highly recommend simplifying things

modules/client/src/internal/utils.ts
```diff
@@ -380,14 +380,24 @@ export function withdrawParamsFromContract(
...
+      return {
+        type: TokenType.ERC721,
+        tokenAddress: to,
+        recipientAddressOrEns: result[1],
+        tokenId: BigInt(result[2]),
+        daoAddressOrEns: result[0],
+      }
   }
+
   // TODO Add ERC721 and ERC1155
```
brickpop:
```suggestion
  // TODO Add ERC1155
```

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -85,6 +85,29 @@ describe("Client", () => {
...
+        tokenId: BigInt(10),
+      };
+
+      const withdrawAction = await client.encoding.withdrawAction(
+        withdrawParams,
+      );
+
+      expect(typeof withdrawAction).toBe("object");
+      expect(withdrawAction.data).toBeInstanceOf(Uint8Array);
+      expect(bytesToHex(withdrawAction.data)).toBe(
```
brickpop, line 110: We should assert on `tokenAddress` as well (to)

modules/client/src/internal/abi/erc721.ts
josemarinas: I guess we could use OpenZeppelin js library and import them from there

modules/client/src/internal/client/methods.ts
josemarinas: Maybe not the best name, but the massage is 

Please, use the token's transfer function directly

modules/client/src/internal/client/methods.ts
josemarinas: yep maybe it is too big right now, because then we will add 1155 and it will be even bigger

modules/client/src/internal/client/decoding.ts
```diff
@@ -27,7 +27,8 @@ import {
   IpfsError,
   resolveIpfsCid,
 } from "@aragon/sdk-common";
-import { erc20ContractAbi } from "../abi/erc20";
+import { abi as ERC20_ABI } from "@openzeppelin/contracts/build/contracts/ERC20Votes.json";
```
brickpop: Shouldn't it be just the standard ERC20 contract ABI? ERC20 votes is an extension of it

modules/client/src/internal/client/encoding.ts
```diff
@@ -19,7 +19,8 @@ import {
   permissionWithConditionParamsToContract,
 } from "../utils";
 import { Contract } from "@ethersproject/contracts";
-import { erc20ContractAbi } from "../abi/erc20";
+import { abi as ERC20_ABI } from "@openzeppelin/contracts/build/contracts/ERC20Votes.json";
```
brickpop: Same

modules/client/src/internal/client/estimation.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from "@aragon/sdk-common";
 import { AddressZero } from "@ethersproject/constants";
 import { Contract } from "@ethersproject/contracts";
-import { erc20ContractAbi } from "../abi/erc20";
+import { abi as ERC20_ABI } from "@openzeppelin/contracts/build/contracts/ERC20Votes.json";
```
brickpop: Same

modules/client/src/internal/client/methods.ts
```diff
@@ -7,28 +7,26 @@ import {
...
   UpdateAllowanceError,
-  UseTransferError,
 } from "@aragon/sdk-common";
 
 import { defaultAbiCoder } from "@ethersproject/abi";
 import { BigNumber } from "@ethersproject/bignumber";
 import { AddressZero } from "@ethersproject/constants";
 import { Contract, ContractTransaction } from "@ethersproject/contracts";
-import { erc20ContractAbi } from "../abi/erc20";
+import { abi as ERC20_ABI } from "@openzeppelin/contracts/build/contracts/ERC20Votes.json";
```
brickpop: +1

modules/client/src/internal/client/methods.ts
```diff
@@ -276,77 +275,49 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+            // Relay the yield steps to the caller as they are received
+            yield* this.setAllowance(
+              {
+                amount: params.amount,
+                spender: daoAddressOrEns,
+                tokenAddress,
+              },
+            );
+          }
+        }
```
brickpop: Wouldn't you also move this to an internal helper function?

modules/client/src/internal/constants.ts
```diff
@@ -6,10 +6,15 @@ import {
   PluginRepoBuildMetadata,
   PluginRepoReleaseMetadata,
 } from "../types";
-import { erc20ContractAbi } from "./abi/erc20";
+import { abi as ERC20_ABI } from "@openzeppelin/contracts/build/contracts/ERC20Votes.json";
```
brickpop: +1

modules/client/src/internal/client/decoding.ts
josemarinas: mmm, no because in some places we need to deploy the votes version, I can split the function in deploy-erc20 and deploy-erc20-votes

modules/client/src/internal/client/methods.ts
josemarinas: Yes the problem is that his calls `this.setAllowance` so i'd nedd to create an internal function inside the class insted o creating it in utils, sinde it was a small piece of code i left it there

brickpop: What is the issue with the call to setAllowance? You can link one iterator to the call of another iterator as we are doing already

josemarinas: I abstracted getting the allowance but I cannot abstract setting the allowance because it uses this, and if I pass it as a function it loses the context, the other option would be abstracting also the setAllowance

brickpop (dismissed):

brickpop (approved):

PR #279, opened 2023-09-21
Feature: Dao Upgrade proposal checks

josemarinas:
## Description

Task ID: [OS-696](https://aragonassociation.atlassian.net/browse/OS-696)

## Type of change

## Checklist:

- [x] I have updated the `CHANGELOG.md` file in the root folder of the package after the `[UPCOMING]` title and before the latest version.

[OS-696]: https://aragonassociation.atlassian.net/browse/OS-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client-common/src/types.ts
+++ b/modules/client-common/src/types.ts
@@ -40,6 +40,15 @@ export enum SupportedNetwork {
 
 export const SupportedNetworksArray = Object.values(SupportedNetwork);
 
+
+export enum SupportedVersion {
+  V1_0_0 = "v1.0.0",
+  V1_3_0 = "v1.3.0",
+  LATEST = "v1.3.0",
+}
+
+export const SupportedVersionsArray = Object.values(SupportedVersion);
+
 /**
  * Contains the payload passed to the global DAO factory so that
  * plugins can be initialized
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -22,7 +22,7 @@ TEMPLATE:
 ### Added
 
 - Add prepareUpdate function to all clients
-- Addd check for dao update proposal
+- Add check for dao update proposal
 
 ### Changed
 
--- a/modules/client/CHANGELOG.md
+++ b/modules/client/CHANGELOG.md
@@ -22,6 +22,7 @@ TEMPLATE:
 ### Added
 
 - Add prepareUpdate function to all clients
+- Addd check for dao update proposal
 
 ### Changed
 
--- a/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
+++ b/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
@@ -1,0 +1,37 @@
+/* MARKDOWN
+---
+title: Check if a proposal valid for updating a dao
+---
+
+### Check if a proposal valid for updating a dao
+
+It will check that the proposal has the correct actions to update a dao.
+It will check that the eencoded previous version matches the dao version
+it will check that the implementation address matches the latest dao factory address
+It will check that the init data is empty becuase the dao factory does not have new params so init data is not necessary
+
+*/
+
+import { Client } from "@aragon/sdk-client";
+import { context } from "../index";
+
+// Instantiate the general purpose client from the Aragon OSx SDK context.
+const client: Client = new Client(context);
+
+const proposalId =
+  "0x1234567890123456789012345678901234567890123456789012345678901234";
+
+// Check proposal validity.
+const daoUpdateValidity = client.methods.isDaoUpdateProposalValid(proposalId);
+console.log(daoUpdateValidity);
+/* MARKDOWN
+    Returns:
+    ```tsx
+    {    
+        version: true,
+        implementation: true,
+        initData: true,
+        actions: true
+    }
+    ```
+    */
--- a/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
+++ b/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
@@ -8,7 +8,7 @@ title: Check if a proposal valid for updating a dao
 It will check that the proposal has the correct actions to update a dao.
 It will check that the encoded previous version matches the dao version
 It will check that the implementation address matches the latest dao factory address
-It will check that the init data is empty becuase the dao factory does not have new params so init data is not necessary
+It will check that the init data is empty because the dao factory does not have new params so init data is not necessary
 
 */
 
--- a/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
+++ b/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
@@ -6,7 +6,7 @@ title: Check if a proposal valid for updating a dao
 ### Check if a proposal valid for updating a dao
 
 It will check that the proposal has the correct actions to update a dao.
-It will check that the eencoded previous version matches the dao version
+It will check that the encoded previous version matches the dao version
 it will check that the implementation address matches the latest dao factory address
 It will check that the init data is empty becuase the dao factory does not have new params so init data is not necessary
 
--- a/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
+++ b/modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
@@ -7,7 +7,7 @@ title: Check if a proposal valid for updating a dao
 
 It will check that the proposal has the correct actions to update a dao.
 It will check that the encoded previous version matches the dao version
-it will check that the implementation address matches the latest dao factory address
+It will check that the implementation address matches the latest dao factory address
 It will check that the init data is empty becuase the dao factory does not have new params so init data is not necessary
 
 */
--- a/modules/client/src/internal/client/methods.ts
+++ b/modules/client/src/internal/client/methods.ts
@@ -64,12 +64,14 @@ import {
   DaoQueryParams,
   DaoSortBy,
   DaoUpdateProposalValidity,
+  DaoUpdateProposalValidityCause,
   DepositErc1155Params,
   DepositErc20Params,
   DepositErc721Params,
   DepositEthParams,
   DepositParams,
   HasPermissionParams,
+  InitializeFromParams,
   PluginQueryParams,
   PluginRepo,
   PluginRepoBuildMetadata,
@@ -86,8 +88,10 @@ import {
   Transfer,
   TransferQueryParams,
   TransferSortBy,
+  UpgradeToAndCallParams,
 } from "../../types";
 import {
+  IsDaoUpdateProposalValidParams,
   SubgraphBalance,
   SubgraphDao,
   SubgraphDaoListItem,
@@ -128,6 +132,7 @@ import { PermissionIds } from "../../constants";
 import {
   ClientCore,
   findLog,
+  LIVE_CONTRACTS,
   MULTI_FETCH_TIMEOUT,
   MultiTargetPermission,
   prepareGenericInstallation,
@@ -137,6 +142,7 @@ import {
   PrepareUpdateParams,
   PrepareUpdateStepValue,
   SortDirection,
+  SupportedVersion,
   TokenType,
 } from "@aragon/sdk-client-common";
 
@@ -1086,11 +1092,21 @@ export class ClientMethods extends ClientCore implements IClientMethods {
   }
 
   public async isDaoUpdateProposalValid(
-    proposalId: string,
+    params: IsDaoUpdateProposalValidParams,
   ): Promise<DaoUpdateProposalValidity> {
+    const { proposalId } = params;
     if (!isProposalId(proposalId)) {
       throw new InvalidProposalIdError();
     }
+    let { version } = params;
+    if (!version) {
+      // if no version get the latest from the dao factory
+      version = await this.getProtocolVersion(
+        this.web3.getAddress("daoFactoryAddress"),
+      );
+    }
+    // convert version to string using the fromat from SupportedVersion (vX.X.X)
+    const versionString = "v" + version.join(".") as SupportedVersion;
     // search for the proposal
     const name = "IProposal";
     const query = QueryIProposal;
@@ -1103,50 +1119,96 @@ export class ClientMethods extends ClientCore implements IClientMethods {
     if (!iproposal) {
       throw new ProposalNotFoundError();
     }
-    const res: DaoUpdateProposalValidity = {
-      version: false,
-      implementation: false,
-      initData: false,
-      actions: false,
-    };
-    // check if the proposal has only one action for upgrading the dao
-    if (iproposal.actions.length === 1) {
-      const action = iproposal.actions[0];
+    const causes: DaoUpdateProposalValidityCause[] = [];
+    // Check if any action try to upgrade the dao
+    for (const [index, action] of iproposal.actions.entries()) {
+      let decodedUpgradeToAndCallParams: UpgradeToAndCallParams;
+      let decodedInitializeFromParams: InitializeFromParams;
       try {
-        const upgradeToAndCallParams = decodeUpgradeToAndCallAction(
+        decodedUpgradeToAndCallParams = decodeUpgradeToAndCallAction(
           hexToBytes(action.data),
         );
-        const decodedInitializeFromParams = decodeInitializeFromAction(
-          upgradeToAndCallParams.data,
+        decodedInitializeFromParams = decodeInitializeFromAction(
+          decodedUpgradeToAndCallParams.data,
         );
-        // action is decoded that mean the actions are valid
-        res.actions = true
-        const currentDaoVersion = await this.getProtocolVersion(
-          iproposal.dao.id,
-        );
-        // check if the version of the specified dao matches the version in the encoded data
-        if (
-          JSON.stringify(currentDaoVersion) ===
-            JSON.stringify(decodedInitializeFromParams.previousVersion)
-        ) {
-          res.version = true;
-        }
-        // check if the implementation matches the implementation in the latest daoFactoryAddress
-        if (
-          this.web3.getAddress("daoFactoryAddress") ===
-            upgradeToAndCallParams.implementationAddress
-        ) {
-          res.implementation = true;
-        }
-        // initData is valid if it is empty becuse the new implementation of protocol version 1.3.0 does not need more parameters
-        if (decodedInitializeFromParams.initData?.length === 0) {
-          res.initData = true;
-        }
       } catch {
-        // if any of the above fails, the proposal is invalid
-        return res;
+        if (index === iproposal.actions.length - 1) {
+          causes.push(DaoUpdateProposalValidityCause.INVALID_ACTION);
+          return { isValid: false, causes };
+        }
+        continue;
       }
+      // check version
+      if (
+        !this.isDaoUpdateVersionValid({
+          previousVersion: decodedInitializeFromParams.previousVersion,
+          daoAddress: iproposal.dao.id,
+        })
+      ) {
+        causes.push(DaoUpdateProposalValidityCause.INVALID_VERSION);
+      }
+      // check implementation
+      if (
+        !this.isDaoUpdateImplementationValid({
+          version: versionString,
+          implementationAddress:
+            decodedUpgradeToAndCallParams.implementationAddress,
+        })
+      ) {
+        causes.push(DaoUpdateProposalValidityCause.INVALID_VERSION);
+      }
+      // check data
+      if (
+        !this.isDaoUpdateInitDataValid({
+          version: versionString,
+          data: decodedInitializeFromParams.initData || new Uint8Array(),
+        })
+      ) {
+        causes.push(DaoUpdateProposalValidityCause.INVALID_INIT_DATA);
+      }
+      // the action was valid so we stop looping
+      break;
     }
-    return res;
+    return { isValid: causes.length === 0, causes };
+  }
+
+  public async isDaoUpdateVersionValid(
+    { daoAddress, previousVersion }: {
+      previousVersion: [number, number, number];
+      daoAddress: string;
+    },
+  ): Promise<boolean> {
+    const currentDaoVersion = await this.getProtocolVersion(daoAddress);
+    return JSON.stringify(currentDaoVersion) ===
+      JSON.stringify(previousVersion);
+  }
+
+  public async isDaoUpdateImplementationValid(
+    { version, implementationAddress }: {
+      version: SupportedVersion;
+      implementationAddress: string;
+    },
+  ): Promise<boolean> {
+    const networkName = this.web3.getNetworkName();
+    const daoFactoryAddress =
+      LIVE_CONTRACTS[version][networkName].daoFactoryAddress;
+    const daoFactoryImplementation = DAOFactory__factory.connect(
+      daoFactoryAddress,
+      this.web3.getSigner(),
+    );
+    const daoBase = await daoFactoryImplementation.daoBase();
+    return daoBase === implementationAddress;
+  }
+
+  public async isDaoUpdateInitDataValid(
+    { version: _version, data }: {
+      version: SupportedVersion;
+      data: Uint8Array;
+    },
+  ): Promise<boolean> {
+    // TODO: decode the data using the abi from the the prepare update
+    // for now the init data must be empty but this can change in the future
+    // atm we cannot know the parameters for each version of the dao
+    return data.length === 0;
   }
 }
--- a/modules/client/src/internal/client/methods.ts
+++ b/modules/client/src/internal/client/methods.ts
@@ -1094,6 +1094,30 @@ export class ClientMethods extends ClientCore implements IClientMethods {
     return version;
   }
 
+  /**
+   * Check if a dao update proposal is valid
+   * It checks that the actions contain an
+   * upgrade to and call action that upgrades the dao
+   * with the correct data obtained from the 
+   * initialize from action
+   * 
+   * The values checked are:
+   * - Version: The version of the dao that created
+   * the proposal should be the same as the previous version
+   * encoded in the initialize from action
+   * 
+   * - Implementation: The implementation address should be the same
+   * as the one from the dao factory for the specified version. If
+   * the version is not specified it will get the latest version from
+   * the dao factory
+   * 
+   * - Init data: The init data should be empty for the current version.
+   * This can change in the future
+   *
+   * @param {IsDaoUpdateProposalValidParams} params
+   * @return {*}  {Promise<DaoUpdateProposalValidity>}
+   * @memberof ClientMethods
+   */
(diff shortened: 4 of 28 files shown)


modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
```diff
@@ -0,0 +1,37 @@
+/* MARKDOWN
+---
+title: Check if a proposal valid for updating a dao
+---
+
+### Check if a proposal valid for updating a dao
+
+It will check that the proposal has the correct actions to update a dao.
+It will check that the eencoded previous version matches the dao version
```
Rekard0:
```suggestion
It will check that the encoded previous version matches the dao version
```

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
```diff
@@ -0,0 +1,37 @@
...
+it will check that the implementation address matches the latest dao factory address
+It will check that the init data is empty becuase the dao factory does not have new params so init data is not necessary
```
Rekard0: I think we should mention for upgrading to protocol version 1.3.0 , right?

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
```diff
@@ -0,0 +1,37 @@
...
+    Returns:
+    ```tsx
+    {    
+        version: true,
+        implementation: true,
+        initData: true,
+        actions: true
+    }
+    ```
+    */
```
Rekard0, line 37: This imply that this method can only be used if a proposal is already created, 
And can not be used inside the proposal creation flow, is that ok @jamesej @Michael-A-Heuer ?

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      version: false,
+      implementation: false,
+      initData: false,
+      actions: false,
+    };
+    // check if the proposal has only one action for upgrading the dao
+    if (iproposal.actions.length === 1) {
+      const action = iproposal.actions[0];
+      try {
+        const upgradeToAdCallParams = decodeUpgradeToAndCallAction(
```
Rekard0:
```suggestion
        const upgradeToAndCallParams = decodeUpgradeToAndCallAction(
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+          hexToBytes(action.data),
+        );
+        const decodedInitializeFromParams = decodeInitializeFromAction(
+          upgradeToAdCallParams.data,
+        );
+        // action is decoded son the actions are valid
```
Rekard0:
```suggestion
        // action is decoded that mean the actions are valid
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+          iproposal.dao.id,
+        );
+        // check if the version of the specified dao matches the version in the encoded data
+        if (
+          JSON.stringify(currentDaoVersion) ===
+            JSON.stringify(decodedInitializeFromParams.previousVersion)
+        ) {
+          res.version = true;
+        }
+        // check if the implementation matches the latest daoFactoryAddress
```
Rekard0:
```suggestion
        // check if the implementation matches the implementation in the latest daoFactoryAddress
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+        if (
+          this.web3.getAddress("daoFactoryAddress") ===
+            upgradeToAdCallParams.implementationAddress
+        ) {
+          res.implementation = true;
+        }
```
Rekard0: Don't u have to the contract of `daoFactory` then get the implementation (`daoBase`) then compare it to `upgradeToAdCallParams.implementationAddress`?

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+        // initData is valid if it is empty becuse the new implementation doees not need more parameters
```
Rekard0:
```suggestion
        // initData is valid if it is empty becuse the new implementation of protocol version 1.3.0 does not need more parameters
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+    });
+    if (!iproposal) {
+      throw new ProposalNotFoundError();
+    }
+    const res: DaoUpdateProposalValidity = {
+      version: false,
+      implementation: false,
+      initData: false,
+      actions: false,
+    };
```
Rekard0: i would return {isValid: bool, resaons: {....}} but i will let u be the judge of that

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
     }
     return version;
   }
+
+  public async isDaoUpdateProposalValid(
+    proposalId: string,
+  ): Promise<DaoUpdateProposalValidity> {
```
Rekard0: we need to think about this function to be scaleable for future
like for example it recieve the version number and based on the version it decide if the proposal is valid or not
currectly i don't see a path for future, right? 
or u have some other idea?

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
josemarinas: Mmm, from what I saw on the figma this should be more or less generic, not specific for the version 1.3.0. Atm this would only work for the latest version of the contracts, this means if we are in version 1.0.0 and want to upgrade to 1.3.0 but 1.4.0 is alredy out this will return an error on the implementation address because it will compare it to the latest one, and we dont have a way of fixing this at the moment because the active contracts only hold the latest version

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
josemarinas: I saw the figma and I only saw this been used when the proposal is alredy done, I can implement it the other way around if you prefer it

modules/client/src/internal/client/methods.ts
josemarinas: Maybe I didnt fully understood the reuirement then, why does de daoBase os the original implmentation has to be equal to the new implementation?

modules/client/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ TEMPLATE:
 ### Added
 
 - Add prepareUpdate function to all clients
+- Addd check for dao update proposal
```
mathewmeconry:
```suggestion
- Add check for dao update proposal
```

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
```diff
@@ -0,0 +1,37 @@
+/* MARKDOWN
+---
+title: Check if a proposal valid for updating a dao
+---
+
+### Check if a proposal valid for updating a dao
+
+It will check that the proposal has the correct actions to update a dao.
+It will check that the encoded previous version matches the dao version
+it will check that the implementation address matches the latest dao factory address
```
mathewmeconry:
```suggestion
It will check that the implementation address matches the latest dao factory address
```

modules/client/src/internal/client/methods.ts
mathewmeconry: The DAOFactory is the contract that creates new DAOs and it holds a public constant called `daoBase` which it uses to create the new DAO. Thus an upgrade should point to the `daoBase` and not the DAOFactory

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1084,69 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+        }
+        // check if the implementation matches the implementation in the latest daoFactoryAddress
+        if (
+          this.web3.getAddress("daoFactoryAddress") ===
+            upgradeToAdCallParams.implementationAddress
+        ) {
+          res.implementation = true;
+        }
+        // initData is valid if it is empty becuse the new implementation of protocol version 1.3.0 does not need more parameters
+        if (decodedInitializeFromParams.initData?.length === 0) {
```
mathewmeconry: this doesn't scale well, when this should be a generic implementation. For example 1.4.0 may need some initialization data

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
Rekard0: it is not about my preference, when i looked at the code i thought of this, so i wanted to raise the point

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+          client.graphql.getClient(),
+        );
+        mockedClient.request.mockResolvedValueOnce({
+          iplugin: null,
+        });
+        expect(
+          () => client.methods.isDaoUpdateProposalValid(proposalId),
+        ).rejects.toThrow(new ProposalNotFoundError());
+      });
+      it("should check if a proposal is valid for updating the dao and fail with a proposal that has more than one action", async () => {
```
mathewmeconry: are these restrictions also in place in the frontend? because a proposal can not only upgrade DAO but also install a plugin at the same time. would that kind of proposal return a false in the check if the proposal is valid?

modules/common/CHANGELOG.md
```diff
@@ -21,8 +21,7 @@ TEMPLATE:
 
 ### Added
 
-- New error classes `InvalidVersionError` and `PluginUpdatePreparationError`
-
+- New error classes `InvalidVersionError`. `PluginUpdatePreparationError` and `ProposalNotFoundError`
```
mathewmeconry:
```suggestion
- New error classes `InvalidVersionError`, `PluginUpdatePreparationError`, and `ProposalNotFoundError`
```

modules/client/src/internal/client/methods.ts
Rekard0: if u look at the daoFactory's code u see it deploy a proxy and set daoBase as implementation
so we needed to compare update's new implem. to the one on the current daoFactory?

mathewmeconry (changes requested):
I have to second @Rekard0 comments about thinking again about how we do the validation. From my point of view they don't scale well and the SDK should support checking a proposal before it was created and not afterwards

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
jamesej: Yeah it's fine for this to be only for existing proposals. When we create a proposal for updates, all the calldata is automatically generated so we can be sure it's valid. This check is to warn the user if an update proposal might have been created outside the app and be malicious or incorrect.

modules/client-common/src/types.ts
```diff
@@ -40,6 +40,15 @@ export enum SupportedNetwork {
 
 export const SupportedNetworksArray = Object.values(SupportedNetwork);
 
+
+export enum SupportedVersion {
+  V1_0_0 = "v1.0.0",
+  V1_3_0 = "v1.3.0",
+  LATEST = "v1.3.0",
```
heueristik, line 47:
```suggestion
  LATEST = SupportedVersion.V1_3_0,
```
Suprisingly, this works, as you can [check on the TS playground](
https://www.typescriptlang.org/play?#code/KYOwrgtgBAymAO8D2AnALsAJgNWCgzgJZIhQDeAUFFNgIwD6ADE1ALxQBEAbrQHSP8OAGio0GAZhbtufcYJHUAMgEEAKgFEYqtrATJ0WXAWIhedepMYiAvhQoBjEviQAbYLxdIA5gAoO55kYALmE4RFQMHDwiEjMGQIBKBydXd09ffwkmEKEw-UijGNNzSyTHEGc3D28-FQ0tHN1wgyjjWLrNVQSgA)

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
heueristik: > the dao factory does not have new params

The `DAO` implementation receives the `initData` in `initializeFrom`, not the `DAOFactory`

In general, please use markdowns "\`" signs to discriminate contract/function/variable names from written language.

heueristik: > we dont have a way of fixing this at the moment because the active contracts only hold the latest version

Please create a task for this if not already present and explain the problem in detail @josemarinas @Rekard0

modules/client-common/src/types.ts
```diff
@@ -40,6 +40,15 @@ export enum SupportedNetwork {
 
 export const SupportedNetworksArray = Object.values(SupportedNetwork);
 
+
+export enum SupportedVersion {
+  V1_0_0 = "v1.0.0",
```
heueristik, line 45: Is it a good idea to use a string for the semantic version?
Why not use an object like
```ts
type SemVer = {
  MAJOR: number;
  MINOR: number;
  PATCH: number;
  RELEASE_CANDIDATE: string;
};
```
or similar? I know that this would enforce a different structure on LIVE_CONTRACTS, but this could be good. If others think so too, we should consider this.

modules/client/src/types.ts
```diff
@@ -422,3 +422,15 @@ export type InitializeFromParams = {
   previousVersion: [number, number, number];
   initData?: Uint8Array;
 };
+
+export type DaoUpdateProposalValidity = {
+  isValid: boolean;
+  causes: DaoUpdateProposalValidityCause[];
+};
+
+export enum DaoUpdateProposalValidityCause {
```
heueristik:
```suggestion
export enum DaoUpdateProposalInvalidityCause {
```

modules/client/src/types.ts
```diff
@@ -422,3 +422,15 @@ export type InitializeFromParams = {
   previousVersion: [number, number, number];
   initData?: Uint8Array;
 };
+
+export type DaoUpdateProposalValidity = {
+  isValid: boolean;
+  causes: DaoUpdateProposalValidityCause[];
```
heueristik:
```suggestion
  invalidityCauses: DaoUpdateProposalInvalidityCause[];
```

heueristik: Maybe there is a redundancy here between the `causes` array and the `isValid` boolean. Isn't an empty `causes` array already signaling validity?

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      throw new InvalidProposalIdError();
+    }
+    let { version } = params;
+    if (!version) {
+      // if no version get the latest from the dao factory
+      version = await this.getProtocolVersion(
+        this.web3.getAddress("daoFactoryAddress"),
+      );
+    }
+    // convert version to string using the fromat from SupportedVersion (vX.X.X)
```
heueristik:
```suggestion
// convert version to string using the fromat from SupportedVersion (vX.Y.Z)
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+    const versionString = "v" + version.join(".") as SupportedVersion;
```
heueristik, line 1136: Again, wouldn't it be better to use a proper struct for this instead of a string?

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      let decodedInitializeFromParams: InitializeFromParams;
+      try {
+        decodedUpgradeToAndCallParams = decodeUpgradeToAndCallAction(
+          hexToBytes(action.data),
+        );
+        decodedInitializeFromParams = decodeInitializeFromAction(
+          decodedUpgradeToAndCallParams.data,
+        );
+      } catch {
+        if (index === iproposal.actions.length - 1) {
```
heueristik, line 1164: What is this line doing? What assumptions are behind this resulting in the classification of as a `DaoUpdateProposal(In)ValidityCause.INVALID_ACTION`? Please write a comment above explaining it.

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+
+  public async isDaoUpdateVersionValid(
+    { daoAddress, previousVersion }: {
+      previousVersion: [number, number, number];
+      daoAddress: string;
+    },
+  ): Promise<boolean> {
+    const currentDaoVersion = await this.getProtocolVersion(daoAddress);
+    return JSON.stringify(currentDaoVersion) ===
+      JSON.stringify(previousVersion);
```
heueristik: Reading it like that, it does not make sense.

Maybe let's rename `currentDaoVersion` to `actualVersion` and `previousVersion` to `specifiedVersion`?
Still not clear enough in my opinion. We need some comments explaininig exactly what this function does. Otherwise, we will forget ourselves in 2 months from now

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+  }
+
+  public async isDaoUpdateImplementationValid(
```
heueristik, line 1230: Again, more comments about the assumptions / context are needed, IMO. What is checked exactly and why

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+
+  public async isDaoUpdateInitDataValid(
+    { version: _version, data }: {
+      version: SupportedVersion;
+      data: Uint8Array;
+    },
+  ): Promise<boolean> {
+    // TODO: decode the data using the abi from the the prepare update
+    // for now the init data must be empty but this can change in the future
+    // atm we cannot know the parameters for each version of the dao
```
heueristik, line 1255: Good, please double check that there is a task for this.

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
       //   "Invalid ENS name"
       // );
     });
+
+    describe("isDaoUpdateProposalValid", () => {
+      it("should check if a proposal is valid for updating the dao and fail with an invalid proposal id", async () => {
```
heueristik:
```suggestion
      it("throws an `InvalidProposalIdError` if the proposal ID is ???", async () => {
```
Please improve the test description.

My recommendation: Do not repeat the part of the description that is already part of the `describe` statement for every test. "isDaoUpdateProposalValid" already conveys thaat this is about checking if a proposal is valid.

modules/client/src/internal/client/methods.ts
```diff
@@ -1075,4 +1090,125 @@ export class ClientMethods extends ClientCore implements IClientMethods {
     }
     return version;
   }
+
+  public async isDaoUpdateProposalValid(
+    params: IsDaoUpdateProposalValidParams,
+  ): Promise<DaoUpdateProposalValidity> {
+    const { proposalId } = params;
+    if (!isProposalId(proposalId)) {
```
heueristik, line 1125: What are the assumptions behind this resulting in the classification as invalid? Please write a comment.

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const ctx = new Context(contextParamsLocalChain);
+        const client = new Client(ctx);
+        const proposalId = TEST_DAO_ADDRESS;
+        expect(
+          () => client.methods.isDaoUpdateProposalValid(proposalId),
+        ).rejects.toThrow(new InvalidProposalIdError());
+      });
+
+      it("should check if a proposal is valid for updating the dao and fail with a proposal that does not exist", async () => {
```
heueristik:
```suggestion
      it("throws an `ProposalNotFoundError`` for a proposal that does not exist", async () => {
```
see above

modules/client/test/integration/client/methods.test.ts
heueristik:
```suggestion
      it("invalidates a proposal that has more than one action and states the invalidity cause", async () => {
```
Let's be more precise in writing the tested behavior in the test description.

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(false);
+        expect(validationResult.implementation).toBe(false);
+        expect(validationResult.initData).toBe(false);
+        expect(validationResult.version).toBe(false);
+      });
+      it("should check if a proposal is valid for updating the dao and fail with a proposal has the wrong action", async () => {
```
heueristik: see above

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(false);
+        expect(validationResult.implementation).toBe(false);
+        expect(validationResult.initData).toBe(false);
+        expect(validationResult.version).toBe(false);
+      });
+      it("should check if a proposal is valid for updating the dao and fail with a proposal that has a wrong implementation address", async () => {
```
heueristik: see above

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(true);
+        expect(validationResult.implementation).toBe(false);
+        expect(validationResult.initData).toBe(true);
+        expect(validationResult.version).toBe(true);
+      });
+      it("should check if a proposal is valid for updating the dao and fail with a proposal that has a wrong version", async () => {
```
heueristik: see above

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(true);
+        expect(validationResult.implementation).toBe(true);
+        expect(validationResult.initData).toBe(true);
+        expect(validationResult.version).toBe(false);
+      });
+      it("should check if a proposal is valid for updating the dao and fail with a proposal that has a wrong init data", async () => {
```
heueristik: see above

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(true);
+        expect(validationResult.implementation).toBe(true);
+        expect(validationResult.initData).toBe(false);
+        expect(validationResult.version).toBe(true);
+      });
+      it("should check if a proposal is valid for updating the dao and succeed", async () => {
```
heueristik:
```suggestion
      it("validates a valid update proposal", async () => {
```

heueristik (changes requested):

heueristik:
General question/remark: 
Assuming we release `v1.4.0`, will the code be able to handle both updates paths,
 - `v1.0.0` -> `v1.4.0`
 - `v1.3.0` -> `v1.4.0`
 correctly (1.4.0, we most likely will have empty `initData` as well)?

I hope we will be able to deal with generic `initData` ASAP in case we urgently need to roll out an update that requires input data.

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
+        });
+
+        const validationResult = await client.methods
+          .isDaoUpdateProposalValid(
+            proposalId,
+          );
+        expect(validationResult.actions).toBe(false);
+        expect(validationResult.implementation).toBe(false);
+        expect(validationResult.initData).toBe(false);
+        expect(validationResult.version).toBe(false);
```
Michael-A-Heuer, line 1759: Would it be possible to write the test so that only `expect(validationResult.actions).toBe(false);` and the rest is `true` as you do below?

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1691,5 +1695,284 @@ describe("Client", () => {
...
```
Michael-A-Heuer, line 1797: Would it be possible to write the test so that only `expect(validationResult.actions).toBe(false);` and the rest is true as you do below?

modules/client/examples/01-client/16-is-dao-update-proposal-valid.ts
josemarinas: We solved this in the version you reviewed @Michael-A-Heuer  we are still missing a way of knowing if the update will need init data

modules/client/src/internal/client/methods.ts
josemarinas: I made a generic implementation for the active contracts, however the init data is not viable atm

modules/client/test/integration/client/methods.test.ts
josemarinas: After talking with james today this tests dont make sene anymore, so i'll need to mupdate them

modules/client-common/src/types.ts
josemarinas: Yep, TS does not complain but it crashes when I try to build it

modules/client-common/src/types.ts
josemarinas: I dont whink I can do that with TS and enums

modules/client/src/types.ts
heueristik: Those are invalidity cause, not validity causes. (Every item below starts with "INVALID_")

PR #285, opened 2023-10-06
F/os 738 move commons to a single package

josemarinas:
## Description

Task ID: [OS-738]()

## Type of change

## Checklist:

[OS-738]: https://aragonassociation.atlassian.net/browse/OS-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Rekard0 (dismissed):

modules/client-common/src/constants.ts
```diff
@@ -407,3 +408,13 @@ Object.freeze(Permissions);
 export { Permissions };
 Object.freeze(PermissionIds);
 export { PermissionIds };
+export const IPFS_CID_REGEX =
```
heueristik, line 411: The regex only applies to CIDv0. Note, that there is also CIDv1 (https://ipfs-search.readthedocs.io/en/latest/ipfs_datatypes.html#cid)

modules/client-common/src/constants.ts
```diff
@@ -407,3 +408,13 @@ Object.freeze(Permissions);
...
+  /^Qm([1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})$/;
+
+export const IPFS_URI_REGEX =
```
heueristik, line 414: See above

modules/client-common/src/encoding.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @return {*}  {{ pluginAddress: string; id: number }}
+ */
+export function decodeProposalId(
+  proposalId: string,
+): { pluginAddress: string; id: number } {
+  if (!isProposalId(proposalId)) {
+    throw new InvalidProposalIdError();
+  }
+  // const matchedRegexResult =
+  //   proposalId.match(/^(0x[A-Fa-f0-9]{40})_(0x[A-Fa-f0-9]{1,64})$/) || [];
```
heueristik: ?

modules/client-common/test/unit/encoding.test.ts
```diff
@@ -0,0 +1,452 @@
...
+      expect(() => decodeRatio(10 ** 10, 9)).toThrow(
+        "The value is out of range",
+      );
+    });
+  });
+  describe("encodeProposalId", () => {
+    it("Should encode a nonce-based proposalId", () => {
+      const entries = [
+        {
+          addr: "0x0000000000000000000000000000000000000000",
```
heueristik, line 212: In OSx, we use format string for better readability. E.g.,
```ts
`0x${'0'.repeat(39)}1`, // address(1)
```
Maybe you want to use this too

modules/client-common/src/constants.ts
josemarinas: I think is missing some parenthesis, but it should support both v1 and v0

heueristik: Ah, in this case, it would be good to write a test for it.

josemarinas: True, I thought i had tests for both cases, my bad

modules/client-common/test/unit/encoding.test.ts
josemarinas: Mmm, i thing is going to be a bit messier to change it just here and leave in the other places as a harcoded string an just here as a function with, repeat, I can considerchanging it But I'd do it in all places at the same time

heueristik (approved):

PR #290, opened 2023-10-17
Feature: Update validation functions so they receive actions instead of proposalId

josemarinas:
## Description

Task ID: [OS-774]()

## Type of change

## Checklist:

[OS-774]: https://aragonassociation.atlassian.net/browse/OS-774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/client/src/internal/client/methods.ts
+++ b/modules/client/src/internal/client/methods.ts
@@ -166,6 +166,7 @@ import {
   DepositErc721Schema,
   DepositEthSchema,
   HasPermissionSchema,
+  IsDaoUpdateValidSchema,
   IsPluginUpdateValidSchema,
   PluginQuerySchema,
 } from "../schemas";
@@ -1316,7 +1317,10 @@ export class ClientMethods extends ClientCore implements IClientMethods {
 
     // find signatures in the actions specified in the proposal
     const grantIndex = findActionIndex(params.actions, grantSignature);
-    const applyUpdateIndex = findActionIndex(params.actions, applyUpdateSignature);
+    const applyUpdateIndex = findActionIndex(
+      params.actions,
+      applyUpdateSignature,
+    );
     const revokeIndex = findActionIndex(params.actions, revokeSignature);
 
     // check that all actions are present and in the correct order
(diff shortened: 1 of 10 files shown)


mathewmeconry (dismissed):
I would rename `findAction()` to `findActionIndex()` because it doesn't return the action but only the index of the action within the actions array. 
otherwise LGTM

modules/client/test/integration/client/methods.test.ts
```diff
@@ -1628,7 +1624,7 @@ describe("Client", () => {
         expect(plugins[0].current.release.metadata).toMatchObject(
           releaseMetadata,
         );
-        expect(plugins[0].releases.length).toBe(2)
+        expect(plugins[0].releases.length).toBe(2);
```
heueristik, line 1627: Is the SDK repo not linting / format checking itself? This should be a task for the future - not for now.

mathewmeconry (approved):

modules/client/src/internal/client/methods.ts
```diff
@@ -1110,12 +1109,12 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
   }
 
-  public isDaoUpdateProposal(
+  public isDaoUpdate(
     actions: DaoAction[],
   ): boolean {
     const initializeFromInterface = DAO__factory.createInterface()
       .getFunction("initializeFrom").format("minimal");
-    return findAction(actions, initializeFromInterface) !== -1;
+    return findActionIndex(actions, initializeFromInterface) !== -1;
```
heueristik, line 1117: I think I've found something here: There will be cases in which we update the DAO **without reinitializing** and thus without calling `initializeFrom`. This method of checking for DAO updates is therefore not sufficient. Shouldn't we rather look for `upgradeTo` or `upgradeToAndCall` in the actions?
@mathewmeconry @Rekard0

heueristik (changes requested):
I have a concern being unrelated to the changes of this PR. Let's take time to discuss it and maybe open a follow-up task.

modules/client/src/internal/client/methods.ts
Rekard0: we can address it here: https://aragonassociation.atlassian.net/browse/OS-777

Rekard0 (approved):

modules/client/test/integration/client/methods.test.ts
josemarinas: Yep, createed a task for that https://aragonassociation.atlassian.net/browse/OS-778

heueristik (approved):

PR #301, opened 2023-11-07
refactor: Plugin Update Security Check

josemarinas:
## Description

Tasks: [OS-813](https://aragonassociation.atlassian.net/browse/OS-813), [OS-835](https://aragonassociation.atlassian.net/browse/OS-835)
This PR also contains the documentation page to be linked in the frontend: [OS-834](https://aragonassociation.atlassian.net/browse/OS-834)

## Type of change

## Checklist:

- [x] I have updated the `CHANGELOG.md` file in the root folder of the package after the `[UPCOMING]` title and before the latest version.

[OS-813]: https://aragonassociation.atlassian.net/browse/OS-813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OS-835]: https://aragonassociation.atlassian.net/browse/OS-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OS-834]: https://aragonassociation.atlassian.net/browse/OS-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/client/src/internal/client/encoding.ts
```diff
@@ -164,26 +164,53 @@ export class ClientEncoding extends ClientCore implements IClientEncoding {
...
+      return [
+        grantUpgradeAction,
+        grantRootAction,
+        {
+          to: pspAddress,
+          value: BigInt(0),
+          data: hexToBytes(hexBytes),
+        },
+        revokeUpgradeAction,
+        revokeRootAction,
```
mathewmeconry:
```suggestion
        revokeRootAction,
        revokeUpgradeAction,
```

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -585,7 +585,73 @@ describe("Client", () => {
         );
       }
     });
-    it("Should encode an applyUpdate action", async () => {
+    it("Should encode an applyUpdate action without permissions", async () => {
```
mathewmeconry, line 588: please add here also checks that the upgrade and the root permission is properly granted and revoked

mathewmeconry (changes requested):

Rekard0 (commented):
apart from what Mathias pointed out, LGTM!

modules/client/src/internal/client/encoding.ts
```diff
@@ -164,26 +164,53 @@ export class ClientEncoding extends ClientCore implements IClientEncoding {
...
-    const revokeAction = this.revokeAction(daoAddress, {
+    // Revoke UPGRADE_PLUGIN_PERMISSION in the plugin to the PSP
+    const revokeUpgradeAction = this.revokeAction(daoAddress, {
       where: daoAddress,
       who: pspAddress,
-      permission: Permissions.ROOT_PERMISSION,
+      permission: Permissions.UPGRADE_PLUGIN_PERMISSION,
     });
+    // if there are permissions to be granted or revoked i need to
+    // grant ROOT_PERMISSION in the DAO to the PSP
```
heueristik:
```suggestion
    // If the update requests permissions to be granted or revoked, the PSP needs temporary `ROOT_PERMISSION_ID` permission
```

modules/client/src/internal/client/encoding.ts
heueristik: +1

modules/client/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ TEMPLATE:
 
 ## [UPCOMING]
 
+### Fixed
```
heueristik:
```suggestion
### Changed
```

modules/client/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ TEMPLATE:
...
+
+- Permission encoding in `applyUpdate`
```
heueristik: Please be more detailed here. What has changed in which context and why.

heueristik (changes requested):

modules/client-common/CHANGELOG.md
```diff
@@ -21,6 +21,12 @@ TEMPLATE:
 
 ### Added
 
+- Added `UPGRADE_PLUGIN_PERMISSION` to `PermissionType` enum
```
heueristik, line 24: More things happened in this PR @josemarinas

modules/client/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ TEMPLATE:
 
 ## [UPCOMING]
 
+### Changed
+
+- Add grant and revoke for permission `UPGRADE_PLUGIN_PERMISSION` in `applyUpdate` encoder
```
heueristik, line 24: More things happened in this PR @josemarinas

modules/client/src/internal/client/encoding.ts
```diff
@@ -164,26 +164,52 @@ export class ClientEncoding extends ClientCore implements IClientEncoding {
...
+        who: pspAddress,
+        permission: Permissions.ROOT_PERMISSION,
+      });
+      // Revoke ROOT_PERMISSION in the DAO to the PSP
+      const revokeRootAction = this.revokeAction(daoAddress, {
+        where: daoAddress,
+        who: pspAddress,
+        permission: Permissions.ROOT_PERMISSION,
+      });
+      return [
```
heueristik, line 193: We should rename the function, e.g., to `applyUpdateActionAndPermissionsBlock`. The function name suggests that you get a single action back, which is not the case. This change should also be reflected and explained in the jsdoc comment.
Descriptive/explicit naming is a must and we should never do shortcuts on this one.

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      params: { id: proposalId.toLowerCase() },
+      name,
+    });
+    if (!iproposal) {
+      return false;
+    }
+    const subgraphActions = iproposal.actions;
+    let actions = toDaoActions(subgraphActions);
+    const classifiedActions = classifyProposalActions(actions);
+    return isDaoUpdateAction(classifiedActions);
```
heueristik:
```suggestion
    return containsDaoUpdateAction(classifiedActions);
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-    daoAddress: string,
-  ): boolean {
-    const pspAddress = this.web3.getAddress("pluginSetupProcessorAddress");
-    return (
-      params.permission === Permissions.ROOT_PERMISSION &&
-      params.permissionId === PermissionIds.ROOT_PERMISSION_ID &&
-      params.where === daoAddress &&
-      params.who === pspAddress
-    );
+  public async isDaoUpdate(
```
heueristik:
```suggestion
  public async isDaoUpdateProposal(
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-      decodedApplyUpdateActionParams,
-      PreparationType.UPDATE,
-    );
-    // get plugin preparation
-    type W = { pluginPreparation: SubgraphPluginUpdatePreparation };
-    const { pluginPreparation } = await this.graphql.request<W>({
-      query: QueryPluginPreparations,
-      params: { where: { preparedSetupId } },
-      name: "pluginPreparation",
+  public async isPluginUpdate(
```
heueristik:
```suggestion
  public async isPluginUpdateProposal(
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-        PluginUpdateProposalInValidityCause.MISSING_PLUGIN_PREPARATION,
-      );
+    if (!iproposal) {
+      return false;
     }
-    return causes;
+    const subgraphActions = iproposal.actions;
+    let actions = toDaoActions(subgraphActions);
+    const classifiedActions = classifyProposalActions(actions);
+    return isPluginUpdateAction(classifiedActions) ||
```
heueristik:
```suggestion
    return containsPluginUpdateActionBlock(classifiedActions) ||
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
+      isPluginUpdateActionWithRootPermission(classifiedActions);
```
heueristik:
```suggestion
      isPluginUpdateActionBlockWithRootPermission(classifiedActions);
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
   /**
    * Check if the specified proposal id is valid for updating a plugin
-   * The failure map should be checked before calling this method
    *
-   * @param {DaoAction[]} actions
+   * @param {string} proposalId
    * @return {*}  {Promise<PluginUpdateProposalValidity>}
    * @memberof ClientMethods
    */
   public async isPluginUpdateValid(
```
mathewmeconry: refactor return type to have another field with proposal settings error cause.
This way the user cannot distinguish if it is a proposal error (invalid failure map, etc) or an error for the first action

modules/client/src/internal/constants.ts
```diff
@@ -121,3 +125,32 @@ export enum PreparationType {
...
+  ProposalActionTypes.APPLY_UPDATE,
+  ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
+];
+export const PLUGIN_UPDATE_WITH_ROOT_ACTION_PATTERN = [
+  ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
+  ProposalActionTypes.GRANT_ROOT_PERMISSION,
+  ProposalActionTypes.APPLY_UPDATE,
+  ProposalActionTypes.REVOKE_ROOT_PERMISSION,
+  ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
+];
```
mathewmeconry: ROOT and PLUGIN_UPDATE permissions don't have to be in this order to be valid. rewrite to have a more generic check

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+export function compareArrays<T>(array1: T[], array2: T[]): boolean {
+  if (array1.length !== array2.length) {
+    return false;
+  }
+  for (const [index, element] of array1.entries()) {
+    if (element !== array2[index]) {
+      return false;
+    }
+  }
+  return true;
```
mathewmeconry: `return JSON.stringify(array1) === JSON.stringify(array2)` would be the faster and easier option

modules/client/src/internal/client/methods.ts
heueristik: This name is getting longer and longer. However, this is ok as long as it is precise and describes exactly what the function is doing.
Generally, the need to give long names is a hint towards not having the right code structure. The length of names can often be reduced by organizing it differently.
E.g., splitting code into smaller files with meaningful file names / module scope (i.e., not having a single `methods.ts` containing everything, but categorizing file by their purpose (e.g., `proposal_analysis.ts`)) or putting them into classes/objects (e.g., a Proposal class that defines blocks/segments).

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+      !plugin.appliedVersion?.build ||
+      plugin.appliedVersion?.build >=
+        decodedParams.versionTag.build
+    ) {
+      causes.push(PluginUpdateProposalInValidityCause.INVALID_PLUGIN_BUILD);
+    }
+  } else {
+    causes.push(PluginUpdateProposalInValidityCause.PLUGIN_NOT_INSTALLED);
+    return causes;
+  }
```
mathewmeconry: is cleaner:
```suggestion
if(!plugin) {
    causes.push(PluginUpdateProposalInValidityCause.PLUGIN_NOT_INSTALLED);
    return causes;
}
    // check release is the same as the one installed
    if (
      plugin.appliedVersion?.release.release !==
        decodedParams.versionTag.release
    ) {
      causes.push(PluginUpdateProposalInValidityCause.INVALID_PLUGIN_RELEASE);
    }
    // check build is higher than the one installed
    if (
      !plugin.appliedVersion?.build ||
      plugin.appliedVersion?.build >=
        decodedParams.versionTag.build
    ) {
      causes.push(PluginUpdateProposalInValidityCause.INVALID_PLUGIN_BUILD);
    }
```

modules/client/src/internal/client/methods.ts
heueristik: Reading the name/jsdoc alone, I ask myself if it validates a single action, a block, or the entire proposal. I wouldn't think that it validates multiple plugin updates, which it does.

What is a good name here?
`isPluginUpdateProposalPartValid` ?
`areContainedPluginUpdatesValid` (-> there might be multiple, indpendent plugin update blocks inside)?

Not perfect yet

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+      !SupportedPluginRepoArray.includes(
+        pluginRepo.subdomain as SupportedPluginRepo,
+      )
+    ) {
+      causes.push(PluginUpdateProposalInValidityCause.NOT_ARAGON_PLUGIN_REPO);
+    }
+  } else {
+    causes.push(PluginUpdateProposalInValidityCause.MISSING_PLUGIN_REPO);
+    return causes;
+  }
```
mathewmeconry: is cleaner
```suggestion
  if (!pluginRepo) {
    causes.push(PluginUpdateProposalInValidityCause.MISSING_PLUGIN_REPO);
    return causes;
  }
    // check if is one of the aragon plugin repos
    if (
      !SupportedPluginRepoArray.includes(
        pluginRepo.subdomain as SupportedPluginRepo,
      )
    ) {
      causes.push(PluginUpdateProposalInValidityCause.NOT_ARAGON_PLUGIN_REPO);
    }
```

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-   * The failure map should be checked before calling this method
+   * Check if the specified proposalId actions are valid for updating a dao
    *
-   * @param {IsDaoUpdateValidParams} params
+   * @param {string} proposalId
+   * @param {SupportedVersion} [version]
    * @return {*}  {Promise<DaoUpdateProposalValidity>}
    * @memberof ClientMethods
    */
   public async isDaoUpdateValid(
```
heueristik:
```suggestion
  public async isDaoUpdateProposalValid(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+      causes.push(
+        PluginUpdateProposalInValidityCause.INVALID_PLUGIN_REPO_METADATA,
+      );
+    }
+  } else {
+    causes.push(
+      PluginUpdateProposalInValidityCause.MISSING_PLUGIN_PREPARATION,
+    );
+  }
+  return causes;
```
mathewmeconry:
```suggestion
  if (!pluginPreparation) {
     causes.push(
      PluginUpdateProposalInValidityCause.MISSING_PLUGIN_PREPARATION,
    );
    return causes;
  }
    // get the metadata of the plugin repo
    // for the release and build specified
    const release = pluginRepo.releases.find((
      release: SubgraphPluginRepoRelease,
    ) => release.release === decodedParams.versionTag.release);
    const build = release?.builds.find((
      build: { build: number; metadata: string },
    ) => build.build === decodedParams.versionTag.build);
    const metadataCid = build?.metadata;

    // fetch the metadata
    const metadata = await ipfs.fetchString(metadataCid!);
    const metadataJson = JSON.parse(metadata) as PluginRepoBuildMetadata;
    // get the update abi for the specified build
    if (
      metadataJson?.pluginSetup?.prepareUpdate[decodedParams.versionTag.build]
        ?.inputs
    ) {
      // if the abi exists try to decode the data
      const updateAbi = metadataJson.pluginSetup.prepareUpdate[
        decodedParams.versionTag.build
      ].inputs;
      try {
        if (
          decodedParams.initData.length > 0 &&
          updateAbi.length === 0
        ) {
          throw new Error();
        }
        // if the decode does not throw an error the data is valid
        defaultAbiCoder.decode(
          getNamedTypesFromMetadata(updateAbi),
          decodedParams.initData,
        );
      } catch {
        // if the decode throws an error the data is invalid
        causes.push(
          PluginUpdateProposalInValidityCause.INVALID_DATA,
        );
      }
    } else {
      causes.push(
        PluginUpdateProposalInValidityCause.INVALID_PLUGIN_REPO_METADATA,
      );
    }
  return causes;
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            classifiedActions.push(
+              ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
+            );
+          } else if (
+            decodedPermission.permission === Permissions.ROOT_PERMISSION
+          ) {
+            classifiedActions.push(
+              ProposalActionTypes.GRANT_ROOT_PERMISSION,
+            );
+          }
```
mathewmeconry: use a switch/case statement

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            classifiedActions.push(
+              ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
+            );
+          } else if (
+            decodedPermission.permission === Permissions.ROOT_PERMISSION
+          ) {
+            classifiedActions.push(
+              ProposalActionTypes.REVOKE_ROOT_PERMISSION,
+            );
+          }
```
mathewmeconry: use a switch/case statement

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+  }
+  // check if it matches the expected pattern
+  // length should be 5
+  if (
+    receivedPattern.length !== 5 ||
+    !compareArrays(receivedPattern, PLUGIN_UPDATE_WITH_ROOT_ACTION_PATTERN)
+  ) {
+    return false;
+  }
+  return true;
```
mathewmeconry: `return receivedPattern.length === 5 && compareArrays(receivedPattern, PLUGIN_UPDATE_WITH_ROOT_ACTION_PATTERN)`

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+  }
+  // check if it matches the expected pattern
+  // length should be 3
+  if (
+    receivedPattern.length !== 3 ||
+    !compareArrays(receivedPattern, PLUGIN_UPDATE_ACTION_PATTERN)
+  ) {
+    return false;
+  }
+  return true;
```
mathewmeconry: `return receivedPattern.length === 3 && compareArrays(receivedPattern, PLUGIN_UPDATE_ACTION_PATTERN)`

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+  daoAddress: string,
+  implementationAddress: string,
+  currentDaoVersion: [number, number, number],
+): DaoUpdateProposalValidity {
+  const classifiedActions = classifyProposalActions(actions);
+  const causes: DaoUpdateProposalInvalidityCause[] = [];
+  // check if the actions are valid
+  // if they are not valid return add
+  // invalid actions to the causes and return
+  if (isDaoUpdateAction(classifiedActions)) {
```
mathewmeconry:
```suggestion
  if (!isDaoUpdateAction(classifiedActions)) {
  causes.push(
      DaoUpdateProposalInvalidityCause.INVALID_ACTIONS,
    );
    return { isValid: false, causes };
  }
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+          DaoUpdateProposalInvalidityCause.INVALID_UPGRADE_TO_AND_CALL_DATA,
+        );
+      }
+    }
+  } else {
+    // add invalid actions to the causes array
+    causes.push(
+      DaoUpdateProposalInvalidityCause.INVALID_ACTIONS,
+    );
+  }
```
mathewmeconry:
```suggestion
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+    }
+    // if the value is different from 0
+    // add the cause to the causes array
+    if (upgradeAction.value.toString() !== "0") {
+      causes.push(
+        DaoUpdateProposalInvalidityCause.INVALID_VALUE,
+      );
+    }
+    // if the upgrade action is upgradeTo
+    if (upgradeActionType === ProposalActionTypes.UPGRADE_TO) {
```
mathewmeconry: use a switch/case statement

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+      const recCauses = await validateUpdatePluginProposalActions(
+        actions,
+        daoAddress,
+        pspAddress,
+        graphql,
+        ipfs,
+      );
+      causes = causes.concat(recCauses.causes);
+    }
+  } else if (isPluginUpdateActionWithRootPermission(classifiedActions)) {
```
mathewmeconry:
```suggestion
    return {
    // every item in the array should be empty
    isValid: causes.every((cause) => cause.length === 0),
    causes,
  };
  } 
  
  if (isPluginUpdateActionWithRootPermission(classifiedActions)) {
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+  ipfs: IClientIpfsCore,
+) {
+  let causes: PluginUpdateProposalInValidityCause[][] = [];
+  const classifiedActions = classifyProposalActions(actions);
+  if (isPluginUpdateAction(classifiedActions)) {
+    causes[0] = [];
+    for (const [index, action] of classifiedActions.entries()) {
+      switch (action) {
+        case ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            await validateGrantUpdatePluginPermissionAction(
+              actions[index],
+              pspAddress,
+              daoAddress,
+              graphql,
+            ),
+          );
+          break;
+        case ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+          );
+          break;
+        case ProposalActionTypes.APPLY_UPDATE:
+          const resCauses = await validateApplyUpdateFunction(
+            actions[index],
+            daoAddress,
+            graphql,
+            ipfs,
+          );
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+        ipfs,
+      );
+      causes = causes.concat(recCauses.causes);
+    }
+  } else if (isPluginUpdateActionWithRootPermission(classifiedActions)) {
+    causes[0] = [];
+    for (const [index, action] of classifiedActions.entries()) {
+      switch (action) {
+        case ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            await validateGrantUpdatePluginPermissionAction(
+              actions[index],
+              pspAddress,
+              daoAddress,
+              graphql,
+            ),
+          );
+          break;
+        case ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            await validateRevokeUpdatePluginPermissionAction(
+              actions[index],
+              pspAddress,
+              daoAddress,
+              graphql,
+            ),
+          );
+          break;
+        case ProposalActionTypes.GRANT_ROOT_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            validateGrantRootPermissionAction(
+              actions[index],
+              daoAddress,
+              pspAddress,
+            ),
+          );
+          break;
+        case ProposalActionTypes.REVOKE_ROOT_PERMISSION:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+            validateRevokeRootPermissionAction(
+              actions[index],
+              daoAddress,
+              pspAddress,
+            ),
+          );
+          break;
+        case ProposalActionTypes.APPLY_UPDATE:
+          causes[0].concat(
```
mathewmeconry: concat doesn't update the first array but only returns the merged array
```suggestion
          causes[0] = causes[0].concat(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,763 @@ export function toPluginPermissionOperationType(
...
+    }
+  } else {
+    causes[0] = [];
+    causes[0].push(
+      PluginUpdateProposalInValidityCause.INVALID_ACTIONS,
+    );
+    return {
+      isValid: false,
+      causes,
+    };
```
mathewmeconry: either rewrite that the `else` is not needed or don't return here. because on the next line you start returning and it will return `isValid = false` because you set her the cause

mathewmeconry: this check could also be moved up and check that the `actions` array is longer than 5 elements. because with the override on line 1603 the whole calculation process in the function is deleted and thus not necessary

mathewmeconry (changes requested):

modules/client/src/internal/client/methods.ts
```diff
@@ -1116,408 +1101,163 @@ export class ClientMethods extends ClientCore implements IClientMethods {
...
-      )
-    ) {
-      causes.push(
-        PluginUpdateProposalInValidityCause.INVALID_REVOKE_PERMISSION,
+    // check failure map
+    if (iproposal.allowFailureMap !== "0") {
+      causes[0] = [];
+      // if the failure map is not 0 return invalid failure map
+      causes[0].push(
+        PluginUpdateProposalInValidityCause.INVALID_ALLOW_FAILURE_MAP,
```
heueristik:
```suggestion
        PluginUpdateProposalInValidityCause.NON_ZERO_ALLOW_FAILURE_MAP_VALUE,
```

mathewmeconry:
@josemarinas the errors where the result of `concat` isn't stored should have been caught by the test. Please add test cases to check if these code snippets work as expected

modules/client/src/internal/types.ts
```diff
@@ -237,3 +237,19 @@ export enum SubgraphPluginPermissionOperation {
   REVOKE = "Revoke",
   GRANT_WITH_CONDITION = "GrantWithCondition",
 }
+
+export enum ProposalActionTypes {
+  UPGRADE_TO = "upgradeTo",
+  UPGRADE_TO_AND_CALL = "upgradeToAndCall",
+  APPLY_UPDATE = "applyUpdate",
+  GRANT_PLUGIN_UPDATE_PERMISSION = "grantUpdatePluginPermission",
```
heueristik:
```suggestion
  GRANT_PLUGIN_UPGRADE_PERMISSION = "grantUpgradePluginPermission",
```

modules/client/src/internal/constants.ts
heueristik: I think it is ok to do this in a second iteration when we implement the rule-based solution you proposed @mathewmeconry. Since the SDK will produce these actions, we can assume that they have this order for now. Deviating from them is not necessary. The goal for now is that we can exclude trickery/malicious stuff.

modules/client/src/internal/graphql-queries/dao.ts
```diff
@@ -47,44 +47,3 @@ export const QueryDaos = gql`
     }
   }
 `;
-
-export const QueryDaoTransfersByAddress = gql`
```
heueristik, line 51: Why is this deleted? Was it leftover code?
(Just wanting to make sure this removal is not accidental)

modules/client/src/types.ts
```diff
@@ -420,13 +420,56 @@ export type DecodedInitializeFromParams = {
...
+  INVALID_REVOKE_UPDATE_PERMISSION_VALUE =
+    "invalidRevokeUpdatePermissionCondition",
+  INVALID_REVOKE_UPDATE_PERMISSION_PERMISSION_ID =
+    "invalidRevokeUpdatePermissionPermissionId",
+  // Grant ROOT_PERMISSION action
+  INVALID_GRANT_ROOT_PERMISSION_WHO_ADDRESS =
+    "invalidGrantRootPermissionWhoAddress",
+  INVALID_GRANT_ROOT_PERMISSION_WHERE_ADDRESS =
+    "invalidGrantRootPermissionWhereAddress",
+  INVALID_GRANT_ROOT_PERMISSION_PERMISSION =
```
heueristik:
```suggestion
  INVALID_GRANT_ROOT_PERMISSION_PERMISSION_ID =
```

modules/client/src/types.ts
```diff
@@ -420,13 +420,56 @@ export type DecodedInitializeFromParams = {
...
+    "invalidGrantRootPermissionPermission",
+  INVALID_GRANT_ROOT_PERMISSION_VALUE = "invalidGrantRootPermissionCondition",
```
heueristik:
```suggestion
  NON_ZERO_GRANT_ROOT_PERMISSION_CALL_VALUE = "invalidGrantRootPermissionCondition",
```

modules/client/src/types.ts
```diff
@@ -420,13 +420,56 @@ export type DecodedInitializeFromParams = {
...
+  INVALID_GRANT_ROOT_PERMISSION_PERMISSION_ID =
+    "invalidGrantRootPermissionPermissionId",
+  // Revoke ROOT_PERMISSION action
+  INVALID_REVOKE_ROOT_PERMISSION_WHO_ADDRESS =
+    "invalidRevokeRootPermissionWhoAddress",
+  INVALID_REVOKE_ROOT_PERMISSION_WHERE_ADDRESS =
+    "invalidRevokeRootPermissionWhereAddress",
+  INVALID_REVOKE_ROOT_PERMISSION_PERMISSION =
+    "invalidRevokeRootPermissionPermission",
+  INVALID_REVOKE_ROOT_PERMISSION_VALUE = "invalidRevokeRootPermissionCondition",
```
heueristik:
```suggestion
  NON_ZERO_REVOKE_ROOT_PERMISSION_CALL_VALUE = "invalidRevokeRootPermissionCondition",
```

modules/client/src/types.ts
```diff
@@ -446,10 +489,16 @@ export type IsPluginUpdateProposalValidParams = {
 };
 
 export enum DaoUpdateProposalInvalidityCause {
+  PROPOSAL_NOT_FOUND = "proposalNotFound",
   INVALID_ACTIONS = "invalidActions",
-  INVALID_IMPLEMENTATION = "invalidImplementation",
-  INVALID_VERSION = "invalidVersion",
-  INVALID_INIT_DATA = "invalidInitData",
+  INVALID_TO_ADDRESS = "invalidToAddress",
+  INVALID_VALUE = "invalidValue",
```
heueristik:
```suggestion
  NON_ZERO_CALL_VALUE = "nonZeroCallValue",
```

modules/client/src/internal/graphql-queries/dao.ts
josemarinas: @Rekard0 asked me to remove it on the next PR is no longer used

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+    daoAddress,
+    decodedPermission.where,
+    graphql,
+  );
+  if (pluginInstallations.length === 0) {
+    causes.push(
+      PluginUpdateProposalInValidityCause
+        .INVALID_GRANT_UPDATE_PERMISSION_WHERE_ADDRESS,
+    );
+  }
```
heueristik: A comment explaining what the context is/what the check is doing would be good

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+  ) {
+    causes.push(
+      PluginUpdateProposalInValidityCause
+        .INVALID_GRANT_UPDATE_PERMISSION_PERMISSION_ID,
+    );
+  }
+  return causes;
+}
+
+export async function validateRevokeUpdatePluginPermissionAction(
```
heueristik: A lot of code duplication just to return the same causes for revoke instead of grant.
We should think about refactoring this in an upcoming task.

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+    plugin.appliedPreparation?.pluginAddress ===
+      decodedParams.pluginAddress
+  );
+  if (plugin) {
+    // check release is the same as the one installed
+    if (
+      plugin.appliedVersion?.release.release !==
+        decodedParams.versionTag.release
+    ) {
+      causes.push(PluginUpdateProposalInValidityCause.INVALID_PLUGIN_RELEASE);
```
heueristik: INVALID_PLUGIN_RELEASE is not a good name.
Maybe: UPDATE_TO_INCOMPATIBLE_RELEASE

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+    }
+    // check build is higher than the one installed
+    if (
+      !plugin.appliedVersion?.build ||
+      plugin.appliedVersion?.build >=
+        decodedParams.versionTag.build
+    ) {
+      causes.push(PluginUpdateProposalInValidityCause.INVALID_PLUGIN_BUILD);
```
heueristik: Instead of INVALID_PLUGIN_BUILD, let's use UPDATE_TO_OLDER_OR_SAME_BUILD.

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+}
+
+/**
+ * Returns true if the actions are valid for a plugin update proposal with root permission
+ *
+ * @export
+ * @param {ProposalActionTypes[]} actions
+ * @return {*}  {boolean}
+ */
+export function isPluginUpdateActionWithRootPermission(
```
heueristik:
```suggestion
export function isPluginUpdateActionBlockWithRootPermission(
```

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+  actions: ProposalActionTypes[],
+): boolean {
+  // get the first 5 actions
+  // if the first is a dao upgrade skip it
```
heueristik: The fact that you are carrying this special case (action[0] being DAO update) through all methods is not good. It should be identified beforehand.

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+  actions: DaoAction[],
+  daoAddress: string,
+  implementationAddress: string,
+  currentDaoVersion: [number, number, number],
+): DaoUpdateProposalValidity {
+  const classifiedActions = classifyProposalActions(actions);
+  const causes: DaoUpdateProposalInvalidityCause[] = [];
+  // check if the actions are valid
+  // if they are not valid return add
+  // invalid actions to the causes and return
```
heueristik: unclear comment

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+  if (isDaoUpdateAction(classifiedActions)) {
+    // if they are valid, the upgrade action must
+    // be the first one
```
heueristik: That's hacky and implict. Wondering what forced you to do it like that. Is it that the App wants to have two functions that can be called separately?

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+      );
+    }
+    // if the upgrade action is upgradeTo
+    if (upgradeActionType === ProposalActionTypes.UPGRADE_TO) {
+      // decode the upgradeTo action
+      const decodedImplementationAddress = decodeUpgradeToAction(
+        actions[0].data,
+      );
+      // check that the implementation address is the same
+      if (implementationAddress !== decodedImplementationAddress) {
```
heueristik:
```suggestion
      if (expectedImplementationAddress !== decodedImplementationAddress) {
```
Where will the `expectedImplementationAddress` come from? From the active contracts by the App or SDK?

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +863,801 @@ export function toPluginPermissionOperationType(
...
+        JSON.stringify(initializeFromDecodedParams.previousVersion) !==
+          JSON.stringify(currentDaoVersion)
+      ) {
+        causes.push(
+          DaoUpdateProposalInvalidityCause.INVALID_UPGRADE_TO_AND_CALL_VERSION,
+        );
+      }
+      // TODO
+      // check that the data can be decoded with the abi in the metadata of the version we are
+      // upgrading to. For now is not possible so we check that the data is empty
```
heueristik: Check that we have a task created for this to not forget. Mention that this is relevant for the security check and put a gh permalink referencing this code line.

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -610,50 +605,200 @@ describe("Client", () => {
...
-      expect(typeof actions[1]).toBe("object");
-      expect(actions[1].data).toBeInstanceOf(Uint8Array);
+      for (const [index, actionName] of expectedActions.entries()) {
+        
+         action = actions[index];
+        expect(typeof action).toBe("object");
+        expect(action.data).toBeInstanceOf(Uint8Array);
+        hexString = bytesToHex(action.data);
+        switch (actionName) {
+          case "grant":
```
heueristik, line 620: Why no checks for grant?

modules/client/test/unit/client/utils.test.ts
```diff
@@ -0,0 +1,1488 @@
...
+          expected: false,
+        },
+        {
+          input: [
+            ProposalActionTypes.UPGRADE_TO_AND_CALL,
+            ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
+            ProposalActionTypes.APPLY_UPDATE,
+            ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
+            ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
+            ProposalActionTypes.APPLY_UPDATE,
```
heueristik, line 1424: will this check fail if we put in an update block that doesn't require ROOT_PERM?

modules/client/test/unit/client/utils.test.ts
```diff
@@ -0,0 +1,1488 @@
...
+            ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
+            ProposalActionTypes.UPGRADE_TO,
+          ],
+          expected: false,
+        },
+        {
+          input: [
+            ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
+            ProposalActionTypes.APPLY_UPDATE,
+            ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
```
heueristik, line 1363: will this check fail if we put in an update block that does require ROOT_PERM?

modules/client/src/internal/client/methods.ts
heueristik: Can we put the new methods into a separate file? E.g. called security-check or something related?
The file is already quite long

modules/client/test/integration/client/encoding.test.ts
```diff
@@ -610,50 +605,200 @@ describe("Client", () => {
...
-          applyUpdateParams.permissions[parseInt(index)]
-            .permissionId,
-        );
+      const pspInterface = PluginSetupProcessor__factory.createInterface();
+      const daoInterface = DAO__factory.createInterface();
+      const expectedActions = ["grant", "grant", "applyUpdate", "revoke", "revoke"];
+      let hexString, argsDecoded, action;
+      expect(actions.length).toBe(5);
+      for (const [index, actionName] of expectedActions.entries()) {
+        action = actions[index];
```
heueristik, line 717: Fix order. Why are case 0 and 1 missing?

modules/client/src/internal/client/methods.ts
josemarinas: We could move the implementation, not the function, but we will need to pass all the required params because they wont have access to the context and the rest of stuff on `this`

modules/client/src/internal/constants.ts
josemarinas: It should be pretty easy to add it here if you really want, but if ultimately we are going to scrap this I agree with @Michael-A-Heuer

modules/client/src/internal/types.ts
josemarinas: The permission is called `PLUGIN_UPDATE_PERMISSION`. I think it makes sense to have it like this

modules/client/src/internal/utils.ts
josemarinas: Then we will be unable to know if the action failing is the grant or the revoke. There is a lot of duplication because for this checks functions it seemed cleearer to me to replicate the function than add 2 aditional params `operation` and `permission` and a bunch of ifelse clauses

modules/client-common/CHANGELOG.md
josemarinas: Not on this package

modules/client/src/internal/types.ts
josemarinas: We are calling it update everywhere, even the method in the contracts is `applyUpdate`

modules/client/src/internal/utils.ts
josemarinas: Yep, at the begining I had only one function for all of tthem but it got pretty messy with all the internal conditions and having to pass an `operation` param and a `permission` param

modules/client/src/internal/utils.ts
josemarinas: No, is what we talked about, the DAO update should always bee the first thing in case it exists. The app will need at least wo functioons, one for checking if is a dao update and another for checking if it is a proposal update, because they can do ir separately

modules/client/src/internal/utils.ts
josemarinas: From the dao factory address that the sdk have, if a specifica version is giveen it'll use that version to read the address in the LIVE_CONTRACTS

modules/client/src/internal/utils.ts
josemarinas: https://aragonassociation.atlassian.net/browse/OS-868

modules/client/test/integration/client/encoding.test.ts
josemarinas: because there is no break, is the JS way of saying `case 'grant' || 'revoke'`

modules/client/test/unit/client/utils.test.ts
josemarinas: This check will fail if there is no UPGRADE_TO or UPGRADE_TO_AND_CALL function or if its not on the first position

modules/client/test/unit/client/utils.test.ts
josemarinas: This check will fail if there is no UPGRADE_TO or UPGRADE_TO_AND_CALL function or if its not on the first position

modules/client/test/integration/client/encoding.test.ts
josemarinas: the order is correct, same as on the other comment its the JS way of saying `case 'grant' || 'revoke'`

modules/client/src/internal/types.ts
heueristik: But the permission name is UPGRADE_PLUGIN_PERMISSION (see https://github.com/aragon/osx/blob/c2be413a998c2995f4ed36ca648a5fd9b4284e77/packages/contracts/src/core/plugin/PluginUUPSUpgradeable.sol#L68)

modules/client/src/types.ts
```diff
@@ -420,18 +420,65 @@ export type DecodedInitializeFromParams = {
...
+  PROPOSAL_NOT_FOUND = "proposalNotFound",
+}
+
+export enum PluginUpdateProposalInValidityCause {
+  // Grant UPDATE_PLUGIN_PERMISSION action
+  INVALID_GRANT_UPDATE_PERMISSION_WHO_ADDRESS =
+    "invalidGrantUpdatePermissionWhoAddress",
+  INVALID_GRANT_UPDATE_PERMISSION_WHERE_ADDRESS =
+    "invalidGrantUpdatePermissionWhereAddress",
+  INVALID_GRANT_UPDATE_PERMISSION_PERMISSION =
```
heueristik: The difference between
INVALID_GRANT_UPDATE_PERMISSION_PERMISSION
and
INVALID_GRANT_UPDATE_PERMISSION_PERMISSION_ID
is not so clear from the name alone.

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +865,812 @@ export function toPluginPermissionOperationType(
...
+    );
+  }
+  if (decodedPermission.who !== pspAddress) {
+    causes.push(
+      PluginUpdateProposalInValidityCause
+        .INVALID_GRANT_UPDATE_PERMISSION_WHO_ADDRESS,
+    );
+  }
+  if (
+    decodedPermission.permission !== Permissions.UPGRADE_PLUGIN_PERMISSION
```
heueristik, line 939: I was struggeling to understand what is stored inside `permission: string;`
```
type PermissionParamsBase = {
  where: string;
  who: string;
  permission: string;
};
type PermissionDecodedParamsBase = PermissionParamsBase & {
  permissionId: string;
};
```
Is it the address flag (unset or allowed or a condition address) ?  I was struggeling to find it

heueristik: Clarified on a call. It is the permission name of which we are doing the keccak256 hash

modules/client/src/internal/client/methods.ts
josemarinas: I left as 

`isPluginUpdateProposal` for the boolean
`isPluginUpdateProposalValid` foir the full check

for me is clear enough, we can give a second thought if you feel is necessary

modules/client/test/unit/client/utils.test.ts
josemarinas: yes

modules/client/test/unit/client/utils.test.ts
heueristik: What I wanted to see is you mix the different update action block types (one that does and one that does not require ROOT, e.g.,
```
          input: [
            ProposalActionTypes.UPGRADE_TO,
            ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
            ProposalActionTypes.APPLY_UPDATE,
            ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
            //
            ProposalActionTypes.GRANT_PLUGIN_UPDATE_PERMISSION,
            ProposalActionTypes.GRANT_ROOT_PERMISSION,
            ProposalActionTypes.APPLY_UPDATE,
            ProposalActionTypes.REVOKE_ROOT_PERMISSION,
            ProposalActionTypes.REVOKE_PLUGIN_UPGRADE_PERMISSION,
          ],
```
it should not fail in this case

modules/client/test/unit/client/utils.test.ts
heueristik: See above

modules/client/src/internal/client/methods.ts
heueristik: Ok, I see

modules/client/src/internal/types.ts
heueristik: Things must match with the contracts - otherwise it gets confusing

modules/client/src/types.ts
heueristik: Please create a task to rename "permission" to "permissionName" or similar everywhere.

Formally, a "permission"  is the entire object containing
```
Permission {
  name,
  who,
  id,
  condition,
  where,
}
```

heueristik:
How will the check behave if subgraph is down/not synced? We will return `PROPOSAL_NOT_FOUND`, right?

modules/client/src/types.ts
```diff
@@ -420,60 +418,96 @@ export type DecodedInitializeFromParams = {
...
 
-export type IsPluginUpdateProposalValidParams = {
-  proposalId: string;
-  version: VersionTag;
-  pluginAddress: string;
-  pluginPreparationIndex?: number;
-};
-
 export enum DaoUpdateProposalInvalidityCause {
+  PROPOSAL_NOT_FOUND = "proposalNotFound",
```
heueristik: This is a duplicate of `PROPOSAL_NOT_FOUND` in `export enum ProposalSettingsErrorCause`

modules/client/src/types.ts
```diff
@@ -420,60 +418,96 @@ export type DecodedInitializeFromParams = {
...
   INVALID_ACTIONS = "invalidActions",
```
heueristik: Is this a duplicate of `INVALID_ACTIONS` in `export enum ProposalSettingsErrorCause` ?

modules/client/src/internal/types.ts
josemarinas: Okay Ill change it

modules/client/src/types.ts
```diff
@@ -420,60 +418,96 @@ export type DecodedInitializeFromParams = {
...
-  INVALID_IMPLEMENTATION = "invalidImplementation",
-  INVALID_VERSION = "invalidVersion",
-  INVALID_INIT_DATA = "invalidInitData",
+  INVALID_TO_ADDRESS = "invalidToAddress",
+  NON_ZERO_CALL_VALUE = "nonZeroCallValue",
```
heueristik, line 489: Why is this only a `DaoUpdateProposalInvalidityCause`? Shouldn't it be a general invalidity?
Note, that there is no `NON_ZERO_CALL_VALUE` for `PluginUpdateProposalInValidityCause`.

josemarinas: There is but is specified for each action type

heueristik: Ah, I see there is `INVALID_APPLY_UPDATE_ACTION_VALUE`. Would it make sense to have a general `NON_ZERO_CALL_VALUE` as this is not specific for the two update types we have?

Are we also checking that the grant and revoke action calls have zero `value`?

josemarinas: I mean if we do a general one wo wont be able to tell wich action has non zero value in a plugfin update

docs/sdk/03-update-security-chek/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+- `to` MUST be the DAO address
+- `value` MUST be zero. If not, we return `"nonZeroGrantRootPermissionCallValue"` / `"nonZeroRevokeRootPermissionCallValue"`.
+- `data` MUST contain the `grant` / `revoke` function selector in the first 4 bytes. The subsequent bytes MUST be as follows:
+  - `where` MUST be the DAO proxy address. If not, we return `"invalidGrantRootWhereAddress"` / `"invalidRevokeRootWhereAddress"`.
+  - `who` MUST be the `PluginSetupProcessor` address. If not, we return `"invalidGrantRootWhoAddress"` / `"invalidRevokeRootWhoAddress"`.
+  - `permissionId` MUST be `bytes32 ROOT_PERMISSION_ID = keccak256("ROOT_PERMISSION")`. If not, we return `"invalidGrantRootPermissionId"` / `"invalidRevokeRootPermissionId"`.
+  - `condition` MUST be the empty address. If not, we return `"invalidGrantRootCondition"` / `"invalidRevokeRootCondition"`
+
+**TODO** What we are checking with `"invalidGrantRootPermissionPermission"` `"invalidRevokeRootPermissionPermission"` exactly?
```
heueristik: @josemarinas

docs/sdk/03-update-security-chek/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+- `to` MUST be the DAO address
+- `value` MUST be zero. If not, we return `"nonZeroGrantUpgradePluginPermissionCallValue"` / `"nonZeroRevokeUpgradePluginPermissionCallValue"`.
+- `data` MUST contain the `grant` / `revoke` function selector in the first 4 bytes. The subsequent bytes MUST be as follows:
+  - `where` MUST be the plugin proxy address. If not, we return `"invalidGrantUpgradePermissionWhereAddress"` / `"invalidRevokeUpgradePermissionWhereAddress"`.
+  - `who` MUST be the `PluginSetupProcessor` address. If not, we return `"invalidGrantUpgradePermissionWhoAddress"` / `"invalidRevokeUpgradePermissionWhoAddress"`.
+  - `permissionId` MUST be `bytes32 UPGRADE_PLUGIN_PERMISSION_ID = keccak256("UPGRADE_PLUGIN_PERMISSION")`. If not, we return `"invalidGrantUpgradePermissionPermissionId"` / `"invalidRevokeUpgradePermissionPermissionId"`.
+  - `condition` MUST be the empty address. If not, we return `"invalidGrantUpgradePermissionCondition"` / `"invalidRevokeUpgradePermissionCondition"`
+
+**TODO** What we are checking with `"invalidGrantUpgradePermissionPermission"` `"invalidRevokeUpgradePermissionPermission"` exactly?
```
heueristik: @josemarinas

docs/sdk/03-update-security-check/index.md
```diff
@@ -0,0 +1,121 @@
...
+    - MUST go to the `initializeFrom` function
+    - the additional data passed to `initializeFrom` MUST be empty. If not, we return `"invalidUpgradeToAndCallData"`.
+      - the semantic version number of the previous DAO must be as expected. If not, we return `"invalidUpgradeToAndCallVersion"`.
+  - `upgradeTo` can be called instead, if no reinitalization of the DAO is required. The call
+    - MUST go to the right implementation address. This can be either the latest version or any other, newer version. If not, we return `"invalidUpgradeToAndCallImplementationAddress"`.
+    - MUST have empty subsequent calldata. If not, we return `"invalidUpgradeToAndCallData"`.
+
+### Plugin Updates
+
+For each plugin Updates, we expect a block of associated actions. There can be multiple, independent plugin updates happening in one update proposal.
```
mathewmeconry:
```suggestion
For each plugin update, we expect a block of associated actions. There can be multiple, independent plugin updates happening in one update proposal.
```

docs/sdk/03-update-security-check/index.md
```diff
@@ -0,0 +1,121 @@
...
+  grant({_where: dao, _who: pluginSetupProcessor, _permissionId: ROOT_PERMISSION_ID}),
+  applyUpdate({_dao: dao, _params: applyUpdateParams}),
+  revoke({_where: dao, _who: pluginSetupProcessor, _permissionId: ROOT_PERMISSION_ID}),
+  revoke({_where: plugin, _who: pluginSetupProcessor, _permissionId: UPGRADE_PLUGIN_PERMISSION_ID})
+]
+```
+
+#### Mandatory `applyUpdate` Call
+
+Each block being related to an plugin update MUST contain an `applyUpdate` action exactly once. This action is composed as follows:
```
mathewmeconry:
```suggestion
Each block being related to a plugin update MUST contain an `applyUpdate` action exactly once. This action is composed as follows:
```

mathewmeconry (changes requested):

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +865,821 @@ export function toPluginPermissionOperationType(
...
+  };
+  const res = await graphql.request<U>({
+    query,
+    params,
+    name,
+  });
+  const { pluginInstallations } = res;
+  return pluginInstallations;
+}
+export async function validateGrantUpdatePluginPermissionAction(
```
heueristik: `validateGrantUpgradePluginPermissionAction`

Make sure, the other names are correct too.

modules/client/src/internal/utils.ts
```diff
@@ -824,3 +865,821 @@ export function toPluginPermissionOperationType(
...
+  ) {
+    causes.push(
+      PluginUpdateProposalInValidityCause
+        .INVALID_GRANT_UPGRADE_PLUGIN_PERMISSION_PERMISSION_ID,
+    );
+  }
+  return causes;
+}
+
+export async function validateRevokeUpdatePluginPermissionAction(
```
heueristik: `validateRevokeUpgradePluginPermissionAction`

heueristik (approved):

mathewmeconry (approved):

PR #304, opened 2023-11-22
build: replaces tsdx with dts-cli

mathewmeconry:
## Description

It replaces tsdx which is not maintained anymore with dts-cli.
It also replaces ganache with hardhat for testing contract calls.
Update workflows to use node 18.
Test workflow now tests the SDK for node 16,18 and 20.

## Type of change

## Checklist:


modules/common/package.json
```diff
@@ -17,13 +17,13 @@
     "node": ">=14"
```
josemarinas: Should we unify the required node version?

modules/client/test/integration/multisig-client/methods.test.ts
```diff
@@ -140,7 +131,7 @@ describe("Client Multisig", () => {
       },
     };
     const ipfsUri = await multisigClient.methods.pinMetadata(metadata);
-    const endDate = new Date(Date.now() + 1000 * 60);
+    const endDate = new Date(Date.now() + 1000 * 60 * 60);
```
josemarinas, line 134: You changed endDAte to new Date(0) on the other places, is there a reason why cant do it heere too?

josemarinas (commented):
Looks good, I saw that you removed the blockchain server start and close, I guess hardhat does not use that, how does the hardhat package launch the test chain? its done automatically or is there a piece of code that i missed?

If so does the blockchain get reseted on each test?

modules/common/package.json
mathewmeconry: good idea. done

modules/client/test/integration/multisig-client/methods.test.ts
mathewmeconry: the multisig doesn't support setting the endDate to 0 and choose the end date automatically. So I had to increase it to make sure that it works

mathewmeconry:
> Looks good, I saw that you removed the blockchain server start and close, I guess hardhat does not use that, how does the hardhat package launch the test chain? its done automatically or is there a piece of code that i missed?
> 

It is baked into the `modules/client/test-environment.ts` setup. There it starts hardhat automatically.

> If so does the blockchain get reseted on each test?

No, it doesn't for now we have the same setup where is redeploys the contracts as it was before. But definitely a possibility. We could also activate proper fork testing where we don't need to deploy the contracts anymore.

josemarinas (approved):

PR #327, opened 2024-02-20
Fix/reference guide auto doc

banasa44:
## Description

Update jsdoc2md library. Deleted templates folder as it was unused (they were the default used, so redundant). Updated returns template in tweaks folder, and updated package.json commands to autogenerate the reference-guide documentation.

Task ID: [OS-1055](https://aragonassociation.atlassian.net/browse/OS-1055)

## Chore/Doc

## Checklist:

[OS-1055]: https://aragonassociation.atlassian.net/browse/OS-1055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/tweaks/helpers.js
```diff
@@ -0,0 +1,11 @@
+
+exports.stripTags = stripTags
+
+function stripTags(input) {
+    // Replace HTML entities with their corresponding characters
+    var decodedInput = input.replace(/&lt;/g, '<').replace(/&gt;/g, '>');
```
josemarinas, line 6: first you replace `&lt;` by `<` and then use this replaced to remove `<p>`?

modules/client-common/src/encoding.ts
```diff
@@ -119,7 +119,7 @@ export function decodeRatio(
  * @export
  * @param {string} pluginAddress
  * @param {number} id
- * @return {*}
+ * @return  {string} proposalId
```
josemarinas: Extra whitespace after `@return`

josemarinas:
```suggestion
 * @return {string} proposalId
```

modules/client-common/src/encoding.ts
```diff
@@ -134,7 +134,7 @@ export function encodeProposalId(pluginAddress: string, id: number) {
  *
  * @export
  * @param {string} proposalId
- * @return {*}  {{ pluginAddress: string; id: number }}
+ * @return {{ pluginAddress: string; id: number }}
```
josemarinas: Does this render correctly?

modules/client-common/src/encoding.ts
```diff
@@ -199,7 +199,7 @@ export function bitmapToBoolArray(bitmap: bigint): Array<boolean> {
  *
  * @export
  * @param {string} proposalId
- * @returns {*} {string}
+ * @returns  {string}
```
josemarinas: same here
```suggestion
 * @returns {string}
```

modules/client-common/src/encoding.ts
```diff
@@ -214,7 +214,7 @@ export const getExtendedProposalId = (proposalId: string): string => {
...
```
josemarinas:
```suggestion
 * @returns {string}
```

modules/client/src/internal/client/encoding.ts
```diff
@@ -610,7 +610,7 @@ export class ClientEncoding extends ClientCore implements IClientEncoding {
    *
    * @param {string} daoAddressOrEns
    * @param {DaoUpdateParams} params
-   * @return {*}
+   * @return  {Promise<DaoAction>}
```
josemarinas:
```suggestion
   * @return {Promise<DaoAction>}
```

modules/client/src/internal/client/methods.ts
```diff
@@ -761,13 +761,12 @@ export class ClientMethods extends ClientCore implements IClientMethods {
   /**
    * Retrieves metadata for DAO with given identifier (address or ens domain)
    *
-   * @param {DaoQueryParams} {
+   * @param {DaoQueryParams} params
    *     limit = 10,
```
josemarinas: How does this render?

banasa44: type = DaoQueryParams
name = parms
description = following (limit=10, etc.)

docs/tweaks/helpers.js
banasa44: Yes, so we can strip the tag if it's written as `<p>` or `&lt;p&gt;`
We are not using this function for the moment... I could also just delete the file. However, if we are to take out the `<p>` tags it can be useful.

modules/client/src/multisig/internal/client/methods.ts
```diff
@@ -374,15 +374,14 @@ export class MultisigClientMethods extends ClientCore
   /**
    * returns the members of the multisig
    *
-   * @param {MembersQueryParams} {
+   * @param {MembersQueryParams} params
    *     pluginAddress,
    *     blockNumber,
```
josemarinas: Same here how does this render in the end?

package.json
```diff
@@ -21,17 +21,17 @@
...
-    "docs:tokenVoting": "jsdoc2md --template ./docs/template/main.hbs  --files ./modules/client/src/tokenVoting/*.ts ./modules/client/src/tokenVoting/internal/client/*.ts --configure ./jsdoc2md.json | sed -r 's/(\\{[a-zA-Z]+<[a-zA-Z]+>\\})/`\\1`/mg' > ./docs/sdk/02-reference-guide/03-token-voting.md",
-    "docs:addresslistVoting": "jsdoc2md --template ./docs/template/main.hbs --files ./modules/client/src/addresslistVoting/*.ts ./modules/client/src/addresslistVoting/internal/client/*.ts --configure ./jsdoc2md.json | sed -r 's/(\\{[a-zA-Z]+<[a-zA-Z]+>\\})/`\\1`/mg' > ./docs/sdk/02-reference-guide/04-addresslist-voting.md",
+    "docs:client": "jsdoc2md --partial docs/tweaks/*.hbs --helper docs/tweaks/*.js --files ./modules/client/src/*.ts ./modules/client/src/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/sdk/02-reference-guide/01-client.md",
+    "docs:multisig": "jsdoc2md --partial docs/tweaks/*.hbs --helper docs/tweaks/*.js --files ./modules/client/src/multisig/*.ts ./modules/client/src/multisig/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/sdk/02-reference-guide/02-multisig.md",
```
josemarinas, line 25: I'm not sure I'm a fan of the `tweaks` naming, @Michael-A-Heuer or @jordaniza any suggestion on how to name this

package.json
```diff
@@ -21,17 +21,17 @@
...
+    "docs:tokenVoting": "jsdoc2md  --partial docs/tweaks/*.hbs --helper docs/tweaks/*.js --files ./modules/client/src/tokenVoting/*.ts ./modules/client/src/tokenVoting/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/sdk/02-reference-guide/03-token-voting.md",
```
josemarinas:
```suggestion
    "docs:tokenVoting": "jsdoc2md --partial docs/tweaks/*.hbs --helper docs/tweaks/*.js --files ./modules/client/src/tokenVoting/*.ts ./modules/client/src/tokenVoting/internal/client/*.ts --configure ./jsdoc2md.json > ./docs/sdk/02-reference-guide/03-token-voting.md",
```

modules/client-common/src/encoding.ts
banasa44: This is in the client-common, for which we are not generating documentation. It has been affected by a replace all.
But I check it and it won't render... Should I change it to something like {object} params?

josemarinas (approved):
Looks good, still unsure abut the tweaks naming but we can change it in the future if we want

modules/client/src/internal/client/methods.ts
banasa44: Added a - in front of each description argument so it renders as a list.

package.json
banasa44: Leave it as it may change in the future.

